PR #27, opened 2023-05-18
Edit the message history

michox:
This is my first time working on an obsidian addon and I don't really have time to test this but I asked gpt4 to give the user the ability to edit the chat history like in the playground. It looks pretty good but I am on a super tight deadline and can't spend any more time on this right now, @logancyang maybe you could take a look if this works as expected?

### DIFF
--- a/src/components/Chat.tsx
+++ b/src/components/Chat.tsx
@@ -172,6 +172,17 @@ const Chat: React.FC<ChatProps> = ({
     );
   };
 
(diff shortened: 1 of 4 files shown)


logancyang:
@michox thanks for the PR, this is a good feature indeed, will take a look when I get some time. Currently doing a significant rewrite so this probably will take a bit longer.

PR #28, opened 2023-05-19
Migrate to using langchain templates, conversation memory and chain

logancyang:
Having issue with https://github.com/hwchase17/langchainjs/issues/1327


logancyang:
Addressed by another PR.

PR #38, opened 2023-05-28
Fix type validation on temperature setting

### LINKED ISSUE #37: Options validation issue: temperature accepts any integer and no decimals
I installed the plugin and went to set the temperature.

Found that it accepts any integer (positive or negative), but errors with a decimal (e.g. `0.7`).

Attached screenshots. The validation error is for `0.2`, which is my desired temperature.

Selection_050
Selection_051
Selection_052
Selection_053

rbellamy:
Fixes #37 

Tested locally.

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -101,7 +101,7 @@ export class CopilotSettingTab extends PluginSettingTab {
         })
       )
       .addText((text) =>{
-        text.inputEl.type = "number";
+        text.inputEl.type = "decimal";
         text
           .setPlaceholder("0.7")
           .setValue(this.plugin.settings.temperature)


src/settings.ts
```diff
@@ -101,7 +101,7 @@ export class CopilotSettingTab extends PluginSettingTab {
         })
       )
       .addText((text) =>{
-        text.inputEl.type = "number";
+        text.inputEl.type = "decimal";
```
logancyang, line 104: Could you point me to the `decimal` type HTML specification? How does it behave compared to `number`?

rbellamy: You're right. I can't and this PR should be closed. I was overly quick to suggest it.

PR #78, opened 2023-06-16
fix: caught (in promise) Error: (Azure) OpenAI API key not found

cl1107:
when using azure open ai, switch to "QA: Active Note" or rebuild index for  Active Note the console will report this error
image

### DIFF
--- a/src/aiState.ts
+++ b/src/aiState.ts
@@ -240,7 +240,7 @@ class AIState {
 
   getEmbeddingsAPI(): Embeddings {
     const OpenAIEmbeddingsAPI = new OpenAIEmbeddings({
-      openAIApiKey: this.langChainParams.openAIApiKey,
+      openAIApiKey: this.langChainParams.openAIApiKey || this.langChainParams.azureOpenAIApiKey,
       maxRetries: 3,
       maxConcurrency: 3,
       timeout: 10000,


logancyang:
@cl1107 Thanks for reporting. I'm guessing it happens when your settings have:

- No OpenAI key
- Full Azure credentials provided
- Model: Azure models
- Embedding Provider: OpenAI

Your change isn't going to fix it though. This happened because the embedding provider is still OpenAI, it needs to call OpenAI's embedding endpoint. For Azure you need to deploy an embedding model specifically besides the chat model. So since Copilot doesn't have a way to set Azure as an embedding provider yet, it still requires either OpenAI, CohereAI, or Huggingface for embeddings. Does it make sense?

So this is actually a feature request: Support Azure embedding models. Could you please open this as an issue? I'll circle back to it. In the meantime, you can provide the OpenAI key and use it as embedding provider, they have significantly reduced the pricing. If you absolutely don't want OpenAI, you can use CohereAI, it's free but with rate limiting.

src/aiState.ts
```diff
@@ -240,7 +240,7 @@ class AIState {
 
   getEmbeddingsAPI(): Embeddings {
     const OpenAIEmbeddingsAPI = new OpenAIEmbeddings({
-      openAIApiKey: this.langChainParams.openAIApiKey,
+      openAIApiKey: this.langChainParams.openAIApiKey || this.langChainParams.azureOpenAIApiKey,
```
logancyang, line 243: This is actually not going to fix it, pls refer to my other comment.

logancyang (changes requested):

cl1107:
> @cl1107 Thanks for reporting. I'm guessing it happens when your settings have:
> 
> * No OpenAI key
> * Full Azure credentials provided
> * Model: Azure models
> * Embedding Provider: OpenAI
> 
> Your change isn't going to fix it though. This happened because the embedding provider is still OpenAI, it needs to call OpenAI's embedding endpoint. For Azure you need to deploy an embedding model specifically besides the chat model. So since Copilot doesn't have a way to set Azure as an embedding provider yet, it still requires either OpenAI, CohereAI, or Huggingface for embeddings. Does it make sense?
> 
> So this is actually a feature request: Support Azure embedding models. Opened an issue myself: #81. In the meantime, you can provide the OpenAI key and use it as embedding provider, they have significantly reduced the pricing. If you absolutely don't want OpenAI, you can use CohereAI, it's free but with rate limiting.

my Embedding Provider is huggingface, but ```const OpenAIEmbeddingsAPI = new OpenAIEmbeddings``  is before switch block
image

logancyang:
@cl1107 I see, it should check if any of OpenAI, CohereAI or Huggingface key exists instead, if all 3 are missing then raise an error. Azure can't be added here until there's a separate form for its embedding deployment credentials.

PR #122, opened 2023-08-06
Add usingDocker setting for Local Copilot

Sokole1:
Hi, thanks for creating this great plugin. I have been trying the new Local Copilot feature.

With the "Use Local Copilot" setting I get the error: _LangChain error: AxiosError: Request failed with status code 503 and body Service Unavailable_ because of the PROXY_SERVER_PORT being 3001.

I am running LocalAI on my Macbook **without Docker** as recommended in [LocalAI Getting Started](https://localai.io/basics/getting_started/#:~:text=How%20about%20you%3F%22%7D%7D%5D%7D-,Note,-If%20running%20on). I have personally noticed Docker being around 2-3 times slower on my M2 Max than from source with Metal acceleration. I am running LocalAI with this command: `./local-ai --models-path ./models/ --debug --cors` (--cors must be included).

This PR changes the openAIProxyBaseUrl to `http://localhost:8080` if the user is not using Docker which resolves the error.

Please let me know if any changes are needed.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -114,6 +114,7 @@ export const DEFAULT_SETTINGS: CopilotSettings = {
   userSystemPrompt: '',
   openAIProxyBaseUrl: '',
   useLocalProxy: false,
+  usingDocker: true,
   localAIModel: '',
   stream: true,
   embeddingProvider: OPENAI,
--- a/src/main.ts
+++ b/src/main.ts
@@ -77,7 +77,7 @@ export default class CopilotPlugin extends Plugin {
       // If using local proxy, 3rd party proxy is overridden
       langChainParams.openAIProxyBaseUrl = this.settings.usingDocker
       ? `http://localhost:${PROXY_SERVER_PORT}`
-      : LOCALAI_URL.replace(/\/v1$/, '');
+      : LOCALAI_URL.substring(0, LOCALAI_URL.lastIndexOf('/'));
 
       langChainParams.useLocalProxy = true;
       await this.startProxyServer(LOCALAI_URL);
--- a/src/main.ts
+++ b/src/main.ts
@@ -41,6 +41,7 @@ export interface CopilotSettings {
   userSystemPrompt: string;
   openAIProxyBaseUrl: string;
   useLocalProxy: boolean;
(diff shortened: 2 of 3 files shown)


logancyang:
@Sokole1 Thanks for the PR! I'd like to understand your issue better though.

I'm not using Docker for my demo either, I used the exact setup I have from the [setup guide](https://github.com/logancyang/obsidian-copilot/blob/master/localai_setup.md) on my MacBook air m1, running localAI at port 8080. The "Use Local Copilot" toggle starts a proxy server in the plugin that runs at port 3001, and the "LocalAI" model listens to that 3001 port via the langchainParam ([code](https://github.com/logancyang/obsidian-copilot/blob/master/src/main.ts#L524)). So 3001 should not be a concern of the user, it's just an internal mechanism. As long as the LocalAI server runs on port 8080, it should work (This [code](https://github.com/logancyang/obsidian-copilot/blob/master/src/main.ts#L79) just starts the proxy for a target port, in this case 8080 for LocalAI).

Not sure if this made sense. The gist is that 3001 should not be a concern of the user, and if 8080 isn't working for you somehow, I should understand the why better. Could you describe in detail how I can reproduce this?

BTW, I see that you have the `--cors` flag for the server, that is something I should experiment with more. Perhaps the proxy server is not necessary at all. At the end of the day, I would avoid having 2 toggles, it would be too much for users. If we can get rid of all toggles it would be the best.

Sokole1:
Hi @logancyang. Since you also weren't using Docker, I tried to set up everything again from scratch. I rebuilt LocalAI and followed the Copilot setup guide. Unfortunately, I still ran into the 503 error. Copilot's requests did not reach my LocalAI server (no debug messages on the server side). Maybe there is some setting on my Mac or other process that caused this issue.

You mentioned that perhaps the proxy server is not necessary, so I explored removing it by setting `OpenAI Proxy Base URL` to `http://localhost:8080/v1`.  This runs into CORS issues, so it is necessary to include CORS when running LocalAI. I made those changes and updated the documentation in this PR: #123 

I will close this PR and hopefully the changes in the new PR is a better solution.

Issue #125, opened 2023-08-09
Chat with PDF


Sokole1:
As mentioned in #49, #117. Be able to select regions/pages of a PDF for chat context. Ideally, be able to handle figures, images, math, etc. inside the PDF.

I will explore the implementation of this feature and note down any progress under this issue. Help is welcome.

Sokole1:
See #131 which is a PR to allow for .pdf in QA: Active File. Although the changes are not complete until there is proper handling of large files.

## Things to add:
- Allow users to select a PDF page range for large files
- Allow the use of highlighted text inside PDF as context
- Support more file formats
- Allow using folders as context

## Dev Notes:
- Document loaders appear to work only in Node.js, so I couldn't use PDFLoader. The docs were a bit misleading as the "Node-only" label is only attached to the "File Loaders" section under "Integrations". https://js.langchain.com/docs/modules/data_connection/document_loaders/how_to/pdf. It would be nice if there was a way to use these, or else a custom parser will be needed for each new file type.
- Uses PDFJs loaded in with https://docs.obsidian.md/Reference/TypeScript+API/loadPdfJs (you must use this PDFJs loader or else npm installing pdfjs-dist clashes with Obsidian's installation)

Issue #117, opened 2023-07-23
Feature Request: Apply custom prompts to highlighted text within PDF viewer 


Awwwrnie:
This plugin is amazing. Appreciate your work on it and excited to see what’s next.

Now that obsidian has compatibility for PDFs, I would love to be able to apply prompts to text within the PDF. It would be great if we could highlight text within the PDF and be able to apply a custom prompt to the selection. 

Thanks.

logancyang:
Thanks for reporting this, and sorry for the slow reply.

Copilot v3 is in code freeze, so we're closing this as not planned. v4 is a ground-up rewrite and is launching soon, with a very different architecture around agent mode.

If this is still an issue for you in v4 once it lands, please open a new issue and we'll take a fresh look. 🙏

PR #131, opened 2023-08-21
QA: Active Note now accepts PDF Text

Sokole1:
#125 QA: Active Note now accepts .pdf files.  All the text in the PDF  is extracted as a single string. I also added LocalAI as a setting option for "Default Model". Tested with GPT and LocalAI.

**This fails for large files (read edit below), I don't think these changes should be merged** until this is addressed, as trying to embed a large file could be expensive. Possible fixes include a hard cutoff file size, embedding the file in chunks, or allow users to pick a PDF page range. I will explore setting up a PDF page range. I'm opening this PR anyways because currently, the changes are small and self-contained.

## Edit: Doesn't actually fail for large files
I was trying to embed a 920 page textbook with LocalAI. I think there was something weird in the PDF that caused a segmentation fault inside LocalAI. ChatGPT's embedding worked fine for that same file (see image below). LocalAI's embedding also worked for a 1000+ page PDF of lorem ipsum. However, it seems difficult to get quality responses when working with so much context. I'll look into allowing users to select PDF page ranges.

image


logancyang:
Thanks for your hard work! This is very exciting! Will take a look and try it out. I'll be AFK until next week, feel free to leave further changes and comments, I'll check back then!

Sokole1:
## Update:
Allow the use of all the selection-based prompts including custom prompts outside of the Markdown editor, such as in the PDF viewer and Copilot Chat #117. Demo GIF attached showcasing "Explain Like I'm 5" and "Make Selection Shorter"

### Note: 
My PDF/Chat selection disappears when I open the command prompt so I have to set up keyboard shortcuts to use these commands outside of the Markdown editor.

I looked into extracting PDF annotations such as highlights and comments. It is definitely doable, but since I don't personally annotate my PDFs with highlights, I will leave this to the side for now and explore how to better handle large files.

CopilotPDFSelectDemo

Sokole1:
The reason my LocalAI embedding failed for my large PDF is because I'm using the BERT model which has a token limit of 512. So in `aiState.ts->buildIndex()` when it calls `const textSplitter = new RecursiveCharacterTextSplitter({ chunkSize: 1000 });` I guess depending on how it's chunked, sometimes it goes over the token limit. This is not a problem for other embedding models that have a larger token limit, hence why OpenAI embeddings worked.

Not sure what is the best way to handle this, maybe have a setting to allow people to choose their chunkSize? (and maybe chunkOverlap?

On a similar note, what does the max token setting do? For example, if I try to use "Explain selection like I'm 5", but my selected text goes over the context limit then LocalAI also fails to respond. Again, not sure how to best resolve this issue. Could just return a Notice and error to the user instead of sending the request, or embed their selection? I could implement some fixes, but I'll wait for feedback on how to best handle these token and context-limit errors.

src/components/Chat.tsx
```diff
@@ -142,6 +142,9 @@ const Chat: React.FC<ChatProps> = ({
       return;
     }
     const noteContent = await getFileContent(file);
+    if (debug) {
+      console.log("ALL NOTE CONTENT IS", noteContent)
```
logancyang, line 146: Should we only show truncated content from the PDF in case the text is super long? And if you intend to keep this debugging message, no need to make it all caps, perhaps some delimiters can make it more visible.

src/constants.ts
```diff
@@ -69,6 +69,7 @@ export const DISPLAY_NAME_TO_MODEL: Record<string, string> = {
   [ChatModelDisplayNames.AZURE_GPT_35_TURBO_16K]: ChatModels.AZURE_GPT_35_TURBO_16K,
   [ChatModelDisplayNames.AZURE_GPT_4]: ChatModels.GPT_4,
   [ChatModelDisplayNames.AZURE_GPT_4_32K]: ChatModels.GPT_4_32K,
+  [ChatModelDisplayNames.LOCAL_AI]: 'localai',
```
logancyang, line 72: Maybe we can move the model providers constants below to the top so this can use the `LOCALAI` constant.

src/utils.ts
```diff
@@ -92,8 +93,12 @@ export const formatDateTime = (now: Date, timezone: 'local' | 'utc' = 'local') =
 };
 
 export async function getFileContent(file: TFile): Promise<string | null> {
-  if (file.extension != "md") return null;
-  return await this.app.vault.read(file);
+  if (file.extension == "md") {
+    return await this.app.vault.read(file);
+  } else if (file.extension == "pdf") {
```
logancyang, line 98: Does this mean this PR works only for a PDF file opened directly in Obsidian, and not embedded in another markdown note? I initially misunderstood and tried an _embedded_ PDF, and it didn't work. Don't get me wrong, this is still phenomenal work!! Just wondering, after your research, do you think there's a path to enable these features for embedded PDFs in a markdown file?

logancyang:
@Sokole1 So sorry for the delay, I was back from my trip and got sick immediately 🤒 . Just pulled your PR and tried it out locally, it works! Thank you for your contribution! 🙏 🎉🎉🎉

I left some comments and it generally looks good! Some questions I have now:

- As mentioned in the comment above, do you plan to enable these features for embedded PDFs as well?
- Based on your research, do you think there is a workaround so the user won't have to create a keyboard shortcut and still be able to keep the text selection and use the command palette?
- For the short PDF paper I tried, since it basically parses the PDF into a single string, it fails to understand its structure, like Abstract, Introduction, Methods, and other headings. Do you think there's a way to improve?

No need to address everything in this one PR. Just want to put them out there to get your feedback.

My general goal for this plugin is to make it as easy to use as possible. It'd be great if it could involve as few tricks as possible. 

Regarding your question about max tokens, when speaking purely about OpenAI APIs, it limits the length of both input and output AFAIK. Though it's mostly used to limit the length of the AI response, it does limit the input side as well since the chat completion is [your prompt + AI generated text], i.e. this must be true: `command/chat prompt + selected text + max_tokens <= model context window`, if false it will trigger context length exceeded.

Here's the definition of [max_tokens](https://platform.openai.com/docs/api-reference/chat/create#max_tokens) at OpenAI.

As for the chunk size and overlap, given the "easy to use" point, I'd say we avoid exposing them as settings at the moment. They are the hyperparams to tune for better QA quality so there's quite some room for unexpected behaviors and mistakes. We can set them based on our own experience for now.

Sokole1:
Hi @logancyang, hope you're doing well. Thank you for the comments, I will make those changes. For your questions:
> Based on your research, do you think there is a workaround so the user won't have to create a keyboard shortcut and still be able to keep the text selection and use the command palette?
- Yes, I believe I can do this by attaching an event listener [EventRef - Developer Documentation (obsidian.md)](https://docs.obsidian.md/Reference/TypeScript+API/EventRef) for when the user opens the command palette then save the currently highlighted text before it loses focus.

> For the short PDF paper I tried, since it basically parses the PDF into a single string, it fails to understand its structure, like Abstract, Introduction, Methods, and other headings. Do you think there's a way to improve?
- Recently, Meta Research released Nougat which is an academic document parser: [facebookresearch/nougat: Implementation of Nougat Neural Optical Understanding for Academic Documents (github.com)](https://github.com/facebookresearch/nougat)
	- This converts the pdf to a .mmd (mathpix markdown format) which is a much better format for LLMs to understand
	- I tried it locally, but without CUDA, it is extremely slow, like over a minute per page. 
	- But this is a good sign for hopefully better PDF parsers in the future
	- Perhaps Langchain PDF Parser would be better. Before I couldn't get it to work inside the plugin, but I might try again [PDF files | 🦜️🔗 Langchain](https://js.langchain.com/docs/modules/data_connection/document_loaders/integrations/file_loaders/pdf)
	- Maybe mathpix API (not free) [Process a PDF – Mathpix API v3 Reference](https://docs.mathpix.com/#process-a-pdf)
	- Maybe a fine-tuned LLM for academic papers
	- Maybe a different Langchain function: [Summarization | 🦜️🔗 Langchain](https://js.langchain.com/docs/modules/chains/popular/summarize), [Analyze Document | 🦜️🔗 Langchain](https://js.langchain.com/docs/modules/chains/additional/analyze_document)
	- Just throwing out some ideas, but I guess the best solution really depends on what people use PDFs for in Obsidian,  maybe we'll see as we go. For me, most of my PDFs will be textbooks.

> Do you plan to enable these features for embedded PDFs as well? 
- I will make selection-based queries work on embedded PDFs with a small fix to the code
- For .md QA that contains an embedded PDF, the process of replacing the ![[]] with PDF text should not be difficult, but what text to include leads to a few questions to be answered:

What should the rules for including embedded PDF content in a .md QA be?
- What is the maximum number of pages (if any) to include before making the user select a page range?
- If they include #page= for their PDF, do we only include that page's content or some of the surrounding content?
- If they include #page=&selection=, do we only include the highlighted text?

Here is what I'm considering:
- All PDFs with >= x pages (setting for the user? Or some constant we define) will have a modal window open for the user to select a page range smaller than x
- Only include PDF content if it is embedded with ![[]], not [[]]
- If page is specified, only include that page's text
- If selection is specified, only include that selection's text

Let me know your thoughts, and take it easy!

logancyang:
@Sokole1 I think it's ok to start with a version that does the simplest things and just works, without having the user set anything. We can always extend it in the future.

Currently I have some ideas about embedded stuff that won't be limited to just embedded PDFs. But it will be a bigger initiative. For now I think it's fine to just be clear what the limitation is, e.g. N pages at maximum, and say it will be updated with more powerful features next.

Your research is very helpful! PDF and document understanding is going to be a constant effort, so no need to aim for too much in the first iteration!

jhmonroe:
Curious if any of this is implemented in the latest versions? Essentially this would be a NotebookLM killer for me!
Does Copilot work with PDFs at all?

logancyang:
@jhmonroe This was a great attempt and already worked for many use cases. The plugin has moved on but I kept it here for reference. 

To answer your question, Copilot will enable PDFs to be indexed locally in the vault just like other md notes in the next big release. In the near future Copilot will have multimodal capabilities so even images in the PDF can be referenced in chat. Stay tuned!

PR #179, opened 2023-11-07
Make it work with with OpenRouter.ai

intari:
This fix should (as far as I understood TypeScript) make it working with OpenRouter.ai
Regular version results in "Please set an "HTTP-Referer" header with the URL of your app"

According to OpenRouter's docs at https://openrouter.ai/docs#format - HTTP-referer could be https://localhost/ but it must be set. Title also should be set if possible.


logancyang:
Have you tested locally? The file you changed is actually not used anymore.

PR #195, opened 2023-12-01
fix: pass openAIProxyBaseUrl to embeddings when set openai proxy

Terencesun:
Hi, when I use the thirdparty proxy of openai, I found the base-url of the embeddings api is not replaced by the openAIProxyBaseUrl, so, I add this code to the repo

### Changes
- import the interface ConfigurationParameters of OpenAI
- add the logic to set the embeddings config - basePath
- add the second param when new OpenAIEmbeddings class

:)


logancyang:
OpenAI has migrated their node client to v4, so I had a new fix in place. But thanks for the PR anyway!

Issue #208, opened 2023-12-18
how to chat with entire vault?


iamrobmat:
Is there an option to chat not only with the active note but entire vaults?

logancyang:
hi @iamrobmat! This mode is coming soon, still need to iron out some issues with the local RAG.

logancyang:
> Hey, I'm willing to help out with that feature, just found out the copilot exists and want to use it but unfortunately, it's not very useful without it (a.k.a. not significantly better than the https://chat.openai.com/ which I'm already paying for)
> 
> Specifically for my use case I am willing to explicitly scope down the context based on tags mentioned in my documents and/or a list of folders in the tree.
> 
> I think with that it should be possible to rollout without additional weight of true RAG / embeddings and just stuff everything into the same prompt Arguably RAG might make the results significantly worse, depending on the implementation (for models that support big token windows)
> 
> If you don't have any strong opinion against it I can try to think of something and hack it around, hopefully not interrupting your progress with RAG

Hey @Noobgam these are great points. I will have new modes in the next revamp: 

- A Chat mode (currently called Conversation mode) with the ability to just stuff multiple notes into the same prompt for long context models. The result will be better than just RAG for less specific questions as you mentioned.
- A QA mode powered by local hybrid search (RAG + full-text), this is better for specific questions on a large collection of notes.

Sorry it's taking a while, I will ship the first one first and the local hybrid search will come after.

peterlionelnewman:
If you were unaware, the recent addition of openai APIs to ollama should make integration more straightforward for all future work. I beleive that this should include support for embeddings within ollama shortly.

https://ollama.com/blog/openai-compatibility

logancyang:
@peterlionelnewman I'm monitoring their progress. Ollama currently doesn't support BERT style embeddings which are required for RAG because they were waiting on BERT support from llama.cpp. Looks like a PR was just merged there https://github.com/ggerganov/llama.cpp/pull/5423. Hopefully it will be available in Ollama soon.

logancyang:
v2.5.0 released

PR #245, opened 2024-01-19
WIP Add huggingface transformers embedding

logancyang:
This one is still WIP, opened issues
- langchainjs: https://github.com/langchain-ai/langchainjs/issues/4091
- transformers.js: https://github.com/xenova/transformers.js/issues/526

PR #272, opened 2024-01-29
Vault wide Q&A: Chat with entire vault

AntoineDao:
First of all, thank you @logancyang for all your hard work on this project! I love this plugin and scratched my own itch when implementing this Pull Request. 

I appreciate you might have already been designing/cooking something up that matches this pull request so I am very open to making changes or closing this Pull Request if it trumps existing initiatives.

That being said, let me walk you through this solution for a Vault wide Q&A feature. Let me know what you think and if you would like to merge it in, let me know what changes you think should go in first 👍 

## Vector Store Management

I have replaced the `ttl` based cache and refresh system that indexed files based on their content hash with a file path index that uses the file modified at time (mtime) to determine freshness. The thinking is that how old a file is should not matter for Vault or File QA, instead we care about the vector store representing the current state of the vault.

The main function to look at to understand the storing mechanism and schema used is the `loadFile` function that is a property of the `VectorDBManager` class.

I added a `mtime` property to the `VectorStoreDocument` object that is saved in PouchDB and changed the basic file saving method to be more opinionated about metadata.

I also added a function that returns a `memoryVectorStore` directly from the `VectorDBManager` for convenience.

## User Settings

Users can no longer set a `ttl` on files in the vector store but instead they can choose between three modes of files -> vector store management strategies:
- *Never*: files are never saved to the vector store unless users run a File QA or run a `Save vault to vector store` command
- *On Startup*: files are saved to the vector store on app load and whenever the user runs the `Save vault to vector store` command
- *On Startup and Save*: Same as above but it also saves to vector store whenever the user saves a file

image

/!\ Note: because we have a file path and last modified deduplication system in place, when we saved all files to the vault we will only start saving files that have been modified after the latest modified file in the vector store /!\

## Vault QA

Users are provided with a new option in the Chain select component: `Vault QA`. This creates a chain that uses the entire vector store as a retriever and plugs it into the same RAG chain system as the file QA chain.

Here is a demo of it in action: 

https://github.com/logancyang/obsidian-copilot/assets/20436635/97a63e41-6c68-4dea-8297-2841c4cfad9d

### 🤮 fileNote
The interface name `fileNote` isn't great but I didn't want to spend too much time thinking about it. We should definitely change it if you can think of a better name for it!

### Filtered Vault QA
I have added metadata to all the embedded vector documents which I think we could later use to run "filtered vault QA" features. My thinking is that users might want to ask questions to a subset of their vault. Here are some narrated versions of what I imagine they could want to query in a filtered way:
- What was that thing I wrote about polar bears `in the last 3 days`?
- Who was talking about polar bears in my `#daily-note` notes?

In a sense, filtered vault QA is a superset feature of the File QA (where we filter the RAG chain to a single file in the vector store).

### Adding Sources for Vault QA
I haven't done it in this Pull Request but it would be nice to add a "cite your sources" feature to the Vault wide RAG chain so users can click on the file the CoPilot used as a source for its response. It should be relatively easy to add but felt out of scope for this first pass pull request 😄 

## Related Issues
https://github.com/logancyang/obsidian-copilot/issues/208


logancyang:
@AntoineDao first of all, thank you so much for your initiative and effort 🤯 ! I'm cooking up something similar but this is a great start. I'll take a closer look tmr and come back with some comments.

In general our visions are similar - an agent-like assistant over the (filtered) vault, with the ability to cite sources and understand Obsidian-related requests such as filtering by time range, tags, frontmatter, etc.

My previous thoughts are mainly around the following points

- How to improve retrieval using a local hybrid search (vector + full-text).
- Add a rough cost estimation and progress visualization for the indexing process (requires pricing info from all embedding providers)
- How to cite sources - clickable note titles listed at the end of the response with `[1] [2] [3]...`. It's also relevant for debugging since it's a completely local search, I don't have access to user logs, so the users need the debugging tools on their end.
- How to manage the index. For the vector index, **it must be labeled with the embedding model** that produced it because if the user switched the embedding model only at query time, it won't work.
- What embedding models to use (MTEB leaderboard is a good reference), can they be served locally. So far I haven't found a good enough solution for local. (Sadly Transformers.js doesn't work in the Obsidian env now https://github.com/logancyang/obsidian-copilot/pull/245)

If you think of more, feel free to add to this list.

src/vectorDBManager.ts
```diff
@@ -79,56 +109,105 @@ class VectorDBManager {
...
-      const oldDocs = allDocsResponse.rows.filter(row => {
-          // Assert the doc type
-          const doc = row.doc as VectorStoreDocument;
-          return doc && doc.created_at < thresholdTime;
+      const allDocsResponse = await this.db.allDocs<VectorStoreDocument>({ include_docs: true });
+      const allDocs = allDocsResponse.rows.map(row => row.doc as VectorStoreDocument);
+      const noteFiles = allDocs.map(doc => {
+        const memoryVectors = JSON.parse(doc.memory_vectors) as MemoryVector[];
+        const noteFile: NoteFile = {
+          path: memoryVectors[0].metadata.path,
```
logancyang, line 180: Can we guarantee `memoryVectors` is always not empty?

src/vectorDBManager.ts
```diff
@@ -79,56 +109,105 @@ class VectorDBManager {
     }
   }
 
-  public static async getMemoryVectors(docHash: string): Promise<MemoryVector[] | undefined> {
+  public static async loadFile(noteFile: NoteFile, embeddingsAPI: Embeddings) {
     if (!this.db) throw new Error("DB not initialized");
+    const textSplitter = new RecursiveCharacterTextSplitter({ chunkSize: 5000 })
```
logancyang, line 132: A chunk size of 5000 could be slightly too big for some embedding models, any particular reason for this number? 

ref: https://huggingface.co/spaces/mteb/leaderboard

src/vectorDBManager.ts
```diff
@@ -79,56 +109,105 @@ class VectorDBManager {
     }
   }
 
-  public static async getMemoryVectors(docHash: string): Promise<MemoryVector[] | undefined> {
+  public static async loadFile(noteFile: NoteFile, embeddingsAPI: Embeddings) {
```
logancyang: Should we call it something like `indexFile`? `load` sounds a bit strange here. I think `noteFile` is fine though, since it's a note and it contains the properties of TFile.

src/vectorDBManager.ts
```diff
@@ -79,56 +109,105 @@ class VectorDBManager {
...
     try {
-      const doc: VectorStoreDocument = await this.db.get(docHash);
-      if (doc && doc.memory_vectors) {
-        return JSON.parse(doc.memory_vectors);
-      }
+      // Attempt to fetch the existing document, if it exists.
+      const existingDoc = await this.db.get(docHash).catch(err => null);
+
+      // Prepare the document to be saved.
+      const docToSave = {
```
logancyang, line 153: Correct me if i'm wrong, so each `noteFile` maps to one `docToSave` with its `serializedMemoryVectors`, then could there be a case when the user switches to a different embedding model at one point, starts to index new files with that model, so they end up having a memory vector store of a bunch of `serializedMemoryVectors` embedded by different models 🤔 ? If that can happen it's quite hard to manage. I wonder what's a good way to avoid this, WDYT?

Could we add the embedding model used into some metadata here to keep track, and pop a warning when the vector store of the vault has mixed embeddings? In my mind, the vector store of the entire vault should have a unified interface that can ensure it always has the same type of embedding.

src/vectorDBManager.ts
```diff
@@ -79,56 +109,105 @@ class VectorDBManager {
...
+        content: splitDocument[i].pageContent,
+        metadata: {
+          ...noteFile.metadata,
+          title: noteFile.basename,
+          path: noteFile.path,
+
+        },
+        embedding: docVector,
+      }))
+    const docHash = VectorDBManager.getDocumentHash(noteFile.path);
```
logancyang, line 145: `docHash` as the hash of the note path works well as the id of the entry, but do you think it's also useful to have a `hash(content)` in the entry that can be used to tell whether the note has been modified? Or is it already covered by the mtime mechanism?

logancyang (commented):
Overall this looks great! Had some open-ended questions for us to think a bit more. Awesome job!

src/vectorDBManager.ts
AntoineDao: From the `loadFile` function above it looks like a pretty solid guarantee that it would always be not empty but it might be worth adding a bit of "defensive coding" in there just in case 🤔

src/vectorDBManager.ts
AntoineDao: I picked a number that vaguely worked. I reckon this should be a variable that users can change. What do you think a good default value would be?

src/vectorDBManager.ts
AntoineDao: Yep more than happy to change this 😄

src/vectorDBManager.ts
AntoineDao: I had not considered the case of different embedding models being used. 

I can think of two options (including the one you proposed):
- Use metadata in documents to filter by embedding model. This is simple at read time and write time. It will make it harder to use the modified time cache strategy currently implemented to limit indexing to files modified since last index
- Create a separate logical database per embedding model. We currently initialise the pouchDB with a database called `copilot_vector_stores`. We could instead use a name based on the embedding model to create multiple parallel databases for each embedding model and managed their indexing cache in the same way.

Any thoughts on the pros/cons of each option?

src/vectorDBManager.ts
AntoineDao: I think this is already covered by the mtime mechanism. Using the doc content hash has it's merits in terms of being quite exact in the case where a document has been changed and then reverted (with ctrl+z for example) but using mtime allows us to efficiently index files without having to run a hash of all of them.

src/vectorDBManager.ts
logancyang: Did 1000 not work when you tried? Then I guess 3000 could be a safe number to start with since most embedding models should be able to handle it. 

Letting the user set it is also one way, but could be out of scope for this PR. This is a key param for RAG tuning, so I guess if people open issues on failure cases in the future I can just tell them to tune it themselves lol!

src/vectorDBManager.ts
logancyang: It's a neat idea to have separate db for different embedding models, but when we have more and more embedding providers added, it creates an additional overhead to think about, and more storage. I'm inclined to a simpler setup first for now. We could add a warning message at the embedding provider dropdown, something like: "Switching embedding provider will result in re-indexing your whole vault", and then make sure we always have the same embedding model used in one index.

AntoineDao:
Great to hear that we're both approaching this with the same goals in mind!

## Vault QA - Further Thoughts
- How to improve retrieval using a local hybrid search (vector + full-text).
 	- I think this is interesting, definitely something we can build upon but probs not in this PR 😅
- Add a rough cost estimation and progress visualization for the indexing process (requires pricing info from all embedding providers)
	- Nice to have, probably not for this PR but we could definitely give an indication of token count and then some rough math
- How to cite sources - clickable note titles listed at the end of the response with `[1] [2] [3]...`. It's also relevant for debugging since it's a completely local search, I don't have access to user logs, so the users need the debugging tools on their end.
	- I would really like to get this in next, I think we can do it with prompt engineering and some fixing of the RAG pipeline (I have done it for another project this way before)
	- Is it something we want to do for this Pull Request?
- What embedding models to use (MTEB leaderboard is a good reference), can they be served locally. So far I haven't found a good enough solution for local. (Sadly Transformers.js doesn't work in the Obsidian env now [WIP Add huggingface transformers embedding #245](https://github.com/logancyang/obsidian-copilot/pull/245))
	- I haven't looked into any of the local embedding models solutions yet. I figured that we could leverage Langchain for this purpose?


## Next Steps for this Pull Request
I would like to get some agreement on what the next steps for this pull request should be so we can merge it in. We could then build some of the features discussed above incrementally.

Here's some comments I will address in this pull request. Let me know if there is anything else you think should be addressed.

- I'll make the memory vector fetch safer by not assuming the array is not empty
- I'll add the text splitter chunk size as a settings parameter
- I'll change `loadFile` to `indexFile`

If we think it's important I might try use a PouchDB instance per embedding model.

src/vectorDBManager.ts
AntoineDao: It was quite quick to set up as a plugin setting so I went ahead an did it 😅

src/vectorDBManager.ts
AntoineDao: I quite like that as an option if we're happy keeping it simple for now. It's nice to not over-optimise early on 😅

That being said, I just tried adding embedding providers in the metadata and it was relatively easy to refactor the code to handle using modified at time + embedding model to decide which files to embed on load.

let me know what you think 😄

logancyang:
Thanks for your hard work @AntoineDao !! I think your version is a solid foundation. I'm going to merge it to a feature branch first, add some incremental changes later to iron out some UX issues before merging everything into master. This is an epic feature and thanks again for your significant contribution!

logancyang:
Just FYI your work is now live on this branch https://github.com/logancyang/obsidian-copilot/tree/feature-vault-qa-beta, will be merged when I get some stuff in.

AntoineDao:
Shout-out is appreciated but not very important. Just happy to get a cool feature into this plugin 😀 Let me know if Therese anything I can help with. I have a few hours to kill weekly I could spend on this plugin 🙂

PR #288, opened 2024-02-11
feat: push tagged chat nodes along the context, reconfigure .editorconfig to match current style better

Noobgam:
The gist is to allow the note context to be configurable, in the current state chat context is still pretty much useless.
Defaults to old behavior, allows comma-separated list of tags (no additional safety checks done for that field)

Reads full file content and looks up the file content

Also minor adjustments to code style because it doesn't seem to even remotely match the actual one in code repo

P.S. I was baffled when I read the `"Please reply with the following word for word"`, that will make the costs skyrocket for literally no reason, but don't think there's a reason to blow up the scope of this PR

### DIFF
--- a/src/components/Chat.tsx
+++ b/src/components/Chat.tsx
@@ -151,33 +151,56 @@ const Chat: React.FC<ChatProps> = ({
     }
 
     let noteFiles: TFile[] = [];
-    if (debug) {
-      console.log('Chat note context path:', settings.chatNoteContextPath);
-    }
-    if (settings.chatNoteContextPath) {
-      // Recursively get all note TFiles in the path
-      noteFiles = await getNotesFromPath(vault, settings.chatNoteContextPath);
-    }
 
-    const file = app.workspace.getActiveFile();
-    // If no note context provided, default to the active note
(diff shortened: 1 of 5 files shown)


src/components/Chat.tsx
```diff
@@ -151,33 +151,56 @@ const Chat: React.FC<ChatProps> = ({
...
+		}
+		if (settings.chatNoteContextTags) {
+			const tags = settings.chatNoteContextTags.split(',');
+			// TODO: there should probably be a better way, obsidian has native link support.
+			const allFiles = app.vault.getFiles();
+			for (const file of allFiles) {
+				try {
+					const content = await getFileContent(file);
+					for (const tag of tags) {
+						if (content?.includes(tag)) {
```
logancyang, line 171: How is this checking for tags? This is checking the entire file content.

logancyang:
Thanks for submitting the PR. I don't think it's ready to be merged. The filter by tags feature isn't properly implemented and tests should be added.

PS, your attitude also doesn't look helpful and collaborative. You are entitled to your opinion and feel free to use your branch on your local vault.

PR #291, opened 2024-02-14
First version of chat context filter by tags

logancyang:
SCR-20240213-uojq

SCR-20240213-upca

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -13,7 +13,8 @@ import {
   RetrievalQAChain
 } from "langchain/chains";
 import moment from 'moment';
-import { TFile, Vault } from 'obsidian';
+import { TFile, Vault, parseYaml } from 'obsidian';
+
 
 export const isFolderMatch = (fileFullpath: string, inputPath: string): boolean => {
   const fileSegments = fileFullpath.split('/').map(segment => segment.toLowerCase());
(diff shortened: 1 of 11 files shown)


src/utils.ts
```diff
@@ -38,6 +39,37 @@ export const getNotesFromPath = async (vault: Vault, path: string): Promise<TFil
   });
 }
 
+export async function getTagsFromNote(file: TFile, vault: Vault): Promise<string[]> {
+  const fileContent = await vault.cachedRead(file);
+  const frontMatter = parseYaml(fileContent.split('---')?.[1] ?? '') || {};
+  const tags = frontMatter.tags || [];
+  // Strip any '#' from the frontmatter tags. Obsidian sometimes has '#' sometimes doesn't...
```
petery789, line 46: 😂

petery789:
Screen Shot 2024-02-14 at 9 12 31 PM

A few feedback on UI: 
1. I guess it would be useful to mention relative/absolute path in "filter by folder path", although after checking code it seems that either works. As a user, this was the first thing I was wondering before typing a path. 
2. As a follow up to 1), would it be useful to have default path populated in the input box? 
3. if I understand it correctly, filtering would work like below:

no path, no tags -->  active note
w/ path, no tags --> path filter respected only
w/ path, w/ tags --> path filter AND tags filter 
no path, w/ tags --> tags filter respected only 

In the future, it might be worth considering to change "no path, no tags" to all the notes, or all the notes under the active directory?. Because, when I changed from "no path, no tags" to "no path, w/ tags", I directly thought "tags" would be the only thing that had changed, but in fact path changed from one file to all files too.

logancyang:
@petery789 good points

> As a follow up to 1), would it be useful to have default path populated in the input box?

My idea is to have the "up" button default to active note when nothing is set here, so a default path wouldn't work. But open to suggestions.

> no path, no tags --> active note
> w/ path, no tags --> path filter respected only
> w/ path, w/ tags --> path filter AND tags filter
> no path, w/ tags --> tags filter respected only

Spot on! Only a small addition to the last one, tags are treated as OR, so `tag1,tag2` will return all notes with either tag1 or tag2 or both. This is in the text instruction too. But yeah, your summary is accurate.

> when I changed from "no path, no tags" to "no path, w/ tags", I directly thought "tags" would be the only thing that had changed, but in fact path changed from one file to all files too.

Yes but from the user perspective, "only tags" filters all notes by tags, "only path" filters all notes by path, "nothing" defaults to active note because you don't have any filtering to do. I feel it's a straightforward ux. Perhaps writing these out explicitly in that modal is clearer.

I have at least 2 followup tasks in mind for this:
- [ ] Add the above 4 combinations explicitly
- [ ] Make tags appear as small pills just like they are rendered in the note property
SCR-20240214-sxzj

petery789:
>  I have at least 2 followup tasks in mind for this:
👍 

> "only tags" filters all notes by tags, "only path" filters all notes by path, "nothing" defaults to active note
alright, can work 🤔

PR #304, opened 2024-02-20
fix: empty tags result as empty list even if folder is good

Lisandra-dev:
See #303


logancyang (approved):
Good catch, thanks for the fix!

Issue #309, opened 2024-02-22
[FR]: Remove frontmatter / properties from notes 


Lisandra-dev:
**Is your feature request related to a problem? Please describe.**
Asking information about a note will sometimes includes the informations from the frontmatter. For example, when resuming in bullet point, GPT will also give the frontmatter information, like cssclasses or title. 

**Describe the solution you'd like**
Before the note is send to the IA, frontmatter must be removed and the note is send after to the IA, without the frontmatter

**Describe alternatives you've considered**
/

**Additional context**
/

logancyang:
Thanks for the last contribution @Lisandra-dev! Are you interested in working on this one too?

Lisandra-dev:
I can do it using the last obsidian API if you want

logancyang:
Another comment I thought of is that this could be an option users can turn on and off. Sometimes it's useful to have frontmatter sent to the prompt, for example, if the frontmatter has time and date for meetings and projects.

Lisandra-dev:
This could be in the settings or custom prompt? Or in context modal?
The last option seems to brme better.

logancyang:
> This could be in the settings or custom prompt? Or in context modal? The last option seems to brme better.

I agree, it could be a toggle in the custom prompt modal.

Lisandra-dev:
Will do it later today! O/

PR #311, opened 2024-02-23
feat: remove frontmatter

Mara-Li:
Using last version of ObsidianAPI, the use of `getFrontMatterInfo` allow to remove easily the frontmatter

See #309


package.json
```diff
@@ -31,7 +31,7 @@
     "esbuild": "0.17.3",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.5.0",
-    "obsidian": "latest",
+    "obsidian": "^1.5.7-1",
```
logancyang, line 34: I pulled this PR locally and ran npm install, `npm list obsidian` gives me 1.5.7-1 correctly. However, I see this when I run the plugin
image

Can you reproduce it? This is weird.

Mara-Li: You need the last obsidian version

logancyang: @Lisandra-dev you mean I need to update the Obsidian app itself, not just the API version?

logancyang: OK I see, I updated the app and it's working now. But I'm hesitant to make this breaking change just for frontmatter removal, it forces all users to upgrade the Obsidian app. Can this just be a simple util we can implement by ourselves?

logancyang: For example, in `getTagsFromNote` there's already a way to get the frontmatter. No need to make a breaking change for this one simple util.

Mara-Li: To be honest, I think it's the simpler and safer way to remove frontmatter. But we can use a simple regex to remove it. I can change it ASAP.

Lisandra-dev:
I noticed the way you handle the frontmatter to find frontmatter tags. We need to use the frontmatter API to get it, as parsing with `---` can rapidly lead to error (for example, wrong frontmatter, text starting with line, or simply a header with `---` syntax)

logancyang:
> I noticed the way you handle the frontmatter to find frontmatter tags. We need to use the frontmatter API to get it, as parsing with `---` can rapidly lead to error (for example, wrong frontmatter, text starting with line, or simply a header with `---` syntax). Typically, a file without frontmatter with lead to an error

I agree. The util from their API must be doing better handling for sure. It just doesn't justify a breaking change IMHO. Most users with older versions of Obsidian will simply think the plugin is broken and they won't actively go and find our message "you have to upgrade your Obsidian to continue using it". So the bad UX falls onto the plugin.

I'm pushing v2.5.0 soon with Vault QA mode. Perhaps that could justify a breaking change. But in general, I really dislike breaking changes from an API that does not come from a major version bump.

Lisandra-dev:
No worry, I totally understand. That's why I changed for another method that use less recent API (reading directly the frontmatter using app.metadatache)

src/utils.ts
```diff
@@ -126,9 +125,12 @@ export const formatDateTime = (now: Date, timezone: 'local' | 'utc' = 'local') =
   return formattedDateTime.format('YYYY_MM_DD-HH_mm_ss');
 };
 
-export async function getFileContent(file: TFile, vault: Vault): Promise<string | null> {
+export async function getFileContent(file: TFile, vault: Vault, app: App): Promise<string | null> {
```
logancyang: If we pass `app` it seems we can access `vault` from it, so there's no need to pass `vault` anymore?

logancyang:
Thanks! Could you rebase on master and update the tests accordingly? Should be good to merge after.

Lisandra-dev:
For the test, I don't know how to fix them, to be honest!

src/utils.ts
Mara-Li: Yup, will update it

Lisandra-dev:
Also, `getMarkdownFiles` doesn't need to be async, why do you use await here?

Lisandra-dev:
Hello, I added the last suggestion in #309 
I don't know how to update the test tbh

logancyang:
@Lisandra-dev thanks for the update. 

1. Were you able to pull from master and rebase? This PR shows changes from master, but it should not.
2. You can update the tests and add more test cases for your change, since the function is updated to handle these cases better. It's helpful to check for cases where it failed with the old way and works with your new change. If you don't know how to write those tests, you can just ask GPT4. It's quite good at it.

Lisandra-dev:
1. What do you mean? 
2. The problem is that I need to integrate the App in __mocks__ and i really don't know how to do it... When I try to just call Obsidian.App() I have App is not a constructor :/

logancyang:
> 1. What do you mean?
> 2. The problem is that I need to integrate the App in **mocks** and i really don't know how to do it... When I try to just call Obsidian.App() I have App is not a constructor :/

1. This PR should contain only your changes, not mine. For example, `tests/customPromptProcessor.test.ts` and a lot of other file changes are merged from master. You rebase your commits on top of the newest master.
2. The obsidian mock is inside `__mocks__/obsidian.js`, you can add your mocks as needed.

Lisandra-dev:
As I said, I don't know how to create __mocks__

src/utils.ts
```diff
@@ -39,29 +39,25 @@ export const getNotesFromPath = async (vault: Vault, path: string): Promise<TFil
   });
 }
 
-export async function getTagsFromNote(file: TFile, vault: Vault): Promise<string[]> {
-  const fileContent = await vault.cachedRead(file);
-  const frontMatter = parseYaml(fileContent.split('---')?.[1] ?? '') || {};
-  const tags = frontMatter.tags || [];
+export async function getTagsFromNote(file: TFile, app: App): Promise<string[]> {
+  const frontMatter = app.metadataCache.getFileCache(file)?.frontmatter
+  const tags = frontMatter?.tags || [];
```
Noobgam, line 44: For the sake of people who will be using this repository, even though Logan seems to not consider the tags in the note content for whatever reason, Obsidian in fact does.

Should be a oneliner-change after this PR, you can take inspiration from here
https://github.com/Noobgam/obsidian-copilot/pull/35/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R54

Noobgam:
@Lisandra-dev as for mocking, I took notice of your PR here and merged a rough sketch into my fork. (thanks for your contribution)

https://github.com/Noobgam/obsidian-copilot/pull/35/files#diff-fd006daabf381c84fffc8940f478f913bb04353b644854edb10d2d100586cf70

You can try to follow the same steps and see how far you can get, I didn't focus on tests much, but the PR includes basic mocking of metadataCache that will allow you to do anything you need

Noobgam:
> Also, getMarkdownFiles doesn't need to be async, why do you use await here?

To answer your previous question since Logan didn't:
It is a bug in his mocking and lack of linter setup in the repository. The method signature is not async, whereas mocks defined are async. This is a bug and in order to fix this you need to change the way mocks are implemented. At runtime there isn't a big difference when you have this additional await, but it does get highlighted in most IDEs automatically.

You can also take a look at the PR in my fork, that issue is addressed there as well

PR #436, opened 2024-04-30
Add support for organization-id for OpenAI

### LINKED ISSUE #435: Support Organization ID for OpenAI
For many companies, Open AI requires an Organization ID along with API key. Add support to include organization id setting and use it when available. The lagnchain APIs already support the `organization` parameter for this purpose.

Add a new field in the settings organization-id that stores the org-id and pass it to langchain or proxy agent when using Open AI.

Alternative would be to set global environment variable for the organization ID. But this option would not be clear for the users who need to set the organization-id to use the OpenAI for work.

**Additional context**
Example setting view with organization id field:
Screenshot 2024-04-29 at 10 26 02 PM

avadhpatel:
resolves #435  

This PR adds a new field 'Open AI Organization ID' in the settings page right after Open AI Key as shown in the screensho.
Screenshot 2024-04-29 at 10 43 06 PM

This field is optional, and if provided it will be passed down to langchain's API calls or to the proxy http-agent.

### DIFF
--- a/src/LLMProviders/chatModelManager.ts
+++ b/src/LLMProviders/chatModelManager.ts
@@ -68,6 +68,7 @@ export default class ChatModelManager {
       [ModelProviders.OPENAI]: {
         modelName: params.openAIProxyModelName || params.model,
         openAIApiKey: decrypt(params.openAIApiKey),
+        openAIOrgId: decrypt(params.openAIOrgId),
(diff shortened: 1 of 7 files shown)


logancyang:
Thanks for the PR! Sorry for the delay!

Issue #360, opened 2024-03-12
OpenAI Proxy base url CORS errors


pperanich:
**Describe the bug**
When attempting to fetch data from an OpenAI proxy URL (https://xxxxx.xxx/chat/completions) within Obsidian, the request is blocked due to a CORS policy error. The console displays an error indicating that the response to the preflight request doesn't pass the access control check because no 'Access-Control-Allow-Origin' header is present on the requested resource.

**To Reproduce**
- Configure Obsidian to use an OpenAI proxy URL.
- Attempt to fetch data from the OpenAI API through the proxy.
- Observe the CORS policy error in the console.

**Expected behavior**
The expected behavior is that the proxy server should handle CORS appropriately by including the necessary Access-Control-Allow-Origin header in the response, allowing requests from app://obsidian.md. This would enable the application to communicate with the OpenAI API without encountering CORS policy errors.

**Proposed Fix**
If an OpenAI proxy base url is provided, I think that the proxy server should be started similar to what is done for the Claude model, but providing the OpenAI proxy base url.

**Additional context**
To enhance user experience and flexibility, I would suggest support for multiple proxy configurations that may be selected from the models drop-down. This way, you could still make use of the OpenAI models if wanted as well, rather than simply overwriting the request if a proxy address is provided. This feature would cater to a variety of use cases and preferences, enabling a more seamless and efficient workflow within Obsidian.

kteppris:
First thanks for your amazing work!

I actually run in the same Issue but from a different Use Case:

We are running our own Text Generation Inference API of Huggingface with same models where some of them share the same API as OpenAI therefore we can actually use the openAI package to use our Inference. But when trying to use the proxy i ran into the same CORS Error. What a pitty, that would be amazing to have, not only for the Chat Model but also for the Embeddings, where our models also can be used same way as OpenAI.

Hope this is fixed soon :)

etlweather:
I had this issue with LocalAI - the solution was to start LocalAI with `--cors`

PR #494, opened 2024-08-01
add support for OpenAI proxy server

Ebonsignori:
Closes https://github.com/logancyang/obsidian-copilot/issues/360

image

Adds a new options to "Use local proxy server for OpenAI"

Was able to test with a custom "OpenAI Proxy Base URL" and get things working. Previously I was blocked by CORS requests.

I don't have anything setup with Anthropic so I tried to leave the existing functionality, but I haven't been able to test it.

### DIFF
--- a/src/LLMProviders/chatModelManager.ts
+++ b/src/LLMProviders/chatModelManager.ts
@@ -108,7 +108,22 @@ export default class ChatModelManager {
       },
     };
 
-    return { ...baseConfig, ...(providerConfig[chatModelProvider as keyof typeof providerConfig] || {}) };
+    const selectedProviderConfig =
+      providerConfig[chatModelProvider as keyof typeof providerConfig] || {};
+
+    // When useOpenAILocalProxy is enabled, use local proxy server
+    // Local proxy server will proxy requests to openAIProxyBaseUrl
+    if (
+      chatModelProvider === ModelProviders.OPENAI &&
+      params.useOpenAILocalProxy &&
+      params.openAIProxyBaseUrl
+    ) {
+      (
+        selectedProviderConfig as (typeof providerConfig)[ModelProviders.OPENAI]
+      ).openAIProxyBaseUrl = `http://localhost:${PROXY_SERVER_PORT}`;
+    }
+
+    return { ...baseConfig, ...selectedProviderConfig };
   }
 
   private buildModelMap() {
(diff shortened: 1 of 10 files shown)


src/components/ChatComponents/ChatIcons.tsx
```diff
@@ -62,25 +64,26 @@ const ChatIcons: React.FC<ChatIconsProps> = ({
     const selectedModel = event.target.value;
     setCurrentModel(event.target.value);
 
-    if (selectedModel === ChatModelDisplayNames.CLAUDE) {
-      // Start the proxy server when CLAUDE is selected
-      await proxyServer.startProxyServer('https://api.anthropic.com/');
+    // Start proxy server based on the selected model & settings
+    const proxyServerURL = proxyServer.getProxyURL(selectedModel);
+    if (proxyServerURL) {
+      await proxyServer.startProxyServer(proxyServerURL, selectedModel !== ChatModelDisplayNames.CLAUDE);
```
Ebonsignori, line 70: The `selectedModel !== ChatModelDisplayNames.CLAUDE` check here and on line 79 may not be needed. 

If proxy support works for Anthropic when rewriting paths, then the 2nd argument can be removed from `startProxyServer`

src/LLMProviders/chatModelManager.ts
```diff
@@ -108,15 +108,31 @@ export default class ChatModelManager {
...
+      chatModelProvider === ModelProviders.OPENAI &&
+      params.useOpenAILocalProxy &&
+      params.openAIProxyBaseUrl
+    ) {
+      (
+        selectedProviderConfig as (typeof providerConfig)[ModelProviders.OPENAI]
+      ).openAIProxyBaseUrl = `http://localhost:${PROXY_SERVER_PORT}`;
+    }
+
+    return { ...baseConfig, ...selectedProviderConfig };
```
Ebonsignori, line 126: This could possibly be made more flexible to support a local proxy server for any model.

It looks like more providers support `openAIProxyBaseUrl`. We could enable the local proxy for any provider that supports `openAIProxyBaseUrl` or a switch/if statement to manuallu set the url to the proxy server when it's another key, like `anthropicApiUrl` and have the proxy server proxy to the respective URL

Ebonsignori: I didn't take this route because my needs were narrow and I don't have any other subs / keys to models to verify that they work

Ebonsignori:
Closed because I didn't want to open a PR against `master` so I could use BRAT with my branch

Issue #275, opened 2024-01-31
Please support mobile device


chisungbae:
**Is your feature request related to a problem? Please describe.**
This plugin is really useful to enhance productivity.
Please support mobile device too.

logancyang:
Hey @chisungbae , thanks for the FR, this is on my roadmap. I'll prioritize it once the revamp is released. Stay tuned!

loerzi:
Hey @logancyang great to hear that mobile support is coming! Thanks for the great plugin!

nickfixit:
Moviley!🙏

hughlett:
Thanks for your work @logancyang!

#8, #52, and #162 could be closed as duplicates now

logancyang:
Please refer to #532 to follow along. Closing this as duplicate.

Issue #162, opened 2023-10-08
Mobile support


dukeofgaming:
It would be nice to have this available in the mobile mode of Obsidian

NickDevoctomy:
+1 for this, it would be awesome to be able to use this across all of my devices!

nickfixit:
Need a method for this. You could easily implement with JavaScript solely as a plugin.

logancyang:
Please refer to #532 to follow along. Closing this as duplicate.

Issue #52, opened 2023-06-01
iPad Pro support


Edodamm:
It is currently impossible to install the plugin on an iPad Pro.
Please, please, please...
Is there any chances for it to be implemented in the future?
Thanks for the amazing work!

logancyang:
@Edodamm Thanks for being an early adopter! 

The reason I set this plugin to be desktop-only, for now, is to explore the options to run LLM models locally and have this plugin work offline without any internet connection. I'm a big believer in local AI and privacy. That said, I definitely see demand for mobile support. Currently I'm exploring the option to run local models, and am deliberately avoiding desktop-dependent components to pave the way to mobile support. So please stay tuned!

dxcore35:
> exploring

Very good ideology! I hope you will utilize Apple M1 chips, as they are amazing with GPU for this job!

v3i1y:
Why don't just disable desktop only feature on mobile instead of disabling the whole plugin? People like me selfhost models on home servers, this can be useful even for private AI.

logancyang:
Please refer to #532 to follow along. Closing this as duplicate.

Issue #8, opened 2023-05-07
Mobile support


tensorcopy:
Great plugin! Just noticed that this plugin is not supported in mobile app. Is there a plan to support it? Thanks!

logancyang:
@np-22 Thank you for the suggestion! Will need to do more research and testing to see what the best mobile experience looks like, but mobile support is definitely in the next steps! Please stay tuned!

Ryonez:
Also going to leave my interest for this here.

logancyang:
Please refer to #532 to follow along. Closing this as duplicate.

PR #513, opened 2024-08-17
Adds mobile support

gianluca-venturini:
Adds mobile support addressing feature requests https://github.com/logancyang/obsidian-copilot/issues/8, https://github.com/logancyang/obsidian-copilot/issues/52, https://github.com/logancyang/obsidian-copilot/issues/162, https://github.com/logancyang/obsidian-copilot/issues/275 and https://github.com/logancyang/obsidian-copilot/issues/532.

In order to add mobile support I needed to remove all dependencies on Nodejs and use browser dependencies instead. There were two Nodejs dependencies:
1. Nodejs `event` module, specifically the class `EventEmitter`. I replaced it with `EventTarget` that is natively supported in the browser.
2. `src/proxyServer.ts` depends on `koa` framework that depends on many Nodejs internal packages. I removed the `proxyServer` for now and verified that we can still successfully interact with OpenAI models.

## How to test
There are two ways to test this change:
1. Emulate mobile mode locally with `this.app.emulateMobile(true);`. This is a good first step, but doesn't guarantee that everything is working as expected on the real device.

(old test: https://www.dropbox.com/scl/fi/4nbxhacidxlzb0r8l1p7w/manual_test_desktop_emulate_mobile.mov?rlkey=sbsqxux49mikt07kfe95k5bvn&st=zfau1qbl&dl=0)

- LMStudio with CORS headers, streaming enabled: https://www.dropbox.com/scl/fi/lrvqzutm20w0xa8w9l3x3/cors-streaming.mov?rlkey=32cg0qe0he0ell7b9fyqmszyd&st=vgwk3j38&dl=0
- LMStudio with No CORS headers, proxy enabled in settings, streaming disabled, `requestUrl()`: https://www.dropbox.com/scl/fi/37uxlb5nj5rjnch8f25up/no-cors-no-streaming.mov?rlkey=b2qk6k009xx02lnlglzf3u84s&st=w12o50ri&dl=0
- VectorDB test: https://www.dropbox.com/scl/fi/8gwjns9s6d3ckvf2aict6/vector-DB.mov?rlkey=b1b3tp8k7kmu7l7e073h74xw7&st=kq7cg458&dl=0

2. Test it on the real device, I used an iPad Pro:
- Build the plugin `npm run build`.
- Create a new vault on the mobile, storing it in iCloud.
- Copy the plugin on mobile manually, you can do that copying `main.js`, `manifest.json` and `style.css` to the `.obsidian/plugins/copilot` directory in iCloud (I did it from the terminal because `.obsidian` directory is not visible in the Finder UI).

- Mobile test: https://www.dropbox.com/scl/fi/28jxahj1vv64saqu8pgyx/mobile-test.MP4?rlkey=z4m6szqginb0aynofkwrermen&st=gisinjji&dl=0


src/proxyServer.ts
```diff
@@ -1,101 +0,0 @@
-import cors from "@koa/cors";
-import Koa from "koa";
-import proxy from "koa-proxies";
-import { CopilotSettings } from "@/settings/SettingsPage";
-import { ChatModelDisplayNames } from "@/constants";
-
-// There should only be 1 running proxy server at a time so keep it in upper scope
-let server: any;
-
-export class ProxyServer {
```
gianluca-venturini, line 11: @logancyang I'm wondering if we can remove this ProxyServer entirely and just make calls directly to the right backend. What was the original reason for having this proxy?

logancyang:
Hey @gianluca-venturini thanks for the PR! Mobile support is going to be big if we can deliver it correctly.

Did you test this change on mobile? Does it work? It would be helpful if you describe in detail how you test, so we can use the same process.

src/proxyServer.ts
logancyang: @gianluca-venturini this is for some model providers such as Claude and all those 3rd party OpenAI replacements who don't allow CORS. Without this workaround they result in CORS errors in Obsidian.

gianluca-venturini: Thanks @logancyang for adding some context here. I found out that we can use the function [requestUrl()](https://docs.obsidian.md/Reference/TypeScript+API/requestUrl) instead of implementing a Proxy, I've an initial implementation here [acd0481](https://github.com/logancyang/obsidian-copilot/pull/513/commits/acd04819e26a02241adf333dba6d5f11facf2dbb). For now I have it working with `ChatAnthropic`, but we can easily replace all the calls made inside the different models.
WDYT?

gianluca-venturini:
> Hey @gianluca-venturini thanks for the PR! Mobile support is going to be big if we can deliver it correctly.
> 
> Did you test this change on mobile? Does it work? It would be helpful if you describe in detail how you test, so we can use the same process.

I tested the change on an iPad Pro, added the descriptions for how to test to the PR description. I also added a couple of recordings that show the manual tests I did.

src/components/Chat.tsx
```diff
@@ -316,9 +315,10 @@ const Chat: React.FC<ChatProps> = ({
   };
 
   useEffect(() => {
-    async function handleSelection(selectedText: string) {
-      const wordCount = selectedText.split(' ').length;
-      const tokenCount = await chainManager.chatModelManager.countTokens(selectedText);
+    async function handleSelection(event: CustomEvent) {
+      console.log('handleSelection1', event);
```
logancyang: Don't forget to remove console.logs once you are confident it works, same for the ones below.

src/langchainWrappers.ts
```diff
@@ -27,3 +30,53 @@ export class ProxyOpenAIEmbeddings extends OpenAIEmbeddings {
...
+    super(fields);
+
+    // Override the client with a custom one that uses myFetch
+    this.clientOptions.fetch = safeFetch as any;
+  }
+}
+
+async function safeFetch(url: string, options: RequestInit): Promise<Response> {
+  // Necessary to remove 'content-length' in order to make headers compatible with requestUrl()
+  delete (options.headers as Record<string, string>)['content-length'];
```
logancyang: What's the implication of removing this? Do we lose functionalities here?

logancyang:
> > Hey @gianluca-venturini thanks for the PR! Mobile support is going to be big if we can deliver it correctly.
> > Did you test this change on mobile? Does it work? It would be helpful if you describe in detail how you test, so we can use the same process.
> 
> I tested the change on an iPad Pro, added the descriptions for how to test to the PR description. I also added a couple of recordings that show the manual tests I did.

@gianluca-venturini the recordings are 🔥, I didn't know how `requestUrl` can support streaming, and this looks great!

My main question is how we can enable mobile and also avoid CORS errors for all providers. The proxy mode is actually just added in the previous release from a community contribution. If we can make sure there are no CORS errors for all OpenAI replacement providers, the local proxy mode may be removed completely. Is the solution to use the same logic from your `ChatAnthropicWrapper` in `ProxyChatOpenAI`?

Also, do you see a way forward with mobile mode as a toggle in the setting? When the toggle's off, we should expect it to behave like before, so that we limit the risk of breaking things in the next release.

Btw, do you know why the lists in your recordings are center-aligned? Just curious.

logancyang:
@gianluca-venturini Oh I misunderstood, from the recording I thought it was streaming 🤔, how did it appear that way? Perhaps I'm missing something.

Using polyfills for mobile could be a solution. As long as we can keep the new stuff incremental and avoid breaking old stuff, it sounds good to me.

gianluca-venturini:
> @gianluca-venturini Oh I misunderstood, from the recording I thought it was streaming 🤔, how did it appear that way? Perhaps I'm missing something.
> 
> Using polyfills for mobile could be a solution. As long as we can keep the new stuff incremental and avoid breaking old stuff, it sounds good to me.

@logancyang The only model that was using `requestUrl()` and therefore not streaming the result back in the recordings is Claude. I didn't have any problems with other models in the test.

Unfortunately on mobile since we need to support CORS, some LLM providers don't set the CORS headers and we can't spin up a Proxy (since there's no Nodejs server support), we're only left with the `request()` and `requestUrl()` functions, that don't seem to support streaming.

I opened a feature request with the Obsidian team in order to support streaming https://forum.obsidian.md/t/support-streaming-the-response-body-in-request-and-requesturl-apis/87381, I believe it's a great addition to the Obsidian javascript APIs since many LLMs return long responses and streaming to the UI seems the de-facto industry standard.

In order to move this change forward and support mobile we have few options:
1. We use `requestUrl()` for all the models that don't support CORS requests, in my test only `Claude` needs it. Once Obsidian introduces the streaming response feature we can bring them back to feature parity.
2. We can attempt to polyfill the node externals on mobile, paying attention that the plugin doesn't crash.

My preference is on (1) because will remove the need for a Proxy completely and the dependency on Nodejs, keeping the build simpler. I didn't have much success with (2) yet, that's because the build is made only once for both mobile and desktop, there's no conditional build that can polyfill just on mobile. If I leave the Proxy (`src/proxyServer.ts`), when launching the plugin on mobile, it crashes immediately after importing the module, since it tries to import Nodejs dependencies that are not present.

logancyang:
@gianluca-venturini I completely agree that `requestUrl` is the correct way to proceed given our situation and Obsidian's official recommendation. A hacky way for (2) doesn't sound like a clean solution and requires a lot of efforts that may be wasted in the future. I marked your feature request and will ask whoever is interested to upvote.

Let's suppose `requestUrl` supports streaming now, it seems we need to use it to override the `_call` method by wrapping every `Chat<Provider>` module we are currently using? 

Also, when I pulled the PR trying to test, I see an error about PouchDB's event emitter messing with the build: 

```
Could not resolve "events"

    node_modules/pouchdb/lib/index-browser.es.js:5:15:
      5 │ import EE from 'events';
```

I added polyfills for it and Chat is working for me on iPhone. But does it mean that PouchDB is also relying on Node so we need to get rid of it for mobile?

gianluca-venturini:
> @gianluca-venturini I completely agree that `requestUrl` is the correct way to proceed given our situation and Obsidian's official recommendation. A hacky way for (2) doesn't sound like a clean solution and requires a lot of efforts that may be wasted in the future. I marked your feature request and will ask whoever is interested to upvote.
> 
> Let's suppose `requestUrl` supports streaming now, it seems we need to use it to override the `_call` method by wrapping every `Chat<Provider>` module we are currently using?

We can just wrap the Chat<Provider> that need to support CORS, no need to wrap all of them.

LMStudio has an option to enable CORS headers while starting the server, while Ollama works natively without any issue, so the only one remained seems to be Claude. I discovered this method here a couple of days ago: https://simonwillison.net/2024/Aug/23/anthropic-dangerous-direct-browser-access, I'll test itnext.

> 
> Also, when I pulled the PR trying to test, I see an error about PouchDB's event emitter messing with the build:
> 
> ```
> Could not resolve "events"
> 
>     node_modules/pouchdb/lib/index-browser.es.js:5:15:
>       5 │ import EE from 'events';
> ```
> 
> I added polyfills for it and Chat is working for me on iPhone. But does it mean that PouchDB is also relying on Node so we need to get rid of it for mobile?

I haven't seen this error before, it seems designed for storing data inside the browser, so probably there's some way to make it work well without changes https://pouchdb.com, I can explore that once after the CORS issue.

gianluca-venturini:
> > @gianluca-venturini I completely agree that `requestUrl` is the correct way to proceed given our situation and Obsidian's official recommendation. A hacky way for (2) doesn't sound like a clean solution and requires a lot of efforts that may be wasted in the future. I marked your feature request and will ask whoever is interested to upvote.
> > Let's suppose `requestUrl` supports streaming now, it seems we need to use it to override the `_call` method by wrapping every `Chat<Provider>` module we are currently using?
> 
> We can just wrap the Chat that need to support CORS, no need to wrap all of them.
> 
> LMStudio has an option to enable CORS headers while starting the server, while Ollama works natively without any issue, so the only one remained seems to be Claude. I discovered this method here a couple of days ago: https://simonwillison.net/2024/Aug/23/anthropic-dangerous-direct-browser-access, I'll test it next.

The header that Anthropic implemented can be added directly to the fetch request, that maintains streaming support, same as before.

@logancyang what's the easiest way to test the openAI compatible APIs? Maybe we can get away adding the `requestUrl()` and loosing streaming (until Obsidian implements our feature request) only to those if the user is activating the CORS mode. 

> 
> > Also, when I pulled the PR trying to test, I see an error about PouchDB's event emitter messing with the build:
> > ```
> > Could not resolve "events"
> > 
> >     node_modules/pouchdb/lib/index-browser.es.js:5:15:
> >       5 │ import EE from 'events';
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I added polyfills for it and Chat is working for me on iPhone. But does it mean that PouchDB is also relying on Node so we need to get rid of it for mobile?
> 
> I haven't seen this error before, it seems designed for storing data inside the browser, so probably there's some way to make it work well without changes https://pouchdb.com, I can explore that once after the CORS issue.

I replaced `pouchdb` with `pouchdb-browser`, hopefully it's compatible and doesn't break the plugin on mobile. What's the easiest way to test if I broke the Vector DB?

gianluca-venturini:
> Btw, do you know why the lists in your recordings are center-aligned? Just curious.

This turned out to be the same class name used by obsidian, that align center on mobile, I just renamed the class and now it stays left aligned.

logancyang:
> Maybe we can get away adding the requestUrl() and loosing streaming (until Obsidian implements our feature request) only to those if the user is activating the CORS mode.

@gianluca-venturini So basically it means we *get rid of the local proxy completely*, add a "CORS mode" where we switch to `requestUrl()` and lose streaming. Do I understand correctly?

BTW, I also found out about the Anthropic header thing today! We discovered it at the same time 😁 I'm cooking up this PR #567 and it will add the header for Anthropic, it won't need the local proxy anymore 🎉

Sorry I've been making a ton of changes, your PR will have many conflicts to resolve when rebased 🥶

logancyang:
> I replaced pouchdb with pouchdb-browser, hopefully it's compatible and doesn't break the plugin on mobile. What's the easiest way to test if I broke the Vector DB?

Custom prompts also use pouchdb, so you could test adding some custom prompts. Or you could switch to Vault QA mode and see if it can index correctly without error.

logancyang:
@gianluca-venturini Did Claude work on mobile with the "dangerously-allow-browser" header without `requestUrl`? I haven't tested on mobile but can confirm it works well on desktop.

With the model revamp #567, my idea is that the CORS column sets a flag that should switch to `requestUrl` for the request. However I haven't been able to override langchain's `ChatOpenAI` correctly using `requestUrl` (it will be the primary thing to change because CORS error happens primarily with 3rd party OpenAI format APIs.)

I missed your previous question:

> what's the easiest way to test the openAI compatible APIs?

I mainly tested using OpenRouter but it did not have CORS issue, so we need to find a 3rd party OpenAI format API that has CORS error. People mentioned some in Issues like #374, so Huggingface Inference API could be a candidate, I haven't tried yet.

gianluca-venturini:
> > Maybe we can get away adding the requestUrl() and loosing streaming (until Obsidian implements our feature request) only to those if the user is activating the CORS mode.
> 
> @gianluca-venturini So basically it means we _get rid of the local proxy completely_, add a "CORS mode" where we switch to `requestUrl()` and lose streaming. Do I understand correctly?

That's right.

> 
> BTW, I also found out about the Anthropic header thing today! We discovered it at the same time 😁 I'm cooking up this PR #567 and it will add the header for Anthropic, it won't need the local proxy anymore 🎉 I will also try out "CORS mode" with `requestUrl` and eliminate local proxy if that sounds good.
> 
> I'm so sorry I've been making a ton of changes, your PR will have many conflicts to resolve when rebased 🥶 So it would be wise to keep the changes as minimal as possible to avoid headache.

No worries, wasn't that bad to refactor so far.

logancyang:
Oh it just occurred to me, we could just switch off CORS in LM Studio server and use it to test the `requestUrl` approach.

gianluca-venturini:
> @gianluca-venturini Did Claude work on mobile with the "dangerously-allow-browser" header without `requestUrl`? I haven't tested on mobile but can confirm it works well on desktop.

Yes this is working well, we don't need to Proxy Claude anymore.

gianluca-venturini:
> Oh it just occurred to me, we could just switch off CORS in LM Studio server and use it to test the `requestUrl` approach. Similarly, start `ollama serve` without `OLLAMA_ORIGINS` is another way.

That's right. I just updated the test videos in the PR description, where I'm using exactly this method to test the proxy. I tested both with CORS and without CORS, if CORS headers are off it's using the `requestUrl()` and loosing streaming.

esbuild.config.mjs
```diff
@@ -41,6 +39,12 @@ const context = await esbuild.context({
   treeShaking: true,
   outfile: "main.js",
   plugins: [svgPlugin(), wasmPlugin],
+  define: {
+    global: "window",
+  },
+  alias: {
+    events: "./null-events.js",
```
gianluca-venturini: I needed to add this in order to make `PouchDB` work correctly. Maybe we can switch out to another DB that doesn't rely on `events` internally and clean this up.

gianluca-venturini:
@logancyang ok all checks are now green and we agree on the implementation at high level. Time to request a deeper code review. You can look at the videos in the PR description to see how I tested it.

Let me know if you have more ideas of features we should be testing before merging to convince ourselves it's working correctly.

esbuild.config.mjs
logancyang: Good to know that this makes pouchdb work on mobile! And yes, I've been thinking about removing pouchdb. 

1. Custom prompt: use data.json directly for better obsidian sync support
2. Vector store: this is my main question. Do you know any browser-native vector db? Right now pouchdb doesn't provide any advantage on vector similarity search. Recommendations welcome!

src/LLMProviders/chatModelManager.ts
```diff
@@ -96,6 +95,7 @@ export default class ChatModelManager {
       [ModelProviders.LM_STUDIO]: {
         openAIApiKey: "placeholder",
         openAIProxyBaseUrl: `${params.lmStudioBaseUrl}`,
+        useOpenAILocalProxy: params.useOpenAILocalProxy,
```
logancyang: Why do we need this for LM Studio? Is this line for testing?

src/LLMProviders/chatModelManager.ts
```diff
@@ -73,7 +72,7 @@ export default class ChatModelManager {
         openAIProxyBaseUrl: params.openAIProxyBaseUrl,
       },
       [ModelProviders.ANTHROPIC]: {
-        anthropicApiUrl: `http://localhost:${PROXY_SERVER_PORT}`,
+        anthropicApiUrl: "https://api.anthropic.com",
```
logancyang: I think you can remove this line? ChatAnthropic should default to this.

src/LLMProviders/embeddingManager.ts
```diff
@@ -46,6 +46,7 @@ export default class EmbeddingManager {
       embeddingModel,
       openAIEmbeddingProxyBaseUrl,
       openAIEmbeddingProxyModelName,
+      useOpenAILocalProxy,
```
logancyang: ~~Aren't we removing the local proxy in this PR? I thought this flag was deprecated because of that?~~

Oh I see, `useOpenAILocalProxy` now indicates using `safeFetch` i.e. `requestUrl`. Should we give it a better name? Like `enableCors` or something since local proxy is actually removed.

src/components/ChatComponents/ChatIcons.tsx
```diff
@@ -188,15 +153,15 @@ const ChatIcons: React.FC<ChatIconsProps> = ({
           <span className="tooltip-text">Model Selection</span>
         </div>
       </div>
-      <button className="chat-icon-button" onClick={onNewChat}>
+      <button className="chat-icon-button clickable-icon" onClick={onNewChat}>
```
logancyang, line 156: I don't see how `clickable-icon` is used? Maybe I'm missing something.

src/components/Chat.tsx
```diff
@@ -310,9 +313,9 @@ const Chat: React.FC<ChatProps> = ({
   };
 
   useEffect(() => {
-    async function handleSelection(selectedText: string) {
-      const wordCount = selectedText.split(" ").length;
-      const tokenCount = await chainManager.chatModelManager.countTokens(selectedText);
+    async function handleSelection(event: CustomEvent) {
```
logancyang, line 312: Nice! As a note: we just need to make sure text selection commands/prompts are working

src/langchainWrappers.ts
```diff
@@ -12,6 +14,7 @@ export class ProxyChatOpenAI extends ChatOpenAI {
     this["client"] = new OpenAI({
       ...this["clientConfig"],
       baseURL: fields.openAIProxyBaseUrl,
+      fetch: fields.useOpenAILocalProxy ? safeFetch : undefined,
```
logancyang: Ditto, maybe call it `enableCors`

src/langchainWrappers.ts
```diff
@@ -24,6 +27,72 @@ export class ProxyOpenAIEmbeddings extends OpenAIEmbeddings {
     this["client"] = new OpenAI({
       ...this["clientConfig"],
       baseURL: fields.openAIEmbeddingProxyBaseUrl,
+      fetch: fields.useOpenAILocalProxy ? safeFetch : undefined,
```
logancyang: Since embedding endpoints don't need streaming, perhaps we should default to `safeFetch`? WDYT?

logancyang:
@gianluca-venturini Thanks for the update 🙏 Left some comments and mostly lgtm! I'll pull it and test manually as well. 

Manual tests should include:

1. Any command or custom prompt that relies on text selection
2. Add/edit/apply/delete custom prompt
3. Vector indexing (by clicking the Refresh Index button or trigger by mode switch to Vault QA)
4. Around `safeFetch`
    1. ChatOpenAI with `safeFetch` on desktop and mobile, for both normal OpenAI API call and a 3rd party call, like LM Studio with CORS off (I see you may have done this)
    2. ChatAnthropic with `safeFetch` on desktop and mobile
    3. Embedding calls with `safeFetch` on desktop and mobile

These are mostly on my mind now; there could be more. Just FYI, my other PR #567 is making some big changes to how models are used, and it will probably land after yours to use `safeFetch`.

I agree that we may get rid of PouchDB eventually. I'm searching for an in-browser vector db with a TS client, let me know if you have one in mind!

src/components/ChatComponents/ChatIcons.tsx
gianluca-venturini: That's to prevent the buttons from having extra padding. It's defined in the obsidian `app.css`:
image

src/components/Chat.tsx
gianluca-venturini: I tested it in this way:
https://www.dropbox.com/scl/fi/vgpcjlwig34xzz3t96wz7/custom-prompt.mov?rlkey=f4gfslsf8vx5wf2i4uczub5dw&st=o1ju6xno&dl=0

src/LLMProviders/chatModelManager.ts
gianluca-venturini: I added it here in order to allow using LM Studio with CORS disabled to test `ProxyChatOpenAI`:
image

Is there any other `providerConfig` that should have it enabled?

src/LLMProviders/chatModelManager.ts
gianluca-venturini: I removed and tested and it works correctly, thanks!

src/LLMProviders/embeddingManager.ts
gianluca-venturini: Definitely, I renamed it.

src/langchainWrappers.ts
gianluca-venturini: Yes, it's not only using `safeFetch()`, I tested it adding `openAIEmbeddingProxyBaseUrl = 'http://localhost:1234/v1'` and using LMStudio with LLama 3 and it's re-endexing all the embeddings when I force re-index.

gianluca-venturini:
> @gianluca-venturini Thanks for the update 🙏 Left some comments and mostly lgtm! I'll pull it and test manually as well.
> 
> Manual tests should include:
> 
> 1. Any command or custom prompt that relies on text selection
> 2. Add/edit/apply/delete custom prompt
> 3. Vector indexing (by clicking the Refresh Index button or trigger by mode switch to Vault QA)
> 4. Some simple chat in Vault QA mode to ensure the QA chain works.
> 5. Around `safeFetch`
>    
>    1. ChatOpenAI with `safeFetch` on desktop and mobile, for both normal OpenAI API call and a 3rd party call, like LM Studio with CORS off (I see you may have done this)
>    2. ChatAnthropic with `safeFetch` on desktop and mobile
>    3. Embedding calls with `safeFetch` on desktop and mobile
> 
> These are mostly on my mind now; there could be more. Just FYI, my other PR #567 is making some big changes to how models are used, and it will probably land after yours to use `safeFetch`.

Thanks for listing the tests, it passed all of them on both desktop (with mobile emulation) and mobile except for "ChatAnthropic with `safeFetch`", since anthropic doesn't need `safeFetch()` anymore since we're now using the header.

Maybe you can take an extra pass of manual tests on my PR in order to be really sure I haven't missed anything.

gianluca-venturini:
> I agree that we may get rid of PouchDB eventually. I'm searching for an in-browser vector db with a TS client, let me know if you have one in mind!

I did some research and there's no yet a library out there that has a large adoption and a community that maintains it. I've seen many wrappers around `IndexedDB` (like https://github.com/yusufhilmi/client-vector-search or https://github.com/nitaiaharoni1/vector-storage). Maybe we can set aside this problem and leave PouchDB for now, until the ecosystem matures a little and one of the libraries has a strong community that is likely to maintain it long term.

rygwdn:
> there's no yet a library out there that has a large adoption and a community that maintains it

I'm new to this space, but would https://github.com/askorama/orama be a good fit? Seems to be browser compatible and have vector search & full text search built in. It does seem like it's fully in-memory though, so I'm not sure if that's a problem.

logancyang:
@rygwdn Thanks for the suggestion! In-memory sounds fine to me. We could dump and load the embeddings to/from a flat file and use this in-memory search. Will take a closer look!

logancyang:
@gianluca-venturini Awesome, let me do some testing, if everything looks good I'll merge this PR.

Once this PR is merged, every change must be mobile-friendly from this point on. Do you think if there are additional checks we can add to the github actions? Just to help make sure a future build won't accidentally break mobile?

src/components/Chat.tsx
```diff
@@ -341,8 +344,10 @@ const Chat: React.FC<ChatProps> = ({
         isVisible = false,
         ignoreSystemMessage = true, // Ignore system message by default for commands
       } = options;
-      const handleSelection = async (selectedText: string, eventSubtype?: string) => {
-        const messageWithPrompt = await promptFn(selectedText, eventSubtype);
+      const handleSelection = async (event: CustomEvent) => {
+        console.log('handleSelection2', event);
```
logancyang: We can remove these console logs now

src/components/ChatComponents/ChatIcons.tsx
logancyang: Was the "extra padding" looking weird on mobile? because with this change it looks weirder to me, they are now big icons and don't share the consistent button look with others.

Now: 
SCR-20240829-ooes

Before (looks better IMO):
SCR-20240829-ooiv

src/aiParams.ts
```diff
@@ -59,7 +59,7 @@ export interface LangChainParams {
   openRouterModel: string;
   lmStudioBaseUrl: string;
   openAIProxyBaseUrl?: string;
-  useOpenAILocalProxy?: boolean;
+  enableCors?: boolean;
```
logancyang, line 62: Personally I'd prefer having this in CopilotSettings instead of LangChainParams since it has nothing to do with Langchain. But no worries, it will be address by my following PR. You can leave it here.

logancyang:
I did some manual testing and it generally looks good! Everything seems to be working. My only issue was the UI on mobile. This plugin hasn't done any UI design for a smaller screen like the phone, so the experience is not as good.

When it's opened in the note area, the input area is cutoff at the bottom. Just as a note. We can make updates in future PRs.

IMG_4622
IMG_4623

Anyways, it's pretty awesome that it's working!!!

I left a few more comments and questions, and then we are good to go! 🚀

src/aiParams.ts
gianluca-venturini: It's here because it's needed while constructing `ProxyOpenAIEmbeddings` in `EmbeddingManager`.

Are you considering adding the settings here? https://github.com/gianluca-venturini/obsidian-copilot/blob/ed96670e84aa2ab0243f1bdd81691fa30c767633/src/LLMProviders/embeddingManager.ts#L23

gianluca-venturini:
> I did some manual testing and it generally looks good! Everything seems to be working. My only issue was the UI on mobile. This plugin hasn't done any UI design for a smaller screen like the phone, so the experience is not as good.
> 
> When it's opened in the note area, the input area is cutoff at the bottom. Just as a note. We can make updates in future PRs.
> 
> IMG_4622 IMG_4623
> Anyways, it's pretty awesome that it's working!!!
> 
> I left a few more comments and questions, and then we are good to go! 🚀

Definitely agree, I didn't change anything in the style of those buttons and they don't look as good as desktop. Agree that should be a fast follow.

src/aiParams.ts
logancyang: @gianluca-venturini yes, in #567 I pass an array of `CustomModel` objects where each model has its own `enableCors` that the user can set in a table. I'll land your PR first and mine later.

logancyang:
@gianluca-venturini Congrats on reaching this huge milestone 🚀 and thank you again for your contribution! I think it's ready to merge. If you think it's ready just give a thumbs up and I'll merge it.

PR #567, opened 2024-08-27
[Model Revamp] Implement Cursor style model settings

logancyang:
- Dynamic list of models
- Add/delete custom chat models by filling in the form of model name, provider, API base URL, API key. No need for separate OpenRouter, LM Studio parameters internally. All 3rd party API with OpenAI format can be added this way
- Same for embedding models

Each model is a CustomModel

```js
export interface CustomModel {
  name: string;
  provider: string;
  baseUrl?: string;
  apiKey?: string;
  enabled: boolean;
  isBuiltIn?: boolean;
  enableCors?: boolean;
}
```

SCR-20240827-mvjr


logancyang:
Note: this PR also completely updates `ChatModelManger`. Claude models won't need a local proxy from now on because of this additional header:

```js
[ChatModelProviders.ANTHROPIC]: {
        anthropicApiKey: decrypt(params.anthropicApiKey),
        modelName: params.anthropicModel,
        clientOptions: {
          defaultHeaders: {
            // Required for Anthropic models to work without CORS error
            "anthropic-dangerous-direct-browser-access": "true",
          },
        },
      },
```

ijwfly:
Maybe it is a good idea to let users disable builtin models (and maybe even remove them).

logancyang:
@ijwfly yeah that's coming next.

Issue #568, opened 2024-08-27
[FR] Add a customizable prompt for "send notes to prompt" button


h4c5:
**Is your feature request related to a problem? Please describe.**
It would be great to allow the user to customize the prompt used by the "send notes to prompt" button.
It is particularly useful to use a different language in this prompt.

**Describe the solution you'd like**
Add a customizable prompt in the settings and use it in the `sendNotesContentPrompt` function.

**Describe alternatives you've considered**
One can rely on custom prompt template but is much less handy than "send notes to prompt" button.

**Additional context**

- Send Notes Prompt Configuration
  ![image](https://github.com/user-attachments/assets/d411f9c9-653c-45db-8a08-ead953686345)

- Debug view of the custom prompt
  ![image](https://github.com/user-attachments/assets/62b27eb3-5799-4d1a-9122-2a79b675f6f9)

logancyang:
As discussed, closing this one now. Thanks for the PR though.

PR #569, opened 2024-08-27
Add user custom "send notes to prompt" prompt

h4c5:
## Add a customizable prompt for "send notes to prompt" button #568 

This PR allow the user to customize the prompt of the "send notes to prompt" button


logancyang:
@h4c5  Thanks for the PR. Actually I plan to deprecate this button. User can already pass the note by directly `[[note title]]` in the chat.

h4c5:
Oh I see. This button is convenient in my opinion but I understand your point of view.

logancyang:
@h4c5 IMO this button's strength was used with copilot command "set note context in chat mode" where you can specify by path or tags. But I'm going enable "ad-hoc custom prompt" inside chat directly next, so that even this use case will be covered. Thanks for the idea anyway!

logancyang:
Closing as not planned

PR #572, opened 2024-08-29
fix: add a new setting to ignore certain folders while indexing

milesvdw:
Pretty much in the title. 
- add a new setting allowing users to configure a comma separated list of folder paths
- if a file's path starts with any of those folder paths, ignore it during indexing

This is important because sometimes I have folders with really eye-catching garbage that the AI _reallllly_ wants to use (like, templates that _look_ very similar to many other files, but don't actually have any content in them) and I don't want the Vault RAG to pick those up.


logancyang:
This is an existing command

SCR-20240829-nppr
SCR-20240829-npmm

There's no centralized documentation now and that's on me.

If you'd like it to be a setting instead of a command, you might have to wait until I release mobile support and model revamp.

milesvdw:
I'll likely just run what I've got locally for now, and when the next release comes I'll circle back. I didn't realize there was a command for it... I was poking around in the settings because that made the most sense to me... How could I see which folders I'd already excluded? I do definitely think it should be visible/persisted somewhere. The other LLM/vectorization/RAG plugins that I tried (I like yours the best) all had a similar setting.

src/constants.ts
```diff
@@ -196,6 +196,7 @@ export const DEFAULT_SETTINGS: CopilotSettings = {
   lmStudioBaseUrl: "http://localhost:1234/v1",
   stream: true,
   defaultSaveFolder: "copilot-conversations",
+  excludedFolders: [],
   indexVaultToVectorStore: VAULT_VECTOR_STORE_STRATEGY.ON_MODE_SWITCH,
   qaExclusionPaths: "",
```
logancyang: It is right here. @milesvdw

src/constants.ts
```diff
@@ -196,6 +196,7 @@ export const DEFAULT_SETTINGS: CopilotSettings = {
   lmStudioBaseUrl: "http://localhost:1234/v1",
   stream: true,
   defaultSaveFolder: "copilot-conversations",
+  excludedFolders: [],
```
logancyang: No need for this new setting.

logancyang:
> I'll likely just run what I've got locally for now, and when the next release comes I'll circle back. I didn't realize there was a command for it... I was poking around in the settings because that made the most sense to me... How could I see which folders I'd already excluded? I do definitely think it should be visible/persisted somewhere. The other LLM/vectorization/RAG plugins that I tried (I like yours the best) all had a similar setting.

The setting exists and it's set via the command. You can check by opening the command modal, the field shows existing paths you set.

logancyang (changes requested):
Can you use the existing setting field, and move the new setting to the QA setting block. Then it should be good to go.

milesvdw:
> Can you use the existing setting field, and move the new setting to the QA setting block. Then it should be good to go.

@logancyang thanks! I made the updates/simplifications as requested. Hopefully this setting being visible is helpful for other people! Loving the plugin, I use it every day.

logancyang:
Can you fix the build error. I recommend testing with `npm run build`, as well as `npm run dev` and manual checks in the app before submitting a PR.

milesvdw:
> Can you fix the build error. I recommend testing with `npm run build`, as well as `npm run dev` and manual checks in the app before submitting a PR.

@logancyang ah yes, apologies for the lax QA. I didn't realize you were storing your folders as a string instead of the list of strings I was using. All fixed now, and it does appear to be working with the current build in my local repo.

milesvdw:
> Owner
> ###

@logancyang ah yes, apologies for the lax QA. I didn't realize you were storing your folders as a string instead of the list of strings I was using. All fixed now, and it does appear to be working with the current build in my local repo.

milesvdw:
> Can you fix the build error. I recommend testing with `npm run build`, as well as `npm run dev` and manual checks in the app before submitting a PR.

@logancyang ah yes, apologies for the lax QA. I didn't realize you were storing your folders as a string instead of the list of strings I was using. All fixed now, and it does appear to be working with the current build in my local repo.

milesvdw:
> Can you fix the build error. I recommend testing with `npm run build`, as well as `npm run dev` and manual checks in the app before submitting a PR.

@logancyang ah yes, apologies for the lax QA. I didn't realize you were storing your folders as a string instead of the list of strings I was using. All fixed now, and it does appear to be working with the current build in my local repo.

src/main.ts
```diff
@@ -545,6 +545,7 @@ export default class CopilotPlugin extends Plugin {
           content: fileContents[index],
           metadata: fileMetadatas[index]?.frontmatter ?? {},
         };
+
```
logancyang: No need to change this file?

src/settings/components/GeneralSettings.tsx
```diff
@@ -79,6 +79,13 @@ const GeneralSettings: React.FC<GeneralSettingsProps> = ({
         value={settings.defaultSaveFolder}
         onChange={(value) => updateSettings({ defaultSaveFolder: value })}
       />
+      <TextComponent
```
logancyang: This should be in QA settings instead.

milesvdw: sure - I assume you'll move this around whenever the larger feature is out of QA?
also fixed a small typo I spotted in qa settings.

@logancyang anything else you need? I assume you'll just merge this whenever you feel like, I plan to forget about it (but will definitely use the UX feature when released with the plugin!)

logancyang (approved):

logancyang:
LGTM, thanks for the contribution!

Issue #510, opened 2024-08-15
Default cursor location is chat-input-container


frarinha:
**Is your feature request related to a problem? Please describe.**
The need to move the mouse and select the chatbox.

**Describe the solution you'd like**
Upon the toggle of the right window on obsidian, where the `copilot-chat-view` is, the default cursor location would be the `chat-input-container`. You could just toggle the window and immediately start interacting with the LLM.

Emt-lin:
@logancyang I try to fixed this bug, pls check it. thanks

logancyang:
@Emt-lin You are welcome to create a PR and I can take a look.

logancyang:
@frarinha @Emt-lin why do we have this issue again? When I toggle the chat window, the cursor is in the chatbox.

Emt-lin:
@logancyang 
- [x] case 1：When I click on the chat icon in the left sidebar, the cursor appears in the chat box.
- [ ] case 2：When I switch from the chat area to the notes area(or others) and then back to the chat area, the cursor isn't in the chat box.
- [ ] case 3：When I switch items in the right sidebar(eg: outline item and chat item), it's the same as in Case 2.

I'm guessing @frarinha's idea is these unfinished case above.

logancyang:
@Emt-lin how do you switch areas in case 2 and 3?

frarinha:
It's when you toggle (to expand and collapse) the obsidian's right side pane. The cursor is not on the chat input box. 
We can now add an hotkey to **Copilot: Toggle Copilot Chat Window**, but if the right side pane is in the collapsed state, you have toggle it twice, off and back on, and it will open the copilot chat window and the cursor will be there.

Emt-lin:
@frarinha because When you toggle **Copilot: Toggle Copilot Chat Window** , the chat window will be closed at first  if  **a chat window** exist.

```ts
  toggleView() {
    const leaves = this.app.workspace.getLeavesOfType(CHAT_VIEWTYPE);
    // here
    leaves.length > 0 ? this.deactivateView() : this.activateView();
  }
```

@logancyang

Emt-lin:
@logancyang sorry,something happened last weekend.
case 1、2、3 in order

![Kapture 2024-09-09 at 10 37 54](https://github.com/user-attachments/assets/f8f36db5-d4a7-4949-82b8-39ae7e7ac589)

logancyang:
[...]

PR #593, opened 2024-09-05
fix: Fix default cursor location is chat-input-container(#510)

Emt-lin:
fixed(#510): When I first clicked on `copilot chat container`, the default cursor will be location the `chat-input-container`


src/components/Chat.tsx
```diff
@@ -75,6 +73,20 @@ const Chat: React.FC<ChatProps> = ({
...
+      EVENT_NAMES.CHAT_IS_VISIBLE,
+      (evt: CustomEvent<{ chatIsVisible: boolean }>) => {
+        setChatIsVisible(evt.detail.chatIsVisible);
+      }
+    );
+  }, []);
+
+  useEffect(() => {
+    visibleCallback();
+  }, []);
```
logancyang, line 91: The event listener for EVENT_NAMES.CHAT_IS_VISIBLE is added in the useEffect hook but is never removed. This can lead to memory leaks if the Chat component is unmounted and remounted multiple times, as the event listener will persist and accumulate.

src/main.ts
```diff
@@ -80,6 +81,8 @@ export default class CopilotPlugin extends Plugin {
 
     this.registerView(CHAT_VIEWTYPE, (leaf: WorkspaceLeaf) => new CopilotView(leaf, this));
 
+    this.initActiveLeafChangeHandler();
```
logancyang, line 97: It registers an event listener for the "active-leaf-change" event but does not remove it during the plugin's onunload lifecycle. This can lead to memory leaks or multiple event handlers being registered if the plugin is reloaded or reinitialized multiple times.

logancyang:
Does it also mean the `getChatVisibility` method in CopilotView can be removed.

Emt-lin:
@logancyang Maybe it only slightly improves the user experience.
- Once the user has edited in the notes area, the user can quickly ask gpt questions

logancyang (approved):
LGTM. Could you rebase to new master, I'll merge then.

Emt-lin:
@logancyang I already merged the latested commit.

Issue #597, opened 2024-09-05
Cohere's Command-R and Command-R+ support


o-mikhailovskii:
**Is your feature request related to a problem? Please describe.**
Although Cohere's API is used for embeddings and I found a request #471 for such a support closed, I was unable to use Cohere's LLMs.

**Describe the solution you'd like**
Here is my proposed patch below from version 2.6.2. I am a newbie in TypeScript and did it how I could, so I am not shure that everything is correct. To ensure the extension is properly built I needed to update the Cohere library as well as change the target build from "es2018" to "es2020". I have not included the lock-file so that the changes more easily understandable.

[cohere_no_lock.patch](https://github.com/user-attachments/files/16889803/cohere_no_lock.patch)

logancyang:
Looks like cohere doesn't conform to openai api format. If you open a PR I can take a look.

PR #619, opened 2024-09-08
Add support for Cohere's Command-R and Command-R+ models

o-mikhailovskii:
As proposed in #597, here is the PR.


logancyang:
@o-mikhailovskii thanks for the PR! We probably need to move the Cohere API key setting from QA setting up to API settings then.

logancyang:
I pulled this PR and tested, it doesn't work for me. Does it work for you?

SCR-20240908-mqec

o-mikhailovskii:
Hey! I faced this kind of issue when trying to run with the older version of Cohere's package (and presumably the corresponding Langchain's package). That's why I updated the dependencies in the `package.json`. Have you updated the installation of the two packages?

o-mikhailovskii:
Answering your question - yes, I am able to select and use Cohere's models.

logancyang:
@o-mikhailovskii oh yes my bad, didn't update the package. Now it works, thanks for the contribution! I'll move the setting up myself later no worries. Merged.

PR #626, opened 2024-09-09
Custom prompt support in chat

### LINKED ISSUE #590: Alternative method for interacting with custom prompts
I've kicked the tires on various GenAI implementation plug-ins in Obsidian. One of the better patterns for managing custom prompts is to configure a folder containing them. This allows for a single source of prompts to be used across multiple GenAI integrations (including outside of Obsidian via the .md files on disk). 

Perhaps should be a separate issue,  but then being able to interface with the prompts, through the copilot chat, via tab-completable @PROMPTNAME. So when you're chatting with CoPilot directly, you can invoke a prompt easily.

### LINKED ISSUE #623: Duplication of Sonnet model
There seems to be a duplication of Anthropic's Sonnet model in the `constants.ts`.

Here are the lines:
https://github.com/logancyang/obsidian-copilot/blob/8a4bd04b6680ca6dae2a77148f7343a2eeab59df/src/constants.ts#L57C1-L67C5

**Expected behavior**
 Shouldn't one of the models be Haiku?

### LINKED ISSUE #624: Custom prompt suggestion
I want to refer to custom prompts in chat and not on file level

I want to save custom prompts in a markdown file and refer to them in chat using @ symbol for example

using already existing custom prompt, but every time I have to copy the text in a temporary note and run custom prompt on selection.

**Additional context**
given I have a saved custom prompt in markdown for example `create_summary` I want to refer to it in chat using @create_summary of [[note]] or copied text

logancyang:
Fixes #590
Fixes #624 
Fixes #623

- Type `/` and it brings up the custom prompts list, selecting one fills it into the chat input.
- `{activeNote}` is supported now
- When `{}` is present but no text is selected, fallback to the entire active note.

### DIFF
--- a/src/components/AddPromptModal.tsx
+++ b/src/components/AddPromptModal.tsx
@@ -43,6 +43,8 @@ export class AddPromptModal extends Modal {
       frag.createEl("br");
       frag.createEl("strong", { text: "- {[[Note Title]]} represents a note. " });
       frag.createEl("br");
+      frag.createEl("strong", { text: "- {activeNote} represents the active note. " });
+      frag.createEl("br");
       frag.createEl("strong", { text: "- {FolderPath} represents a folder of notes. " });
       frag.createEl("br");
       frag.createEl("strong", {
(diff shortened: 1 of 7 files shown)


Emt-lin:
@logancyang I think this way of implementing custom prompts is a bit intrusive. I don't like having folders that don't belong to notes.

logancyang:
@Emt-lin you always have the option to set the folder to be hidden

Emt-lin:
@logancyang I can't find this option

logancyang:
@Emt-lin  just name your folder with `.` at the start.

Issue #279, opened 2024-02-04
[Feature] ability to rename prompts


crypdick:
**Is your feature request related to a problem? Please describe.**

It's possible to edit a prompt, but it is not possible to rename a prompt. You have to delete the note and create it again with a new name.

**Describe the solution you'd like**

Ability to rename a prompt in the edit page.

**Describe alternatives you've considered**

I tried to rename the prompts manually, but I can't find where they are actually stored.

logancyang:
Hi @crypdick, thanks for the FR. Currently it's a simple schema where id is the title, so you can't rename them and have to recreate them. Ofc adding an editable title is possible but it will require schema migration for all users, and it needs to make sure all titles are unique with separate logic. 

There are currently higher priority tasks so I'll come back to this one after. Putting this one up as a good first issue if anyone is interested.

nicolas-g:
+1

logancyang:
Now that custom prompts are in markdown, editing title works. Reopen it you see errors.

PR #631, opened 2024-09-10
feat: Support for renaming title of prompt.

Emt-lin:
Trying to address the problem that title of prompt can't change without altering the `CustomPrompt` structure.

I think the best solution is to change the `CustomPrompt` structure appropriately.
 
relation: #279


logancyang:
Seems you are updating an old version. The code you are changing no long exists, please pull the latest code.

src/main.ts
```diff
@@ -224,16 +224,35 @@ export default class CopilotPlugin extends Plugin {
               if (doc.prompt) {
                 new AddPromptModal(
                   this.app,
-                  (title: string, newPrompt: string) => {
-                    this.dbPrompts.put({
```
logancyang: Seems you have an old version.

logancyang (changes requested):
Seems you are updating an old version. The code you are changing no long exists, please pull the latest code.

Emt-lin:
this PR is canceled.

PR #651, opened 2024-09-18
Embeddings enhancements

o-mikhailovskii:
1. Added support for Google's embedding models.
2. Cohere's embeddings are currently broken since they require the inputType parameter (see the screenshot below). Now, there is a dropdown menu for making the selection.

input_type must be provided with the model


src/constants.ts
```diff
@@ -146,6 +155,13 @@ export const BUILTIN_EMBEDDING_MODELS: CustomModel[] = [
   },
 ];
 
+export const COHEREAI_EMBEDDING_INPUT_TYPES = [
```
logancyang: Should this be a user setting or set internally? I'd like to avoid cluttering the settings and overwhelming users with options.

o-mikhailovskii: You are right on one side. On the other side (at least in the current version of API), there are these four options, and that's it. Why would you need to make custom options?

However, when I dug into more details, a "best practice" usage of Cohere's API would be as follows:
1. Embed documents with the `search_document` input type,
2. Search over the document with the `search_query` input type.

This suggests that dynamically changing the input type throughout Langchain usage would be ideal. However, it seems this would require substantial changes in plugin's codebase.

What do you think would be the best way to move forward?

logancyang: I see. The use of langchain is for switching between providers easily. If it doesn't give us the convenience, it'll be counterproductive to adapt its code for individual providers. Since there's a best practice, does langchain cohere client already have something out of the box for us to use? If not, I'd rather not adapt to it here in Copilot.

That said, I'm considering providing Copilot exclusive embedding models next. The user won't need an API key, just call that embedding provider through my gateway. For that, I'm considering Cohere vs. Voyage embeddings. Right now I'm inclined to Voyage for its better MTEB ranking and goodies like rerank API with explanation. If Cohere actually performs better in practice, I'll add it there as the Copilot's default embedding model.

o-mikhailovskii: Ok. The issue with the `inputType` parameter appears to be addressed in `@langchain/cohere` versions later than 0.2.0 (rendering my corresponding commit obsolete). Updating this package forces a cascade of dependency updates, ultimately leading to the need for a more recent `langchain` package, such as version 0.3.0. Consequently, this results in changes to `import` statements.

logancyang:
Thanks for the update! @o-mikhailovskii

PR #659, opened 2024-09-19
Added ability to tag saved conversations

logicsec:
**Purpose**
This pull request adds the ability to tag conversations within Obsidian Copilot, allowing for easier referencing and analysis using Dataview.

**Motivation:** 
Currently, Obsidian Copilot doesn't offer a way to tag conversations, making it difficult to organize and retrieve specific conversations for analysis using Dataview. 

This feature would enable users to:
- Categorize conversations: Tag conversations based on topics, projects, or other relevant criteria.
- Filter conversations: Use Dataview queries to retrieve conversations based on specific tags.
- Analyze conversations: Gain insights from tagged conversations by using Dataview's powerful analysis capabilities.

**Changes:**
- Added a new settings called **Default Conversation Tag** which defaults to ai-conversation. 
- Updated SettingsPage.tsc CopilotSettings interface to reference the new defaultTag string
- Updated Chat.tsx to reference the new setting
- Updated GeneralSettings.tsx to include new TextComponent

### DIFF
--- a/src/settings/SettingsPage.tsx
+++ b/src/settings/SettingsPage.tsx
@@ -22,6 +22,7 @@ export interface CopilotSettings {
   googleApiKey: string;
   openRouterAiApiKey: string;
   defaultModelKey: string;
+  defaultTag: string;
   embeddingModelKey: string;
   temperature: number;
   maxTokens: number;
--- a/src/settings/SettingsPage.tsx
+++ b/src/settings/SettingsPage.tsx
@@ -22,7 +22,6 @@ export interface CopilotSettings {
   googleApiKey: string;
   openRouterAiApiKey: string;
   defaultModelKey: string;
-  defaultTag: string;
   embeddingModelKey: string;
   temperature: number;
   maxTokens: number;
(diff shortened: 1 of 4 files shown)


src/settings/SettingsPage.tsx
```diff
@@ -22,6 +22,7 @@ export interface CopilotSettings {
   googleApiKey: string;
   openRouterAiApiKey: string;
   defaultModelKey: string;
+  defaultTag: string;
```
logancyang: How about `defaultConversationTag` and put it below the `defaultSaveFolder` below

src/settings/components/GeneralSettings.tsx
```diff
@@ -79,6 +79,13 @@ const GeneralSettings: React.FC<GeneralSettingsProps> = ({
         value={settings.defaultSaveFolder}
         onChange={(value) => updateSettings({ defaultSaveFolder: value })}
       />
+      <TextComponent
+        name="Default Conversation Tag"
+        description="The default tag to be used when saving a conversation. Default is 'ai-conversations'"
```
logancyang, line 84: I don't see where you set the default `ai-conversations` though, it's in constants.ts

logancyang:
Thanks for the PR. Added some comments.

logicsec:
Updated. I think for now, 1 tag should be ok. In the future, we should take a look at ability to do tag lookups and allow for multi-select.

PR #750, opened 2024-10-26
Allow entering [[ anywhere in the prompt

zeroliu:
Allow users to enter `[[` anywhere in the prompt instead of only at the end.


logancyang:
Welcome and congrats on the first PR! 🚀🚀🚀 @zeroliu

PR #768, opened 2024-11-02
Add suggested prompts

zeroliu:
- Added new suggested prompt when chat is empty. 
- Removed the default AI message for the QA Vault mode. Replaced it with a message in the suggested prompt area and display it only when the embedding mode set to NEVER.

https://github.com/users/logancyang/projects/2?pane=issue&itemId=84610884&issue=logancyang%7Cobsidian-copilot%7C686


logancyang:
The new bubbles look awesome! 

Not sure if Obsidian plugin can use tailwindcss with the bundling, if it can we should definitely move over.

src/components/ChatComponents/SuggestedPrompts.tsx
```diff
@@ -0,0 +1,139 @@
...
+  title: string;
+  prompts: string[];
+}
+const SUGGESTED_PROMPTS: Record<string, NotePrompt> = {
+  activeNote: {
+    title: "Active Note Insights",
+    prompts: [
+      `Provide three follow-up questions worded as if I'm asking you based on {activeNote}?`,
+      `What key questions does {activeNote} answer?`,
+      `Give me a quick recap of {{activeNote}} in two sentences.`,
```
logancyang: This should be single braces instead of double?

src/components/ChatComponents/SuggestedPrompts.tsx
```diff
@@ -0,0 +1,139 @@
...
+    prompts: [
+      `9.11 and 9.8, which is bigger?`,
+      `What's the longest river in the world?`,
+      `If a lead ball and a feather are dropped simultaneously from the same height, which will reach the ground first?`,
+    ],
+  },
+  qaVault: {
+    title: "Vault Q&A",
+    prompts: [
+      `What are the main takeaways from my recent project notes?`,
```
logancyang: These 3 questions will likely fail in Vault QA because "recent project notes", "deadline", "brainstorming sessions" may not exist literally in the user's vault. They are more appropriate as Copilot Plus questions.

In general I recommend specific questions for this mode, so ideally it needs something that literally exists in the user's vault which we don't have access to.

(or do we? We could 

1. scan their vault and produce a word cloud or knowledge graph that suggests what to ask
2. scan their tags to suggest tag to ask about
3. even let llm run through some notes (*privacy hazard*, the user didn't explicitly send the notes to llm)

but probably out of the scope for this PR)

More effective questions for this mode may be
1. `What did I mention about <topic>`
2. `Summarize my notes on <subject>`
3. `What are the main takeaways from <concept>`

And add language to let the user know they must replace the target manually.

src/components/ChatComponents/ChatControls.tsx
```diff
@@ -76,13 +75,6 @@ const ChatIcons: React.FC<ChatIconsProps> = ({
         if (vault_qa_strategy === VAULT_VECTOR_STORE_STRATEGY.ON_MODE_SWITCH) {
           await onRefreshVaultContext();
         }
-        const activeNoteOnMessage: ChatMessage = {
-          sender: AI_SENDER,
-          message: `OK Feel free to ask me questions about your vault: **${app.vault.getName()}**. \n\nIf you have *NEVER* as your auto-index strategy, you must click the *Refresh Index* button below, or run Copilot command: *Index vault for QA* first before you proceed!\n\nPlease note that this is a retrieval-based QA. Specific questions are encouraged. For generic questions like 'give me a summary', 'brainstorm based on the content', Chat mode with direct \`[[note title]]\` mention is a more suitable choice.`,
```
logancyang, line 81: This message below is worth keeping above the "NEVER" warning unconditionally for Vault QA. 

"Please note that this is a retrieval-based QA. Questions should contain keywords and concepts that exist literally in your vault."

src/components/ChatComponents/SuggestedPrompts.tsx
```diff
@@ -0,0 +1,139 @@
...
+    prompts: [
+      `Provide three follow-up questions worded as if I'm asking you based on {activeNote}?`,
+      `What key questions does {activeNote} answer?`,
+      `Give me a quick recap of {{activeNote}} in two sentences.`,
+    ],
+  },
+  quoteNote: {
+    title: "Note Link Chat",
+    prompts: [
+      `Based on [[noteName]], what improvements should we focus on next?`,
```
logancyang: Should we use `<>` to indicate that the user must replace it `[[<your note>]]`?

src/components/ChatComponents/SuggestedPrompts.tsx
zeroliu: I'm struggling to come up with meaningful prompts that's more generic than this. I changed them and ask user to fill specific topic or concept with `<topic>` and `<concept>` placeholder. Similar to what you pointed out in another comment, we can train user to understand `<string>` as an indicator that this needs to be replaced. We can also swap the out when the prompt is inserted into the input and place the cursor in place. In the scope of this PR, I suggest just trying the new prompt for now. PTAL

src/components/ChatComponents/ChatControls.tsx
zeroliu: Updated
Screenshot 2024-11-02 at 11 01 01 PM
Screenshot 2024-11-02 at 11 01 28 PM

src/components/ChatComponents/SuggestedPrompts.tsx
zeroliu: updated.

I originally used this format but removed `<>` because it makes double click to select text to replace more difficult because `<>` is not selected with double clicks. However, I don't think this matters any more because select to replace is tedious to use even with double clicks. We need to follow up to make replace placeholder easier if people use this feature a lot.

logancyang (approved):
LGTM!

Issue #763, opened 2024-10-31
Model table too wide for mobile


logancyang:
The model table in Copilot setting is too wide for mobile screens. Need better responsive UI for setting.

logancyang:
Assigned to you @Emt-lin as requested!

Emt-lin:
> Will asign to you @Emt-lin as requested! Pls comment here.

ok

PR #774, opened 2024-11-04
feat: Optimize model setting style for mobile devices.

Emt-lin:
related: #763

https://github.com/user-attachments/assets/8a129a0f-01f9-4333-b63a-0b4bdf4ba117

### DIFF
--- a/src/settings/components/GeneralSettings.tsx
+++ b/src/settings/components/GeneralSettings.tsx
@@ -74,8 +74,8 @@ const GeneralSettings: React.FC<GeneralSettingsProps> = ({
         isEmbeddingModel={false}
       />
       <div className="chat-icon-selection-tooltip">
+        <h2>Default Mode</h2>
         <div className="select-wrapper">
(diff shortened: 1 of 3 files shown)


logancyang:
Nice! One small problem I see is that the radios and toggles of each row have misalignment in the original wide UI:
SCR-20241104-msjy

src/settings/components/GeneralSettings.tsx
```diff
@@ -85,7 +85,7 @@ const GeneralSettings: React.FC<GeneralSettingsProps> = ({
             <option value={ChainType.LLM_CHAIN}>Chat</option>
             <option value={ChainType.VAULT_QA_CHAIN}>Vault QA (Basic)</option>
           </select>
-          <span className="tooltip-text">Default Mode Selection</span>
+          {/*<span className="tooltip-text">Default Mode Selection</span>*/}
```
logancyang: This line can be removed/

PR #779, opened 2024-11-05
Fix ollama embedding context length with truncation

logancyang:
Temporary fix for ollama embedding using truncation.


chenyuz3:
hi @logancyang , after upgrading to 2.7.1 my local ollama embedding model starts to show the following errors, I thought it might be helpful to report.

image
image

PR #797, opened 2024-11-09
FE: Refactor chat control tooltips

zeroliu:
- Use radix tooltip to implement tooltips in the chat control. It's easier to maintain and handles more edge cases.
- Made chat control icon bigger.
- Made "unsaved history will be lost" message display conditionally based on auto save settings

### DIFF
--- a/src/components/ChatComponents/ChatControls.tsx
+++ b/src/components/ChatComponents/ChatControls.tsx
@@ -3,42 +3,33 @@ import { CopilotPlusModal } from "@/components/CopilotPlusModal";
 import { VAULT_VECTOR_STORE_STRATEGY } from "@/constants";
 import { CustomError } from "@/error";
 import { CopilotSettings } from "@/settings/SettingsPage";
-import { ChatMessage } from "@/sharedState";
-import { Notice, Vault } from "obsidian";
+import { Notice } from "obsidian";
 import React, { useEffect, useState } from "react";
 
 import { ChainType } from "@/chainFactory";
-import { RefreshIcon, SaveAsNoteIcon, UseActiveNoteAsContextIcon } from "@/components/Icons";
 import { stringToChainType } from "@/utils";
 import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
-import { ChevronDown } from "lucide-react";
+import { ChevronDown, Download, Puzzle, RefreshCw } from "lucide-react";
+import { TooltipActionButton } from "@/components/ChatComponents/TooltipActionButton";
 
 interface ChatControlsProps {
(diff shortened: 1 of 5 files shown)


src/components/ChatComponents/ChatInput.tsx
```diff
@@ -247,40 +242,24 @@ const ChatInput: React.FC<ChatInputProps> = ({
 
         <div className="chat-input-buttons">
           {isGenerating && (
-            <Tooltip.Provider delayDuration={0}>
```
zeroliu, line 250: removed because there is no tooltip for these buttons

src/components/ChatComponents/ChatControls.tsx
```diff
@@ -106,36 +98,25 @@ const ChatControls: React.FC<ChatControlsProps> = ({
...
-              <span>Refresh Index for Vault</span>
-            </div>
-          </button>
-          {/* <button className="chat-icon-button clickable-icon" onClick={handleFindSimilarNotes}>
-            <ConnectionIcon className="icon-scaler" />
-            <span className="tooltip-text">Find Similar Notes for Active Note</span>
-          </button> */}
-        </>
+        <TooltipActionButton
+          onClick={onSaveAsNote}
```
logancyang: Shouldn't this be `onRefreshVaultContext`?

src/components/ChatComponents/ChatControls.tsx
```diff
@@ -106,36 +98,25 @@ const ChatControls: React.FC<ChatControlsProps> = ({
...
+          Icon={<UseActiveNoteAsContextIcon className="icon-scaler" />}
```
logancyang: Since we are updating this, `UseActiveNoteAsContextIcon` is a legacy name, we can rename it to "RefreshIndexIcon" or something.

src/components/ChatComponents/ChatControls.tsx
zeroliu: good catch

src/components/ChatComponents/ChatControls.tsx
zeroliu: I saw we are using lucide icons. I updated them to use the library instead of custom Icons. WDYT?

zeroliu: we can follow up to replace the rest of icons in the app.

logancyang: @zeroliu That would be great! These were added by a total frontend noob - me. Glad this will be cleaned up now!

logancyang (approved):
LGTM!

Issue #785, opened 2024-11-06
Allow user to delete the first two options 'gpt-4o' and 'gpt-4o-mini'


freiz:
First thanks for the great plugin I used it heavily everyday!

**Is your feature request related to a problem? Please describe.**
I have multiple models configured, but they are not from openai endpoint directly while I always can see these two options... so it is a little hard for me to pick the correct one other than default.
<img width="759" alt="image" src="https://github.com/user-attachments/assets/b7e22ae3-b306-409b-873f-587597750ea6">

**Describe the solution you'd like**
Allow me to delete the first two. (adding the delete button like other ones)

**Describe alternatives you've considered**

**Additional context**

Issue #756, opened 2024-10-28
Also exclude files that are excluded by Obsidian's settings


TroySigX:
**Is your feature request related to a problem? Please describe.**
When I want to exclude a file from being indexed, I have to exclude it twice in Obsidian setting tab (Files and links tab) and in Copilot.

**Describe the solution you'd like**
I'd like Copilot to exclude all the files that are already excluded by Obsidian settings.

**Describe alternatives you've considered**
Make a button to let users decide whether to also exclude the files excluded by Obsidian settings

**Additional context**
![image](https://github.com/user-attachments/assets/6eb8c85c-32e2-4ec6-b50b-1d9f31866b33)
![image](https://github.com/user-attachments/assets/063ff5a7-6b81-4c41-b8eb-9b868db3465a)

logancyang:
Ok we can get the user setting with `app.vault.getConfig('userIgnoreFilters')` (not documented in Obsidian API but I got this answer from Obsidian's discord)

Emt-lin:
@TroySigX If you want to prevent files to be indexed, you only set exclude files option in Copilot setting. Why you need to set it in the Obsidian setting.

TroySigX:
Because I also don't want those files to appear in Obsidian search as well. Those notes are mainly drafts.

logancyang:
@Emt-lin It may be better UX to inherit from the Obsidian "master exclusion" setting, since the user explicitly excluded these files from Obsidian overall, all the plugins should operate within the "master setting". If we inherit it, the user doesn't need to repeat themselves in Copilot settings.

Emt-lin:
@logancyang pls assign me

PR #817, opened 2024-11-15
Support delete core models, inherit excluding from Obsidian

Emt-lin:
related：#785、#756

### DIFF
--- a/src/VectorStoreManager.ts
+++ b/src/VectorStoreManager.ts
@@ -359,8 +359,21 @@ class VectorStoreManager {
   private async getFilePathsForQA(filterType: "exclusions" | "inclusions"): Promise<Set<string>> {
     const targetFiles = new Set<string>();
 
-    if (filterType === "exclusions" && this.settings.qaExclusions) {
-      const exclusions = this.settings.qaExclusions.split(",").map((item) => item.trim());
+    // not documented in Obsidian API, but I got this answer from Obsidian's discord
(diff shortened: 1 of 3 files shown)


src/VectorStoreManager.ts
```diff
@@ -286,8 +286,21 @@ class VectorStoreManager {
   private async getFilePathsForQA(filterType: "exclusions" | "inclusions"): Promise<Set<string>> {
     const targetFiles = new Set<string>();
 
-    if (filterType === "exclusions" && this.settings.qaExclusions) {
-      const exclusions = this.settings.qaExclusions.split(",").map((item) => item.trim());
+    // not documented in Obsidian API, but I got this answer from Obsidian's discord
+    const userIgnoreFilters: string[] = (app.vault as any).getConfig("userIgnoreFilters");
```
logancyang: It could be better to be a bit more defensive in case `getConfig` fails.

logancyang:
SCR-20241126-pykx

SCR-20241126-pyhn


The notice doesn't show the combined exclusions correctly.

Emt-lin:
@logancyang Can we need to add some default exclusions.(eg: copilot-conversations、copilot-custom-prompts folders)

logancyang:
> @logancyang Can we need to add some default exclusions.(eg: copilot-conversations、copilot-custom-prompts folders)

Good point. Custom prompts definitely needs to be excluded. Conversations may be helpful in some cases.

PR #888, opened 2024-12-02
feat: add tailwind + shadcn/ui.

Emt-lin:
Support out-of-the-box the Tailwind + shadcn/ui functionality


styles.css
```diff
@@ -1,41 +1,416 @@
...
+  .container {
+    max-width: 1280px;
+  }
+}
+@media (min-width: 1536px) {
+  .container {
+    max-width: 1536px;
+  }
+}
+.visible {
```
zeroliu: should this file be checked in? It looks like the generated tailwind CSS file

src/styles/tailwind.css
```diff
@@ -0,0 +1,1101 @@
...
+/*
+
+This CSS file will be included with your plugin, and
+available in the app when your plugin is enabled.
+
+If your plugin does not need CSS, delete this file.
+
+*/
+
+.button-container {
```
zeroliu: what are these styles?

src/styles/tailwind.css
```diff
@@ -0,0 +1,1101 @@
+@import "tailwindcss/base";
+@import "tailwindcss/components";
+@import "tailwindcss/utilities";
+
+@layer base {
+  :root {
+    --background: 0 0% 100%;
```
zeroliu: to make UI library compatible with obsidian plugin, we should use the CSS variables defined in obsidian (https://docs.obsidian.md/Reference/CSS+variables/CSS+variables). Many obsidian users use themes, which are made by altering those CSS variables (https://docs.obsidian.md/Themes/App+themes/Build+a+theme). Users would expect a well-designed plugin to change the appearance accordingly.

This is the main reason why I believe shadcn is the right choice for obsidian plugin UI library. It provides developer great flexibility to change how components are styled. However, we will lose the value if we use the shadcn default CSS variables.

I suggest:
1. Do not introduce new CSS variables
2. update tailwind.config.js to create tokens that map to obsidian CSS variables, including color, spacing, sizing, etc
3. whenever we introduce a shadcn component, restyle it with the new token before we integrate it into the plugin.

logancyang:
I checked out this branch and ran locally, it seems to override some obsidian native CSS vars
SCR-20241203-seis

Emt-lin:
@logancyang @zeroliu 

now, `style.css` is already generated by tailwind cli. The original traditional css style are in`tailwind.css` file. Perhaps We should make these traditional css style into tailwind style.

logancyang:
> @logancyang @zeroliu
> 
> now, `style.css` is already generated by tailwind cli. The original traditional css style are in`tailwind.css` file. Perhaps We should make these traditional css style into tailwind style.

Is this going to be quite a big effort to make everything look the same/similar? I'm not familiar with how it can be converted though. @zeroliu is the expert here I'll wait for his recommendation.

src/styles/tailwind.css
Emt-lin: @zeroliu sometimes, I don't want to use some of the built-in theme styles.`(eg: radius)`
so, I think we should select some css styles to map.

Emt-lin:
> > @logancyang @zeroliu
> > now, `style.css` is already generated by tailwind cli. The original traditional css style are in`tailwind.css` file. Perhaps We should make these traditional css style into tailwind style.
> 
> Is this going to be quite a big effort to make everything look the same/similar? I'm not familiar with how it can be converted though. @zeroliu is the expert here I'll wait for his recommendation.

I already delete `styles.css` file in latest commit, it will be created by `tailiwind cli` When you are running `npm run dev or build`

src/styles/tailwind.css
```diff
@@ -159,14 +163,6 @@ If your plugin does not need CSS, delete this file.
   opacity: 1;
 }
 
-.context-note .note-name {
```
logancyang: These 3 red blocks were added recently, why are they removed here?

package.json
```diff
@@ -4,8 +4,11 @@
   "description": "ChatGPT integration for Obsidian",
   "main": "main.js",
   "scripts": {
-    "dev": "node esbuild.config.mjs",
-    "build": "tsc -noEmit -skipLibCheck && node esbuild.config.mjs production",
+    "dev": "npm-run-all --parallel dev:*",
+    "dev:tailwind": "npx tailwindcss -i src/styles/tailwind.css -o ./styles.css --watch",
+    "dev:esbuild": "node esbuild.config.mjs",
+    "build:tailiwind": "npx tailwindcss -i src/styles/tailwind.css -o ./styles.css --minify",
```
logancyang: Typo: `build:tailiwind`. I recommend some form of AI assistance to spot these things in the future. It's hard for reviewers to spot.

package.json
```diff
@@ -4,8 +4,11 @@
...
+    "build": "npm run build:tailiwind && tsc -noEmit -skipLibCheck && node esbuild.config.mjs production",
```
logancyang: How about something like
```
{
  "scripts": {
    "build:tailwind": "npx tailwindcss -i src/styles/tailwind.css -o ./styles.css --minify",
    "build:esbuild": "tsc -noEmit -skipLibCheck && node esbuild.config.mjs production",
    "build": "npm run build:tailwind && npm run build:esbuild || exit 1"
  }
}
```

logancyang:
Another concern is that the bundle size is up significantly from 5MB+ to 21MB+. Is there any way we can make it smaller?

src/styles/tailwind.css
Emt-lin: @logancyang Maybe I don't use the latest `styles.css` in master branch. I will replace it by  the lasted `styles.css`

src/styles/app.css
zeroliu, line 1: what is this file?

tailwind.config.js
```diff
@@ -0,0 +1,70 @@
...
+/** @type {import("tailwindcss").Config} */
+module.exports = {
+  content: ["./src/**/*.{js,ts,jsx,tsx}", "./src/styles/tailwind.css"],
+  darkMode: ["class", '[data-theme="dark"]'],
+  plugins: [tailwindcssAnimate],
+  corePlugins: {
+    preflight: false,
+  },
+  theme: {
+    extend: {
```
zeroliu: can we not extend but redefine all the tokens? Otherwise, people can still use predefined tailwind tokens without realizing that it's not connected to obsidian CSS variables. We can also turn on this eslint rule to enforce only supported classnames are used in the code base: https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/no-custom-classname.md

src/components/ui/input.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { cn } from "@/lib/utils";
+
+const Input = React.forwardRef<HTMLInputElement, React.ComponentProps<"input">>(
+  ({ className, type, ...props }, ref) => {
+    return (
+      <input
+        type={type}
+        className={cn(
+          "flex h-9 w-full rounded-md border border-input bg-transparent px-3 py-1 text-base shadow-sm transition-colors file:border-0 file:bg-transparent file:text-sm file:font-medium file:text-foreground placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring disabled:cursor-not-allowed disabled:opacity-50 md:text-sm",
```
zeroliu: do we support file type input? If not, I suggest not adding the related style until it's needed

src/components/ui/input.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react";
+
+import { cn } from "@/lib/utils";
+
+const Input = React.forwardRef<HTMLInputElement, React.ComponentProps<"input">>(
```
zeroliu: not blocking this PR but maybe we can look into React 19 and use the `ref` props now that it's officially GA. Maintaining forwardRef for RadixUI can be pretty tedious.

zeroliu (commented):
Left some more comments. Looking good overall! Looking forward to using tailwindCSS in the near future

logancyang:
@Emt-lin Thanks for the sync today! When you get a chance, could you address zero's comments? And I think this should be ready to merge after!

src/styles/app.css
Emt-lin: nothing, I will remove it.

src/components/ui/input.tsx
Emt-lin: yeah, I agree with your opinion. However, I don't want to increase the complexity of maintenance If I remove the styles related to file type input.

src/constants.ts
```diff
@@ -121,7 +121,8 @@ export enum EmbeddingModelProviders {
   OPENAI = "openai",
   COHEREAI = "cohereai",
   GOOGLE = "google",
-  AZURE_OPENAI = "azure_openai",
+  // AZURE_OPENAI = "azure_openai",
+  AZURE_OPENAI = "azure openai",
```
zeroliu, line 127: does it belong to this PR?

src/styles/tailwind.css
```diff
@@ -1,3 +1,110 @@
+@import "tailwindcss/base";
+@import "tailwindcss/components";
+@import "tailwindcss/utilities";
+
+@layer base {
+  body {
+    /* preDefine CSS variables in Obsidian.(https://docs.obsidian.md/Reference/CSS+variables/Foundations/Colors) */
+    /*Primary background*/
+    --ob-bg-primary: var(--background-primary);
```
zeroliu: why do we need to redefine obsidian CSS with `--ob-bg` prefix?

src/styles/tailwind.css
```diff
@@ -1,3 +1,110 @@
...
+    --ob-text-accent: var(--text-accent); /*Accent text*/
+    --ob-text-accent-hover: var(--text-accent-hover); /*Accent text (hover)*/
+
+    /*Text background colors*/
+    --ob-text-selection: var(--text-selection); /*Selected text background*/
+    --ob-text-highlight-bg: var(--text-highlight-bg); /*Highlighted text background*/
+  }
+}
+
+/* this layer for overwrite Obsidian default css style(file: app.css) and theme css style */
```
zeroliu: I don't think we should be touching the default component style as it will affect UI outside of obsidian-copilot

tailwind.config.js
```diff
@@ -0,0 +1,105 @@
+// tailwind.config.js
+
+import tailwindcssAnimate from "tailwindcss-animate";
+import { colorOpacityPlugin } from "./src/lib/plugins/colorOpacityPlugin";
+import colors from "tailwindcss/colors";
+
+/** @type {import("tailwindcss").Config} */
```
zeroliu, line 7: we can use tailwind.config.ts to get full type support instead of this type comment

tailwind.config.js
```diff
@@ -0,0 +1,105 @@
...
+  corePlugins: {
+    preflight: false,
+  },
+  theme: {
+    colors: {
+      inherit: colors.inherit,
+      current: colors.current,
+      transparent: colors.transparent,
+      black: colors.black,
+      white: colors.white,
```
zeroliu: what's the use case of black, white, red, yellow? Are there existing semantic tokens that can replace them?

tailwind.config.js
```diff
@@ -0,0 +1,105 @@
...
+        md: "var(--radius-m)", // 8px
+        sm: "var(--radius-s)", // 4px
+      },
+      spacing: {
+        1: "var(--size-4-1)", // 4px
+        2: "var(--size-4-2)",
+        3: "var(--size-4-3)",
+        4: "var(--size-4-4)",
+        5: "var(--size-4-5)",
+        5.5: "calc(var(--size-4-5) + 2px)",
```
zeroliu, line 152: When do we need this token? It doesn't feel necessary to introduce a new token that needs to calculate a new value outside of the obsidian design system?

src/settings/model.ts
zeroliu, line 1: does this change belong to this PR?

src/constants.ts
Emt-lin: yeah,  it's a separate commit and @logancyang also learn about this.

src/styles/tailwind.css
Emt-lin: I think that  redefine obsidian CSS will more clarify. I want to provide out-of-the-box functionality, when we add some shadcn/ui component.

src/styles/tailwind.css
Emt-lin: @zeroliu It doesn't affect UI outside of obsidian-copilot, It only work for copilot ui.

tailwind.config.js
Emt-lin: The JS type documentation also provides full type support. Therefore, both TS and JS can work.

tailwind.config.js
Emt-lin: I think we don't need full semantic tokens support. It only base color. We don't need to fully trap ourselves in the concept of Obsidian. It's also tailwind project.

tailwind.config.js
Emt-lin: In the button component, I need the define of `5.5`, others else can be removed.

src/styles/tailwind.css
zeroliu: Do you have any specific features in mind that would require us to break apart from the default css variables? This is an additional layer that we have to maintain. If we want to apply a new obsidian css variable that is not included in this PR, the process of adding it would require us to:
1. Make a copy of the css variable 
2. Create a new tailwind token with a different semantic value 

The more undocumented overhead we add, the more difficult it is for new contributors to make changes. 

If you feel strongly that creating our own semantic token is necessary, and it can simplify the component library adoption, I'm supportive of that. However, we should at least consider not adding another layer of complexity and use the obsidian css variable directly.

Also worth noting that any external CSS variable that we introduce will not be compatible with the obsidian theme, which can be a dealbreaker for some users.

tailwind.config.js
zeroliu: Using colors like white and black directly will likely to cause dark theme to break, as well as causing unexpected edge cases for people using non default themes. Do you have a good way to avoid these potential breakage?

src/styles/tailwind.css
zeroliu: This will affect all matching components. I created a sample plugin and added a button with "select-trigger-reset" class. By loading this version of copilot, my settings button now looks different because of this style set in copilot plugin:

Screenshot 2024-12-21 at 9 04 53 PM

zeroliu: Just checked again, this is not a new problem introduced by this PR. The existing style can also affect other plugins. If I add `warning-message` to my plugin element and load the master version copilot, my element will also be affected. I sent a question to the obsidian discord to understand how style isolation works in obsidian. It seems very error-prone.

tailwind.config.js
Emt-lin: I test some top theme. It's work well in light/dark mode. Dialog overlay will use the black color. I think you can pull up pr #955, you will know about something.

src/styles/tailwind.css
Emt-lin: `select-trigger-reset` is my custom css classname. Others have a very low probability of using it. 

> This will affect all matching components. 

I was aware of css conflicts when developing the UI earlier.

The reason：
CleanShot 2024-12-23 at 10 42 55

src/styles/tailwind.css
Emt-lin: I am merely listing all the obsidian css variables rather than defining a new set of variables.

> If we want to apply a new obsidian css variable

I have listed most of the color css variables.

> The more undocumented overhead we add, the more difficult it is for new contributors to make changes.

When a new contributor joins, he/her doesn't need to do anything extra work, and he/her doesn't need to care about these. 

He can use existing components or download new ones from the shadcn/ui, and then he doesn't need to do anything, the component will automatically change with the theme.

Emt-lin:
@logancyang @zeroliu please review this pr.

tailwind.config.js
```diff
@@ -0,0 +1,221 @@
...
+  // https://github.com/tailwindlabs/tailwindcss/blob/main/stubs/config.full.js
+  theme: {
+    colors: {
+      inherit: colors.inherit,
+      current: colors.current,
+      transparent: colors.transparent,
+      black: colors.black,
+      red: colors.red,
+      yellow: colors.yellow,
+      // slate: colors.slate,
```
zeroliu: please clean up the comments

zeroliu (approved):

tailwind.config.js
```diff
@@ -0,0 +1,221 @@
...
+        25: "var(--color-base-25)",
+        30: "var(--color-base-30)",
+        35: "var(--color-base-35)",
+        40: "var(--color-base-40)",
+        50: "var(--color-base-50)",
+        60: "var(--color-base-60)",
+        70: "var(--color-base-70)",
+        100: "var(--color-base-100)",
+      },
+      // red: "var(--color-red)",
```
zeroliu: can we not use the obsidian defined red and yellow color?

Emt-lin: The defalut color offers more choices. Why we need the red color?

CleanShot 2024-12-26 at 14 21 29

zeroliu: This is the principle of limiting the choice for developers and consistency. 

Regarding limiting the choice, TailwindCSS is a generic CSS library; its goal is to allow people to build any applications. TW has to provide not only a complete set of primitive tokens and semantic tokens and let people decide what they want. Our goal is not to replicate TailwindCSS tokens but to tailor them down to what the copilot plugin may need. We don't want developers to add tokens like `bg-color-red-50` without a specific reason in the plugin because #fef2f2 is not a color in either the obsidian design system or copilot design.

For consistency, this PR defines red, yellow, and black differently from the rest of the colors, making it harder to predict when applying color tokens. We would either go with Tailwind color tokens entirely or choose only obsidian colors. Staying in the middle will make decision-making harder in the future.

Unless we have a concrete use case that the obsidian color tokens cannot fit and it adds value to the copilot design, we should refrain from introducing new tokens.

Emt-lin: > Staying in the middle will make decision-making harder in the future.

agree.

But Why we need red color? Isn't it a primitive color? It's not a semantic token.

Emt-lin: below pic -> text: yellow-700, bg: yellow-50, What should I do? eg: text-yellow-rgb/10, bg-yellow-rgb/50?

Google Chrome 2024-12-26 16 09 28

zeroliu: This is the callout component used in the obsidian app and site. 
image


It uses --callout-warning which is --color-orange-rgb. The background is rgba(var(--callout-warning), 0.1)). There is no specific token for it. 

My obsidian tailwind project didn't include all the component tokens from obsidian yet. You can add this token to your PR. Then you can build it as `bg-callout-warning/10` and `text-callout-warning`. Here's the reference https://docs.obsidian.md/Reference/CSS+variables/Editor/Callout

logancyang:
Just to confirm, this new setup compiles tailwind to `styles.css`, yes? So the release process is still to run `npm run build` and release `main.js`, `styles.css` and `manifest.json` without any additional steps, correct? @Emt-lin @zeroliu

Emt-lin:
> Just to confirm, this new setup compiles tailwind to `styles.css`, yes? So the release process is still to run `npm run build` and release `main.js`, `styles.css` and `manifest.json` without any additional steps, correct? @Emt-lin @zeroliu

yeah. cc @zeroliu

PR #955, opened 2024-12-20
feat: New Setting UI.

Emt-lin:
**New Setting UI.**

`I tested some popular themes. It's work well in light/dark mode. `

CleanShot 2025-01-10 at 18 51 20


logancyang:
Tailwind PR is merged! This one can rebase master now.

zeroliu:
Thanks for integrating with the latest tokens! Before I dive into the implementation, I have a few pieces of feedback on the UI

1. Can we add a border to the slider control? It's hard to see in the following example

Screenshot 2025-01-05 at 1 14 59 PM


2. Can we use a neutral color for a switch when the value is false? The opaque version makes it look like it's disabled.

Screenshot 2025-01-05 at 1 17 08 PM


3. Can we make the tabs show text by default? The animation is nice, but it's hard to tell what each tab is about from the tab icon. There are many horizontal rooms for all the tabs to show text by default. The larger tab is also easier to use on mobile.

Screenshot 2025-01-05 at 1 18 05 PM

4. A list view may be more suitable for API key management. The current UI feels like people can set one API key for one provider at a time. It's also challenging to know what API keys I have set and have to pick providers one at a time to reveal the API keys. Ideally, all the providers with API keys set should show up by default.

Screenshot 2025-01-05 at 1 21 38 PM

I built an alternative version with V0. Please feel free to use it as a reference. https://v0.dev/chat/F2dDiCZXeX2?b=b_1x2gOAr4gtO

image



5. The border of this table can be removed or should use the highlight color. The gap looks off.

Screenshot 2025-01-05 at 1 32 59 PM

6. The grouping may benefit from some adjustments
- The API key management and chat model config can live in the same tab
- Configs such as "default mode", "default conversation folder name" and configs that are not AI model related can live in the "basic" tab.
- Everything Embedding-related can have its tab

7. Can we configure temperature and context length in each AI model? How is it related to the global temperate and context length settings? This doesn't exist in the existing settings.

Screenshot 2025-01-05 at 1 36 59 PM

logancyang:
@Emt-lin This is a major milestone, very nice and clean! Congrats! Once this gets released I will bump the version to v2.8 🚀 !

+1 for everything @zeroliu said. Here are some of my own initial observations:

1. Can we move this to the top as Plus will be our first-class citizen and we should highlight it wherever possible. Also small nit: this question mark doesn't show the tooltip on hover but only with a click.
SCR-20250105-pmis

2. Is "User System Prompt" missing? I don't see it on any tab.
SCR-20250105-pmbm

3. This one should be "Chat Model[s]", and echoing @zeroliu 's comment, we don't have per-model param at the moment, so this edit button and modal shouldn't be implemented now
SCR-20250105-pkmd

4. The provider name can get cutoff
SCR-20250105-pjlo

logancyang:
Very close to the finishing line! This is the biggest change we have before the official launch. It would be awesome if we can ship this in less than a week. Thanks for your hard work on this!! @Emt-lin 

Some more comments:
1. Should API be its own section and not in the Chat section? Since both Chat and Embedding models require API keys. We can also add a "?" or something obvious to say "API key required for chat and QA features".
SCR-20250107-pxhl

2. Add "default" to "chat model" and "embedding model"
3. Auto index strategy should be in the QA tab (I think we should have QA as a separate tab)
4. Again, we don't need these edit buttons at the moment 
SCR-20250107-pyax

5. User System Prompt should be in Advanced settings as before
SCR-20250107-pyfa

6. Everything below this line should be in the QA tab
SCR-20250107-pylp

Just some reorganization, overall looks great and 100x better than what I have now! 👍

src/LLMProviders/chatModelManager.ts
```diff
@@ -337,8 +339,13 @@ export default class ChatModelManager {
...
         return true;
       } catch (error) {
-        console.error("Chat model ping failed:", error);
-        throw error;
+        const msg =
+          "\nwithout CORS Error: " +
+          err2String(firstError) +
+          "\nwith CORS Error: " +
+          err2String(error);
+        // console.error("Chat model ping failed:", error);
```
logancyang: Nit: keep or remove?

src/aiParams.ts
```diff
@@ -73,6 +73,17 @@ export interface CustomModel {
   isBuiltIn?: boolean;
   enableCors?: boolean;
   core?: boolean;
+  stream?: boolean;
+  temperature?: number;
+  context?: number;
```
logancyang, line 78: What is `context`? I assume you mean `maxToken`?

Emt-lin: > What is `context`? I assume you mean `maxToken`?

yeah, I will remove it in this PR. The next PR will implement the single model edit feature.

logancyang: Great, single model parameters is a first step to the "project" feature next. Each project has its own:
- Model key
- Model params
- User system prompt
- Subset of the index with inclusions filter

logancyang:
@Emt-lin The reorganization looks good! How does mobile settings work with this PR?

I'd like to merge this one asap since we are close to the official launch and we need some more user testing before it. @zeroliu let me know if you have any concerns.

src/settings/SettingsPage.tsx
```diff
@@ -44,6 +45,10 @@ export class CopilotSettingTab extends PluginSettingTab {
     const div = containerEl.createDiv("div");
     const sections = createRoot(div);
 
+    if (!Platform.isMobile) {
+      sections.render(<SettingsMainV2 plugin={this.plugin} />);
+      return;
+    }
```
logancyang: Does this mean mobile is still using the old settings? Do we plan to update mobile too in the future?

Emt-lin: > Does this mean mobile is still using the old settings? 

yes, maybe some componments will adopt new implementations in mobile devices.

src/contexts/TabContext.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const [selectedTab, setSelectedTab] = useState("basic");
+  const [modalContainer, setModalContainer] = useState<HTMLElement | null>(null);
+
+  useEffect(() => {
+    // fix
+    if (!modalContainer) {
+      const modal = document.querySelector(".modal-container") as HTMLElement;
+      setModalContainer(modal);
+    }
+  }, []);
```
zeroliu: there is an eslint error here. Also what's the reason for this effect?

src/settings/SettingsPage.tsx
zeroliu: What's the reason why mobile can't work with the new settings? It will add overhead for us to maintain two versions of settings. I would highly recommend us to keep one version if possible

src/settings/v2/components/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  // 在 dialog 关闭时清空未验证的 keys
+  useEffect(() => {
+    if (!open) {
+      unverifiedKeys.forEach((provider) => {
+        const settingKey = ProviderSettingsKeyMap[provider];
+        updateSetting(settingKey, "");
+      });
+    }
+  }, [open]);
```
zeroliu: please fix the dependency

src/settings/v2/components/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,211 @@
...
+}) => {
+  const [verifyingProviders, setVerifyingProviders] = useState<Set<DisplayKeyProviders>>(new Set());
+  const [unverifiedKeys, setUnverifiedKeys] = useState<Set<DisplayKeyProviders>>(new Set());
+
+  // 在 dialog 打开时重置未验证的 keys 集合
+  useEffect(() => {
+    if (open) {
+      setUnverifiedKeys(new Set());
+    }
+  }, [open]);
```
zeroliu: why can't this happen when the dialog closes too?

src/settings/v2/components/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,211 @@
...
+                          <>
+                            <Loader2 className="mr-2 h-4 w-4 animate-spin" />
+                            Verify
+                          </>
+                        ) : (
+                          "Verify"
+                        )}
+                      </Button>
+                    ) : (
+                      <span className="text-[#4CAF50] text-sm flex items-center justify-center h-9">
```
zeroliu: why do we need an arbitrary text color here?

src/settings/v2/components/BasicSettings.tsx
```diff
@@ -0,0 +1,372 @@
...
+    setOpenPopoverIds((prev) => {
+      const newSet = new Set(prev);
+      newSet.delete(id);
+      return newSet;
+    });
+  };
+
+  return (
+    <div className="space-y-4">
+      {/* Copilot Plus */}
```
zeroliu:
```suggestion
```
remove?

src/styles/tailwind.css
```diff
@@ -2,6 +2,27 @@
...
+  input[type="text"].input-reset,
+  input[type="password"].input-reset,
+  input[type="number"].input-reset {
+    @apply h-9 !min-w-[50px] border border-solid rounded-md bg-transparent px-3 py-1 md:text-base text-sm shadow-sm transition-colors;
+    @apply focus-visible:shadow-sm focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring;
+  }
+
+  .modal-container button.select-trigger-reset {
+    @apply h-9 border border-solid rounded-md bg-dropdown px-3 py-1 text-base md:text-sm shadow-sm transition-colors;
+    @apply focus-visible:shadow-sm focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring;
```
zeroliu: is there any reasons why we don't add these classes directly to the component? (e.g. make them part of SelectPrimitive.Trigger). Same for the others

src/contexts/TabContext.tsx
Emt-lin: Portal component need container arg. eg: select、dialog、popover

src/settings/v2/components/ApiKeyDialog.tsx
Emt-lin: ......,AI generated.

Emt-lin: I will change to `text-success` color.

src/settings/SettingsPage.tsx
Emt-lin: @zeroliu Model list will continue to use old layout in mobile device.

Emt-lin: I will remove the old settings.

src/contexts/TabContext.tsx
zeroliu: Is there any reason why we need to specify the modal container as the portal container? I suggest using body directly like the current tooltip if possible. Radix uses body by default and because of obsidian special requirement, we need to attach the activeDocument body to it as containers. It will simplify the implementation a lot.

Emt-lin:
@zeroliu @logancyang Embedding Model title has already modified, but I don't want to modify the implementation of the dialog. If there is a lot of feedback after launching the `New setting UI`, I will replace the shadcn/ui dialog with native `modal` .

logancyang:
I think this is lgtm generally. 

One potential caveat: we actually do not allow deleting core models at the moment (the code was lost from before), but that's actually a good thing because we allow deleting the "default chat model" with this new UI. Since gpt 4o is the top chat model in our builtin models, deleting a default chat model automatically sets the default to gpt 4o which is not deletable. This is an acceptable behavior now. Just wanted to point it out.

logancyang:
@zeroliu let me know when you think this is ready to merge and I'll merge it.

logancyang:
Merging now. Will follow up with minor updates in the future according to user feedback. Huge congrats @Emt-lin !!! 🚀🚀🚀

PR #967, opened 2024-12-25
Fix lost embeddings

logancyang:
Remove unnecessary `documents`, use `RawData` directly.


logancyang:
@zeroliu Sorry about this! This should fix it, let me know if there's any other issue!

PR #979, opened 2024-12-28
Clean up function args

zeroliu:
Clean up constructor and function arguments when singleton can be used. See inline comments for more details

Tests
- QA vault works
- `@vault`, `@web`, `@youtube`, `@pomodoro` works
- removing files index works
- custom db search command works
- updating exclusion path works

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,6 +1,5 @@
 import { BrevilabsClient } from "@/LLMProviders/brevilabsClient";
 import ChainManager from "@/LLMProviders/chainManager";
-import EmbeddingsManager from "@/LLMProviders/embeddingManager";
 import { CustomModel } from "@/aiParams";
 import { parseChatContent, updateChatMemory } from "@/chatUtils";
 import { registerBuiltInCommands } from "@/commands";
@@ -29,7 +28,6 @@ import {
 } from "@/settings/model";
 import SharedState from "@/sharedState";
 import { FileParserManager } from "@/tools/FileParserManager";
-import { Embeddings } from "@langchain/core/embeddings";
 import {
   Editor,
   MarkdownView,
@@ -68,13 +66,12 @@ export default class CopilotPlugin extends Plugin {
     // Always have one instance of sharedState and chainManager in the plugin
     this.sharedState = new SharedState();
 
-    this.vectorStoreManager = new VectorStoreManager(this.app);
-    await this.vectorStoreManager.waitForInitialization();
+    this.vectorStoreManager = VectorStoreManager.getInstance();
 
     // Initialize BrevilabsClient
     this.brevilabsClient = BrevilabsClient.getInstance();
 
-    this.chainManager = new ChainManager(this.app, this.vectorStoreManager, this.brevilabsClient);
+    this.chainManager = new ChainManager(this.app, this.vectorStoreManager);
 
     // Initialize FileParserManager early with other core services
     this.fileParserManager = new FileParserManager(this.brevilabsClient);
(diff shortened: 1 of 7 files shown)


src/main.ts
```diff
@@ -68,13 +66,12 @@ export default class CopilotPlugin extends Plugin {
     // Always have one instance of sharedState and chainManager in the plugin
     this.sharedState = new SharedState();
 
-    this.vectorStoreManager = new VectorStoreManager(this.app);
-    await this.vectorStoreManager.waitForInitialization();
```
zeroliu, line 72: no need to explicitly block onload. The initialization will happen in the background and all public methods will wait for initialization.

src/search/hybridRetriever.ts
```diff
@@ -122,11 +108,12 @@ export class HybridRetriever extends BaseRetriever {
...
     try {
       const promptResult = await this.queryRewritePrompt.format({ question: query });
-      const rewrittenQueryObject = await this.llm.invoke(promptResult);
+      const chatModel = ChatModelManager.getInstance().getChatModel();
+      const rewrittenQueryObject = await chatModel.invoke(promptResult);
 
       // Directly return the content assuming it's structured as expected
       if (rewrittenQueryObject && "content" in rewrittenQueryObject) {
-        return rewrittenQueryObject.content;
+        return rewrittenQueryObject.content as string;
```
zeroliu, line 116: there is a type issue here but this refactor doesn't have any logic change. It should be safe to cast to string and keep the same type definition as before.

src/search/vectorStoreManager.ts
```diff
@@ -1,22 +1,22 @@
...
 import { IndexEventHandler } from "./indexEventHandler";
 import { IndexOperations } from "./indexOperations";
 
 export default class VectorStoreManager {
+  private static instance: VectorStoreManager;
   private indexOps: IndexOperations;
   private eventHandler: IndexEventHandler;
   private initializationPromise: Promise<void>;
   private lastKnownSettings: CopilotSettings | undefined;
-  public dbOps: DBOperations;
```
zeroliu, line 16: dbOps shouldn't be accessed before the initialization finishes

src/search/vectorStoreManager.ts
```diff
@@ -26,11 +26,18 @@ export default class VectorStoreManager {
     this.setupSettingsSubscription();
   }
 
+  static getInstance(): VectorStoreManager {
```
zeroliu, line 29: made this singleton. It was only instantiated once

src/search/vectorStoreManager.ts
```diff
@@ -52,6 +59,10 @@ export default class VectorStoreManager {
       ) {
         await this.eventHandler.updateExcludedFiles();
       }
+    };
+
+    subscribeToSettingsChange(() => {
+      this.initializationPromise = reinitialize();
```
zeroliu, line 65: we need to wait for initialization again when settings change causes the singleton to reinitialize

src/search/vectorStoreManager.ts
```diff
@@ -121,15 +132,16 @@ export default class VectorStoreManager {
     this.dbOps.onunload();
   }
 
-  public async getOrInitializeDb(embeddingsAPI: Embeddings): Promise<Orama<any>> {
```
zeroliu, line 124: We shouldn't have another place to initialize DB. `initialize` method be the one that instantiate db

src/search/hybridRetriever.ts
```diff
@@ -2,28 +2,21 @@ import { BrevilabsClient } from "@/LLMProviders/brevilabsClient";
...
 export class HybridRetriever extends BaseRetriever {
   public lc_namespace = ["hybrid_retriever"];
 
-  private llm: BaseLanguageModel;
   private queryRewritePrompt: ChatPromptTemplate;
-  private embeddingsInstance: Embeddings;
-  private brevilabsClient: BrevilabsClient;
 
   constructor(
-    private dbOps: DBOperations,
```
zeroliu, line 22: hybridRetriever becomes a lot easier to use after taking out these singleton instance args

logancyang:
Nice! Thanks for the cleanup! ✅

Issue #800, opened 2024-11-10
Show similar notes


zeroliu:
**Is your feature request related to a problem? Please describe.**
When working on a note, I often want to see similar notes about the same topic to get inspired. It helped me connect notes without manually linking them.

**Describe the solution you'd like**
We can show similar notes along with the suggested prompts. Both can inspire you to come up with prompts for the chat.

A possible mock can look like this:

![image](https://github.com/user-attachments/assets/e8d0075a-9a5e-4c04-8fc4-f2ab565713d3)

wwjCMP:
https://github.com/logancyang/obsidian-copilot/issues/257

PR #981, opened 2024-12-29
Relevant note new UI

zeroliu:
Introduce a new Relevant note feature! https://github.com/logancyang/obsidian-copilot/issues/800

One of Obsidian's selling points is connecting notes and building a knowledge network. However, manually creating connections with links requires a lot of effort. Even after years of practice, I find it challenging to organize my notes regularly. My notes are pretty discrete, even though many contents are "connected."

The indexing provided by Copilot is super valuable. There is an "ah-ha" moment when you work on a subject and realize you wrote about similar things a few years ago. Thanks to the embedding, this is possible. It connects the notes without users manually organizing them. Users can free their minds to create content, and the content will be automatically linked.

The Relevant Notes section has been added to surface these hidden links for you, making the Copilot plugin more powerful. 

The following screenshot shows cases of Elixir notes in my vault that are presented as relevant notes to a note talking about Elixir operations. The notes are ranked by relevance, considering content similarity, outgoing links, and backlinks. You can see badges under each note title to understand why a note is ranked higher than the others.

Screenshot 2024-12-29 at 11 53 42 AM

Once the chat starts, the section will be folded and kept at the top of the chat area. It allows users to look up relevant notes while remaining less intrusive to the chatting experience.

Screenshot 2024-12-29 at 11 56 05 AM

You can take two actions from the Relevant Note area: add the title to the chat input as a context and navigate to the note. Both actions are accessible from the expanded view and folded view.

Screenshot 2024-12-29 at 12 01 37 PM

For those who prefer a cleaner chat interface, turning off the feature from the settings is always possible.

Screenshot 2024-12-29 at 12 03 38 PM

Please feel free to provide feedback and help us iterate on this feature.


src/components/chat-components/RelevantNotes.tsx
```diff
@@ -0,0 +1,261 @@
...
+                </RelevantNotePopover>
+              ))}
+            </CardDescription>
+          )}
+        </CardHeader>
+        <CollapsibleContent>
+          <CardContent className="px-4 pb-2 pt-0 max-h-96 overflow-y-auto flex flex-col gap-2">
+            {relevantNotes.map((note) => (
+              <RelevantNote
+                showPath={!inSameFolder(activeFile?.path ?? "", note.document.path)}
```
zeroliu, line 298: We will only show the note path when the note is not in the same folder as the active note.

src/components/chat-components/SuggestedPrompts.tsx
zeroliu, line 1: Redesigned suggested prompt section to match the visual style of the new relevant notes

Screenshot 2024-12-29 at 12 06 36 PM

logancyang:
This levels up the plugin as a major feature addition! Congrats!! 🚀🚀🚀

Some initial comments that came to mind

- Should we have a "refresh" button on the new UI? When a user has no index initially it shows a warning, but after the index is built they have to switch to a new note to refresh it.
- Although we are not showing any numeric metric, it may still be confusing to the user why a "High Similarity" is ranked lower. The difference between our definition of "relevance" and "similarity" may not be immediately clear to the user. Should we add a ℹ icon with a hover "relevance is a combination of semantic similarity and links" (with possible future extension)
- I found this warning in the dev console, not sure if it's critical or not
SCR-20241229-mdbc

- The "Add to Chat" function is not working at the moment, it adds directly to the chat input but not the context. Currently you have to make an additional manual change to the chat input to trigger context update. My personal preference is to have "Add to Context" and not "Add to Chat". 
SCR-20241229-mcvi

- At the same time, we can directly add them as half-opaque candidates in the chat context menu similar to how Cursor does it. User can add it with a simple click on the half-opaque badge. We can always suggest 2, if one is clicked, add one more suggestion to keep always 2 suggested notes. (This can be a future PR)
SCR-20241229-mgti

- Some aesthetics (again my personal preference). The similarity or relevance metric badge here looks a bit too big and bold, even bigger and bolder than the note title. The path can have a lower opaqueness to differentiate with the title. And it's better to add some visual spacing between the path and the badges.
SCR-20241229-mcme

- Looks like it's not yet responsive on smaller screens? A bit cutoff on the right.
SCR-20241229-mjdp

logancyang:
Also right now if I scroll down with a long chat, this new UI is scrolled up and invisible. Do you think it's better to have it sticky at the top? 

I'm thinking potentially sharing this sticky UI with "Projects". "Projects" has its own tab beside "Relevant Notes", where the user can define a project name, index filters, system and custom prompts etc. Then it becomes a sticky "Copilot Menu" of sorts. 

The current design still leaves quite some space at the top.
SCR-20241229-mqkh

src/components/chat-components/RelevantNotes.tsx
```diff
@@ -0,0 +1,261 @@
...
+function inSameFolder(path1: string, path2: string) {
+  const folder1 = path1.split("/").slice(0, -1).join("/");
+  const folder2 = path2.split("/").slice(0, -1).join("/");
+  return folder1 === folder2;
+}
+
+function SimilarityBadge({ score }: { score: number }) {
+  let text = "🔴 Low Similarity";
+  if (score > 0.5) text = "🟠 Medium Similarity";
+  if (score > 0.7) text = "🟢 High Similarity";
```
logancyang: From my quick test with copilot-plus-small, ~80% of the similar notes fall in Medium Similarity, ~10+% High Similarity and rarely Low Similarity. Of course this distribution depends heavily on the embedding model used. Not really a problem but just to report an observation.

src/search/findRelevantNotes.ts
```diff
@@ -4,7 +4,7 @@ import { getSettings } from "@/settings/model";
 import { InternalTypedDocument, Orama, Result } from "@orama/orama";
 import { TFile } from "obsidian";
 
-const MIN_SIMILARITY_SCORE = 0.3;
+const MIN_SIMILARITY_SCORE = 0.4;
```
logancyang, line 7: Did you find 0.3 shows too many low similarity notes?

Right now if similarity ranges from 0.4->1, the weighted is 0.7*(0.4->1), then the lowest weighted similarity score is roughly 0.3 i.e. the highest of links only. Is this a design choice? For some notes they have so many similar notes, the linked notes are only visible when you scroll to the bottom. Again, just reporting an observation.

zeroliu: That's a great observation. I turned up the score because many seemingly unrelated notes showed up in the results. A single link will give 2.4 (0.8*0.3) after the last PR. 0.3 only applies to notes that have both ongoing and incoming links.

The weights and score threshold need more fine-tuning. I think this is something that we can keep iterating.

zeroliu: I just saw some very weird results. 

A high similar note (78%) is ranked lower than a low similar note (45) with outgoing links.
0.78 * 0.7 = 0.546
0.45 * 0.7 + 0.24 = 0.555

Let me think about these formula more closely.

zeroliu:
Love all the comments! Thanks so much for reviewing!

> This levels up the plugin as a major feature addition! Congrats!! 🚀🚀🚀
> 
> Some initial comments that came to mind
> 
> * Should we have a "refresh" button on the new UI? When a user has no index initially it shows a warning, but after the index is built they have to switch to a new note to refresh it.

Added the button. It's currently only reloading relevant notes. 

Does it also make sense to trigger indexing? The current indexing button on the chat input will refresh the index of the entire vault, which might be too intrusive for people who only want to reindex the current active file. This will allow them to reindex the file even when not in the qa vault or plus mode. It can be helpful to have multiple options. One in the relevant note and chat area to index the active file, and one in the new menu bar to index the whole vault. WDYT?

> * Although we are not showing any numeric metric, it may still be confusing to the user why a "High Similarity" is ranked lower. The difference between our definition of "relevance" and "similarity" may not be immediately clear to the user. Should we add a ℹ icon with a hover "relevance is a combination of semantic similarity and links" (with possible future extension)

Yeah, it can still confuse me sometimes. I added the information icon, but this needs more thought because it's not an intuitive UI if we need to explain it.

> * I found this warning in the dev console, not sure if it's critical or not
> 
> SCR-20241229-mdbc

This is fixed. Thanks for catching it. I need to create a separate PR to turn on React eslint to catch these automatically.

> * ~The "Add to Chat" function is not working at the moment, it adds directly to the chat input but not the context. Currently you have to make an additional manual change to the chat input to trigger context update. My personal preference is to have "Add to Context" and not "Add to Chat".~ Oh I realized this is for all modes and not just plus. "Add to Chat" is fine then. We need to do something in the Plus mode to make the context menu reactive to this correctly.
> 

> SCR-20241229-mcvi

I want to defer this to a follow-up PR. The root cause is that the context notes are not responsive to the chat input changes. There are some optimizations we can do in that code, too. I prefer not to expand the scope of this PR.


> * At the same time, we can directly add them as half-opaque candidates in the chat context menu similar to how Cursor does it. User can add it with a simple click on the half-opaque badge. We can always suggest 2, if one is clicked, add one more suggestion to keep always 2 suggested notes. (This can be a future PR)
> 
> SCR-20241229-mgti

Absolutely. Let's add it as a quick follow-up.


> * Some aesthetics (again my personal preference). The similarity or relevance metric badge here looks a bit too big and bold, even bigger and bolder than the note title. The path can have a lower opaqueness to differentiate with the title. And it's better to add some visual spacing between the path and the badges.
> 
> SCR-20241229-mcme

Fine-tuned the style. Here's what it looks like with the latest change:

Screenshot 2024-12-29 at 6 00 36 PM


> * Looks like it's not yet responsive on smaller screens? A bit cutoff on the right.
> 
> SCR-20241229-mjdp


How do you get to this state? The UI is responsive when I drag the right panel area. I see that sometimes when resizing the window, the right panel area size does not update correctly. That's more of an Obsidian bug since we rely on the width of the view to control the size of the plugin elements.

> if I scroll down in a long chat, this new UI is scrolled up and invisible. Do you think it's better to have it sticky at the top?

Yeah, I noticed that and explored this option. I also concluded that this would need a menu bar at the top to make the sticky header work well with the chat messages. Do you think this UI change is necessary for the first version? I also thought a non-sticky version would be less intrusive so we could test how users react to new elements in the chat UI.

zeroliu:
It's also OK to introduce the top bar as part of this PR so we won't frequently introduce too many visible UI changes to users. Not sure how sensitive Copilot users are to UI updates though. It would be nice we can incrementally roll out and test with small groups

logancyang:
1. "refresh" button can trigger "reindex current active file", I think this is a good approach. Plus mode already had a denounced "reindex current active file" after any modification, the other two modes can benefit from a manual trigger.

2. Yeah we can defer the "add to chat" context menu issue in plus to a future PR, that part needs some optimization for sure!

3. The updated aesthetic looks very nice!!!

4. I clicked this icon to get to mobile view, and it shows a cutoff on the right side.
SCR-20241229-schr

5. My two cents is that we could add it as a sticky menu in the first go, since if we add it as non-sticky and then update it to sticky it feels like two completely different "breaking change" experiences. If we make it sticky, even if we update it in the future it will still be the same sticky element. But at the moment even if it's collapsed, it still takes quite a lot of room with the note titles. As a sticky menu it can be nice to have a minimal single-row banner. WDYT?

zeroliu:
Sounds good. Let's try the sticky menu change. Shall we introduce this sticky menu to the no-message state as well? I think it would be more consistent to have one way of rendering relevant notes

zeroliu:
Does the UI respond correctly when you resize the window or right panel area?

logancyang:
> Shall we introduce this sticky menu to the no-message state as well? I think it would be more consistent to have one way of rendering relevant notes

Yes, consistency is important, it helps reduce uncertainty and mental load.

> Does the UI respond correctly when you resize the window or right panel area?

Yes resizing works perfectly!

zeroliu:
> I saw this and didn't understand what triggered it, do you know? SCR-20241229-sjht

Do you know if this is introduced by this PR? I can't tell from the compiled binary stacktrace. If you have a repro step, I can look into it.

zeroliu:
Pushed new changes!

- Relevant notes become a sticky header. (see screenshot)
  - In right panel view: Screenshot 2024-12-29 at 11 04 11 PM
  - In note area: 
    Screenshot 2024-12-29 at 11 04 01 PM
- Made the relevant note area use grid layout when the container is getting too large
  Screenshot 2024-12-29 at 11 04 34 PM
- Updated similarity category threshold. [0, 0.55, 0.7]. I collected a distribution of similarity scores of all the notes in my vault. The thresholds would be [0, 0.563, 0.688] for my vault to have an even distribution. We don't need to optimize my vault, so 0.55 and 0.7 are reasonable values that can be applied to regular users.
- Updated the ranking algorithm again. It now follows the following rules:
  - Notes of the lower category cannot rank higher than those of higher category. (e.g High similarity note will always show up on top of medium similarity note even if the first one is 0.71 and the second one is 0.69 + outgoing link + incoming link)
  - When notes are in the same category, sort them based on the score we had before.
This rule can avoid the confusion of similarity badge. It may still lead to confusion when people look into the actual score. It's fine because the score is for hardcore users, they can guess the logic based on the information tooltip.

logancyang:
Excellent! Can we make the collapsed menu even smaller with only 1 row of notes? That is my only comment now. Thinking about future "Project" tab that will show a single row "project name" when collapsed.

zeroliu:
Done!

PR #1006, opened 2025-01-05
Include context in user message

zeroliu:
Render context with the user message.

Screenshot 2025-01-05 at 12 50 54 AM
Screenshot 2025-01-05 at 12 51 45 AM

### DIFF
--- a/src/components/chat-components/ChatInput.tsx
+++ b/src/components/chat-components/ChatInput.tsx
@@ -21,7 +21,11 @@ import { TooltipActionButton } from "./TooltipActionButton";
 interface ChatInputProps {
   inputMessage: string;
   setInputMessage: (message: string) => void;
-  handleSendMessage: (toolCalls?: string[]) => void;
+  handleSendMessage: (metadata?: {
+    toolCalls?: string[];
+    urls?: string[];
+    contextNotes?: TFile[];
+  }) => void;
   isGenerating: boolean;
   onStopGenerating: () => void;
   app: App;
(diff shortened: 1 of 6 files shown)


src/components/chat-components/ChatInput.tsx
```diff
@@ -86,52 +90,19 @@ const ChatInput = forwardRef<{ focus: () => void }, ChatInputProps>(
       },
     }));
 
-    const debounce = <T extends (...args: any[]) => any>(
```
zeroliu, line 87: the debounce function and the effect that listens to input message change do the same thing. It caused duplicated context values. Tested that everything still works after removing all the debounce calls

zeroliu:
Note that editing doesn't work as intended yet. It will make people think the edited message is sent with context but in reality it doesn't. Not sure if we want to push this change before we fix the issue.

logancyang:
> Note that editing doesn't work as intended yet. It will make people think the edited message is sent with context but in reality it doesn't. Not sure if we want to push this change before we fix the issue.

Right, let me look into fixing the edit and this one can rebase on that.

zeroliu:
> > Note that editing doesn't work as intended yet. It will make people think the edited message is sent with context but in reality it doesn't. Not sure if we want to push this change before we fix the issue.
> 
> Right, let me look into fixing the edit and this one can rebase on that.

I can look into this, too. It will be related to the edit UI work that will follow up on this one.

logancyang:
> > > Note that editing doesn't work as intended yet. It will make people think the edited message is sent with context but in reality it doesn't. Not sure if we want to push this change before we fix the issue.
> > 
> > 
> > Right, let me look into fixing the edit and this one can rebase on that.
> 
> I can look into this, too. It will be related to the edit UI work that will follow up on this one.

Sounds good. Right now we have "hidden user message" that has added context in "handleSendMessage" in Chat.tsx. We also have some ad-hoc processing and appending to the user message using `contextProcessor`. We also have tool calls results appending. So it's a bit of a mess 😅 sorry about that.

I think we need a clear design of a new ChatMessage object with context in it (currently in `sharedState.ts`), and a new consolidated way for making changes to its context, be it from an edit, or a send time processing. It may touch a bunch of places like Chat.tsx, ChainRunners, and wherever that has logic "enhancing" the chat message.

logancyang:
One thing to be careful about is that any context should be fresh in case the context notes have changed. It is working correctly at the moment for chat input, but we should avoid saving a previous context in an old ChatMessage object.

zeroliu:
@logancyang I added the temporary fix for editing message as what we discussed on Friday. Not perfect but at least make it clear whether context is used in different scenario. The editing message has other issues, and I'd like to fix them in a separate PR.

logancyang:
Nice! I tested non-activenote context and they are working, but seems the active note badge doesn't get included?
SCR-20250112-nxxx

zeroliu:
> Nice! I tested non-activenote context and they are working, but seems the active note badge doesn't get included? SCR-20250112-nxxx

what is this active note feature? I thought you have to add `{activeNote}` keyword but I don't see it in your message

logancyang:
@zeroliu It's just the active note added to context menu manually
notebadge

zeroliu:
@logancyang the current note issue is fixed. Although I'm not sure about the value of having a separate active note. It would simplify the logic a lot if we only have context notes. Food for thoughts.

logancyang:
Pinged you in discord. There's a new issue where additional context in chat input don't get preserved after hitting send, only active note is preserved.

logancyang (approved):
LGTM! Just as a note, in the follow up PRs we need to address 

1. Edit with context (better design of the chat message object)
2. "Save as Note" and "Load" with context references

PR #1018, opened 2025-01-07
Use reranking on NaN chunks

logancyang:
If only NaN scores came back from Orama, apply reranking. (Only in PLUS mode).


logancyang:
@zeroliu I chatted with Jason and one common issue I see with huge vaults is having all NaN scores from Orama. I still don't have an answer to why that happens. The interesting thing is that although they are NaN, the returned chunks are actually relevant to the query. So as a temporary fix for the users, I'm calling reranking API for cases where there are only NaN chunks.

Issue #937, opened 2024-12-16
On Save error, custom prompt modal should return to custom prompt modal


i13-ctflora:
- [x] Disable all other plugins besides Copilot **(required)**
- [x] Screenshot of note + Copilot chat pane + dev console added **(required)**

Copilot version: 2.7.9

**Describe how to reproduce**
When saving a custom prompt, if there are illegal characters in the prompt name (such as ":") then an error is logged and you're kicked back to the main Obsidian screen.  This means you lose the custom prompt you've just entered.

**Expected behavior**
After the error, you should instead return to the modal, where your prompt name is highlighted with error styling.  Additionally, the prompt you've entered should still be there.  

Ideally, the error styling should happen immediately in the modal on entering the forbidden characters, or at least a note should be present indicating that these particular characters are not allowed.

**Screenshots**
<img width="970" alt="image" src="https://github.com/user-attachments/assets/23a8e185-b4a0-4219-8cbf-238a0140a953" />

**Additional context**
Not a huge deal, but pretty frustrating when you've entered a fairly long prompt without realizing that ":" were not acceptable in the name, and then having to re-do the prompt a second time.  
Also - the Obsidian notification that flashes up briefly indicates that the problem is another custom prompt of the same name, not the character problem which is only visible in the browser console.

i13-ctflora:
This one has been done, thank you!  Marking complete.

PR #1045, opened 2025-01-14
Some user experience optimizations

Emt-lin:
> To be continued...

1. optimize error handling.
  CleanShot 2025-01-14 at 18 31 35

2. optimize the style of PromptModal.
    CleanShot 2025-01-14 at 18 32 37

related: #937


logancyang:
Nice. I'm also thinking about the initial vectorstore failure UX at new user onboarding time. What do you think is a better experience?

Emt-lin:
> Nice. I'm also thinking about the initial vectorstore failure UX at new user onboarding time. What do you think is a better experience?

Let me think about it.

logancyang:
Can we make "enter to send" into a separate PR? Ideally one PR addresses one issue, easier to test and review. @Emt-lin

Emt-lin:
@logancyang 
> Can we make "enter to send" into a separate PR? Ideally one PR addresses one issue, easier to test and review

yeah, One small issue corresponds to one PR, which might have seemed like a waste of resoureces to me before.

Emt-lin:
@logancyang pls review the code.

logancyang:
Need some time. Will try to add this one to v2.8.4.

logancyang:
@Emt-lin in what cases do we see the first error message? For now we have a bug, e.g. if I don't have Anthropic API and tried switching to sonnet, it will show a notice that model switching failed, but the model dropdown still shows sonnet. When you chat with it, you are talking to your previous model. That should be fixed in a separate PR. But just trying to understand how we can arrive at that error message you showed in 1 with the current flow.

Emt-lin:
@logancyang When the error message contains `401|invalid|not valid` field, it will show the first error message.

CleanShot 2025-02-06 at 10 50 28

PR #1077, opened 2025-01-21
docs: update local copilot instructions for macOS

joshmedeski:
Hey @logancyang, thanks for a great plugin!

I updated the docs to include how to setup macOS without having to manually run the CLI command but rather update the launchctl env variable. I've tested and confirmed this works.


logancyang:
@joshmedeski thanks for the PR. Could you include the source of this info too.

PR #1085, opened 2025-01-22
Support drag-and-drop image

zeroliu:


logancyang:
What sorcery is this!! 💯 Works buttery smooth 🤌

Issue #1144, opened 2025-02-01
Improve test coverage for TImeTools.js


wenzhengjiang:
The current test cases are incomplete and do not closely match the time expression created from the backend.

PR #1145, opened 2025-02-01
Improve test cases for time range.

wenzhengjiang:
#1144

### DIFF
--- a/tests/TimeTools.test.ts
+++ b/tests/TimeTools.test.ts
@@ -116,15 +116,144 @@ describe("Time Expression Tests", () => {
 
   describe("Month Patterns", () => {
     test.each([
+      // Full month names
       {
-        expression: "january",
+        expression: "January",
         expected: { startDate: "2024-01-01", endDate: "2024-01-31" },
       },
       {
-        expression: "december 2023",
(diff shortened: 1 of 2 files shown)


tests/TimeTools.test.ts
```diff
@@ -25,255 +40,189 @@ describe("Time Expression Tests", () => {
...
+      {
+        expression: "december 2023",
+        expected: { startDate: "2023-12-01", endDate: "2023-12-31" },
+      },
+    ])("$expression", async ({ expression, expected }) => {
+      await verifyDateRange(expression, expected);
     });
   });
 
-  describe("Month Names", () => {
```
logancyang, line 99: Why is this removed? Seems no new tests are covering every month name like this. We must make sure all the month names and their abbr are working.

wenzhengjiang: We have "Month Patterns" tests in line 117. But they are only for the testing patterns not for all the values. Abbr was not added because the new prompt only return full names. But I guess there is no harm to add abbr tests.

wenzhengjiang: Done.

PR #1148, opened 2025-02-02
Avoid full vault scan on incremental indexing

zeroliu:
Previously, we added an incremental indexing feature for copilot plus users. On any note update, It generates the excluded file list and checks whether the current file should be updated. This process can take 5 seconds for large vaults and causes the plugin to be unusable in those extreme scenarios. This PR fixes the issue by stopping the generation of included/excluded file lists and matching the current file path against the matching patterns. 

- Added a new `shouldIndexFile` method that checks whether a TFile should be indexed. It consolidates all the custom validation logic that checks `inclusions` and `exclusions`.
- Updated the pattern-matching logic to use native obsidian API when possible.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -11,7 +11,7 @@ import { MemoryVariables } from "@langchain/core/memory";
 import { RunnableSequence } from "@langchain/core/runnables";
 import { BaseChain, RetrievalQAChain } from "langchain/chains";
 import moment from "moment";
-import { MarkdownView, Notice, TFile, Vault, parseYaml, requestUrl } from "obsidian";
+import { MarkdownView, Notice, TFile, Vault, requestUrl } from "obsidian";
 import { CustomModel } from "./aiParams";
 
 export const getModelNameFromKey = (modelKey: string): string => {
@@ -47,7 +47,7 @@ export async function getNoteFileFromTitle(vault: Vault, noteTitle: string): Pro
 }
 
 /** TODO: Rewrite with app.vault.getAbstractFileByPath() */
-export const getNotesFromPath = async (vault: Vault, path: string): Promise<TFile[]> => {
+export const getNotesFromPath = (vault: Vault, path: string): TFile[] => {
   const files = vault.getMarkdownFiles();
 
   // Special handling for the root path '/'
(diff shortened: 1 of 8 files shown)


tests/utils.test.ts
```diff
@@ -2,14 +2,68 @@ import * as Obsidian from "obsidian";
...
   extractNoteTitles,
-  getFilePathsFromPatterns,
   getNotesFromPath,
   getNotesFromTags,
   isFolderMatch,
   isPathInList,
   processVariableNameForNotePath,
 } from "../src/utils";
 
+// Mock Obsidian's TFile class
```
zeroliu, line 12: need to mock these because of the usage of app.vault.metadataCache

logancyang:
Thanks for fixing this! How did you test it?

zeroliu:
In terms of functionality, I tested inclusion and exclusion with tags, file extension, and single file match with `[[]]`. I observe whether `shouldIndexFile` returns true for the matching file path. I also tested different patterns with a full vault index and verified that the number of files indexed was correct.

In terms of performance, this part gets confusing. I created a vault with 40000 files. I was hoping to see the 5-second delay as described in the issue, but I couldn't reproduce it. As you can see in my screenshot, even after I added `,,,,,,,,,,,` as exclusion patterns that create a number of excluded strings. It still only takes ~40ms to run the file change handler. There must be a scenario that the user is in, but I can't reproduce it. My change shows that the change handler is now consistently faster, so this is still a good PR to merge. I hope it will resolve the reported performance issue, too.

Before:
Screenshot 2025-02-01 at 10 20 10 PM

After:
Screenshot 2025-02-01 at 10 14 53 PM

src/utils.ts
```diff
@@ -82,49 +82,60 @@ export const getNotesFromPath = async (vault: Vault, path: string): Promise<TFil
...
+export function stripHash(tag: string): string {
+  return tag.replace(/^#/, "").trim().toLowerCase();
+}
+
+/**
+ * @param file - The note file to get tags from.
+ * @param vault - The vault to get tags from.
+ * @returns An array of lowercase tags without the hash symbol.
+ */
+export function getTagsFromNote(file: TFile): string[] {
```
logancyang: This is now returning both inline and frontmatter tags, while before it only returns frontmatter tags. 

The reason I only get frontmatter tags is to 

1. Limit the number of notes from tags for custom prompt templating {#tag1, #tags2...} and avoiding overcrowding the context window
2. Have a differentiation between frontmatter tags and inline tags. Consider frontmatter tags a "note-level" property and inline tags as "line/block-level" property.

So this change will affect not only index filtering but also custom prompt processing. Currently, inline tags are used via salient terms during vault search.

Should we put a parameter here to control what tags to return? Such as `frontmatterOnly = true`

zeroliu: Updated

Issue #1029, opened 2025-01-09
Use Obsidian link instead of URL link in vault QA


LingJingMaster:
I propose that the Source feature in the application uses the native Obsidian Wikilink syntax [[name]] instead of the current [name](obsidian:vault) format. This change aligns with Obsidian’s native syntax and improves compatibility and readability for users accustomed to working directly within the Obsidian environment.

LingJingMaster:
The Regex：`\[(.*?)\]\(obsidian:\/\/[^\)]+\)`will be a good choice

logancyang:
@LingJingMaster Thanks for the suggestion! I was already looking into this a bit https://github.com/logancyang/obsidian-copilot/issues/764

The current markdown rendering doesn't render clickable wikilinks `[[]]` correctly. If anybody knows how to render clickable Wikilinks please let me know. The code is in `ChatSingleMessage` component, first time contributor welcome!

wwjCMP:
I think it's best to keep the current situation, otherwise, when preserving the conversation, a large number of invalid double links will be introduced.

iinkov:
https://github.com/logancyang/obsidian-copilot/pull/1153
Perhaps this is the fix

iinkov:
Ah, I see the issue now. In any case, it is better to have the links clickable in the chat response than shortened in the copied note.
Perhaps the right approach is to replace all links with the wikilink format during copying.

PR #1153, opened 2025-02-02
fix: Use [title](url) format for note titles

iinkov:
Make internal note links clicable in chat response
https://github.com/logancyang/obsidian-copilot/issues/1029?utm_source=chatgpt.com

### DIFF
--- a/src/components/chat-components/ChatSingleMessage.tsx
+++ b/src/components/chat-components/ChatSingleMessage.tsx
@@ -105,7 +105,7 @@ const ChatSingleMessage: React.FC<ChatSingleMessageProps> = ({
       );
 
       const clickable = (file: TFile) =>
-        `[${file.name}](obsidian://open?vault=${encodeURIComponent(app.vault.getName())}&file=${encodeURIComponent(file.name)})`;
+        `[${file.name}](obsidian://open?vault=${encodeURIComponent(app.vault.getName())}&file=${encodeURIComponent(file.path)})`;
       // Transform [[link]] format but exclude ![[]] image links
       const noteLinksProcessed = replaceLinks(
         noteImageProcessed,
--- a/src/components/chat-components/ChatSingleMessage.tsx
+++ b/src/components/chat-components/ChatSingleMessage.tsx
@@ -7,7 +7,7 @@ import { cn } from "@/lib/utils";
 import { ChatMessage } from "@/sharedState";
 import { insertIntoEditor } from "@/utils";
 import { Bot, User } from "lucide-react";
-import { App, Component, MarkdownRenderer } from "obsidian";
+import { App, Component, MarkdownRenderer, TFile } from "obsidian";
 import React, { useCallback, useEffect, useRef, useState } from "react";
 
 function MessageContext({ context }: { context: ChatMessage["context"] }) {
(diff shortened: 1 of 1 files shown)


src/constants.ts
```diff
@@ -13,7 +13,7 @@ export const DEFAULT_SYSTEM_PROMPT = `You are Obsidian Copilot, a helpful assist
   4. If the user mentions "@vault", it means the user wants you to search the Obsidian vault for information relevant to the query. The search results will be provided to you in the context. If there's no relevant information in the vault, just say so.
   5. If the user mentions any other tool with the @ symbol, check the context for their results. If nothing is found, just ignore the @ symbol in the query.
   6. Always use $'s instead of \\[ etc. for LaTeX equations.
-  7. When showing note titles, use [[title]] format and do not wrap them in \` \`.
+  7. When showing note titles, use [title](url) format and do not wrap them in \` \`.
```
logancyang: Have you tested this? Does it work for you? I don't see how the model can get the `url` from anywhere.

src/constants.ts
```diff
@@ -13,7 +13,7 @@ export const DEFAULT_SYSTEM_PROMPT = `You are Obsidian Copilot, a helpful assist
   4. If the user mentions "@vault", it means the user wants you to search the Obsidian vault for information relevant to the query. The search results will be provided to you in the context. If there's no relevant information in the vault, just say so.
   5. If the user mentions any other tool with the @ symbol, check the context for their results. If nothing is found, just ignore the @ symbol in the query.
   6. Always use $'s instead of \\[ etc. for LaTeX equations.
-  7. When showing note titles, use [[title]] format and do not wrap them in \` \`.
+  7. When showing note titles, use [title](obsidian://open?file=FILENAME) format and do not wrap them in \` \`.
```
logancyang: Again, have you tested this? Can you please show an example with this system prompt, what does it really return in the response.

logancyang:
I don't think hardcoding system prompt is the best way to solve this. The ideal solution is to modify the md renderer directly.

src/constants.ts
iinkov: It was hallucinating a lot, I changed it to this `[title](obsidian://open?file=FILENAME)`
Let me play with it for a week

src/constants.ts
iinkov: this is response

```
There is no note titled "My Note for Logan". There is a note titled [My Note](obsidian://open?file=My%20Note), but it is not specifically for Logan.

Sources:
```

iinkov: Yes I tested it with gemini

iinkov:
> I don't think hardcoding system prompt is the best way to solve this. The ideal solution is to modify the md renderer directly.

Yes, you are right. I did it in VaultQAChainRunner this time. However, it returns different formats of links. Perhaps some additional regex will need to be added. To get a wikilink for copying, back conversion could be done.

logancyang:
> > I don't think hardcoding system prompt is the best way to solve this. The ideal solution is to modify the md renderer directly.
> 
> Yes, you are right. I did it in VaultQAChainRunner this time. However, it returns different formats of links. Perhaps some additional regex will need to be added. To get a wikilink for copying, back conversion could be done.

I did have something like this in the past releases and some reported that they didn't like it.

Instead of making changes to the chain runner, we should update the `ChatSingleMessage` rendering directly with a post-processing step. That ensures it is not chain-dependent. And we make it a post-processing step just to make deep links as a display fix in AI responses, when the user copy/insert it to note, it should still have `[[]]` as it was.

logancyang:
@iinkov congrats on your first PR for Copilot! Great job, I already hear positive feedback for this! 🎉

PR #1154, opened 2025-02-02
Debounce settings input

zeroliu:
Improved UX for text/number/textarea/password inputs and slider

### DIFF
--- a/src/components/ui/password-input.tsx
+++ b/src/components/ui/password-input.tsx
@@ -18,25 +18,21 @@ export function PasswordInput({
   className?: string;
 }) {
   const [showPassword, setShowPassword] = useState(false);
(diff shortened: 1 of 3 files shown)


logancyang:
How does this improve the UX?

Here's some AI-generated questions

--- 

The shift from controlled to uncontrolled components across multiple files represents a significant architectural change. While this can improve performance, it may:
- Make form validation more complex
- Make it harder to implement features that need to react to immediate value changes
- Create inconsistency in how form state is managed across the application

Recommendations:
- Add documentation about the shift from controlled to uncontrolled components
- Consider adding a prop to toggle between controlled and uncontrolled behavior for more flexibility
- Add proper cleanup for the debounce timeouts when components unmount

PR #1160, opened 2025-02-04
Improve plus user onboarding

zeroliu:


src/LLMProviders/chatModelManager.ts
```diff
@@ -265,6 +265,7 @@ export default class ChatModelManager {
 
   async setChatModel(model: CustomModel): Promise<void> {
     const modelKey = getModelKeyFromModel(model);
+    setModelKey(modelKey);
```
zeroliu, line 268: @logancyang Attention! This change is critical. Otherwise, there is a race condition that can lead to infinite loop. This setter must be called before the async getModelConfig call.

src/main.ts
```diff
@@ -63,6 +64,7 @@ export default class CopilotPlugin extends Plugin {
     // Initialize BrevilabsClient
     this.brevilabsClient = BrevilabsClient.getInstance();
     this.brevilabsClient.setPluginVersion(this.manifest.version);
+    await checkIsPlusUser();
```
zeroliu, line 67: This function checks plus status on load

logancyang:
Awesome UX improvements! It could translate directly to more plus users!! 🔥🔥🔥 

Several observations
1. When the license key is invalid, this modal says it has expired. Could be better to say "either invalid or has expired" to be more precise.
SCR-20250203-plxi
2. After hitting "switch", the embedding model is switched to "copilot-plus-small". Normally, for a user who has an existing index, switching embedding model leads to a modal that triggers a complete reindex. But this "switch" doesn't. WDYT is a good solution here? Simpler solutions is better.
3. There's a merge conflict.

Otherwise LGTM! Look forward to the user feedback after it goes live!

zeroliu:
> After hitting "switch", the embedding model is switched to "copilot-plus-small". Normally, for a user who has an existing index, switching embedding model leads to a modal that triggers a complete reindex. But this "switch" doesn't. WDYT is a good solution here? Simpler solutions is better.

This is a great insight! It makes me wonder whether auto switch embedding model is a good idea. I think two changes are needed:

1. In the welcome modal, we should show the model to switch and highlight that changing embedding model will lead to a reindex.
2. When switching copilot plus status to false, we should not touch the embedding model to avoid accidentally reindexing the vault. If the user tries to index with the premium models, it will throw an error in the notice. We can provide a better error message in the future.

zeroliu:
Updated with my proposal:

1. in the plus welcome dialog, I added the models to switch and added a warning message if the current embedding model is not copilot-plus-small. Once the user clicks switch, we will reindex the embedding if the embedding model key changes.

Screenshot 2025-02-03 at 9 53 48 PM

2. in the plus expired dialog, I stopped switching chat and embedding models automatically. I added a warning message instead.
Screenshot 2025-02-03 at 9 54 16 PM

logancyang:
@zeroliu tried again and saw an issue: when a valid license key becomes invalid, the chat model still get updated to gpt-4o even though the default model value was not switched.

src/plusUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+    setChainType(DEFAULT_SETTINGS.defaultChainType);
+  }
+}
+
+export function turnOnPlus(): void {
+  const isPlusUser = getSettings().isPlusUser;
+  updatePlusUserSettings(true);
+  // Do not show the welcome modal if the user is already a plus user before
+  // 2024/02/04 (isPlusUser === undefined)
+  if (isPlusUser === false) {
```
logancyang: Should this be `isPlusUser !== true`? Right now this welcome modal is not showing up for new users (undefined).

zeroliu:
> @zeroliu tried again and saw an issue: when a valid license key becomes invalid, the chat model still get updated to gpt-4o even though the default model value was not switched.

This is fixed now.

> And if I click reset to default settings, the license key still exists. Is that intentional?

This is related to my last debounce and uncontrolled settings change. It's also fixed now.

> Should this be isPlusUser !== true? Right now this welcome modal is not showing up for new users (undefined).

isPlusUser set to undefined was intentional to hide the welcome dialog for the existing paid users. I overlooked the impact on users who clicks resetSettings. For new users, this should be fine because the onload event will set the value to false first. I changed the default value to false now and it will show to existing paid users once they upgrade. I think about it again and think it's an acceptable experience.

zeroliu:
@logancyang added another round of iteration with the following changes:

1. The welcome dialog will only show up after the user clicks "apply" in the settings. It won't affect existing users.
2. When is plus user check passes on load, it won't auto set the default chain type any more. This may disrupt existing users workflow. In the latest version, the onload check will only set the `isPlusUser` state and show a the key invalid dialog if the user goes from active to inactive. The current valid plus user will not be affected by this change.

PR #1203, opened 2025-02-05
Disable welcome modal

zeroliu:


logancyang:
One thing I noticed is that when a user has an issue and has to reset the settings or reinstall the plugin, even if they have a full index from the past for the copilot-plus-small, they will still be forced to run complete reindex on Switch after putting in the license key again. 

We can think more about the force reindex logic on switch. @zeroliu

PR #1207, opened 2025-02-06
refactor: optimize the switching experience of the model.

Emt-lin:
Actual style：

CleanShot 2025-02-06 at 16 19 20

### DIFF
--- a/src/components/chat-components/ChatInput.tsx
+++ b/src/components/chat-components/ChatInput.tsx
@@ -8,7 +8,8 @@ import { COPILOT_TOOL_NAMES } from "@/LLMProviders/intentAnalyzer";
 import { Mention } from "@/mentions/Mention";
 import { getModelKeyFromModel, useSettingsValue } from "@/settings/model";
 import { getToolDescription } from "@/tools/toolManager";
-import { extractNoteTitles } from "@/utils";
+import { err2String, extractNoteTitles } from "@/utils";
+import { DisplayKeyProviders, ProviderSettingsKeyMap } from "@/constants";
 import {
   ArrowBigUp,
   ChevronDown,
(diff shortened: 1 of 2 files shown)


logancyang:
This is nice! But don't we need to do the same for "Default Chat Model" in the setting? WDYT? Hide those without an API key? Probably a bit weird but we can discuss.

Emt-lin:
@logancyang wait review

src/components/chat-components/ChatInput.tsx
```diff
@@ -483,24 +486,59 @@ const ChatInput = forwardRef<{ focus: () => void }, ChatInputProps>(
...
+                            setCurrentModelKey(getModelKeyFromModel(model));
+                          } catch (error) {
+                            const msg = `Model switch failed: ` + err2String(error);
+                            setModelError(msg);
+                            new Notice(msg);
+                            // Restore to the last valid model
+                            const lastValidModel = settings.activeModels.find(
+                              (m) => getModelKeyFromModel(m) === currentModelKey
+                            );
+                            if (lastValidModel) {
```
logancyang, line 531: Can there be a case where there's no lastValidModel?

src/settings/v2/components/BasicSettings.tsx
```diff
@@ -158,6 +160,21 @@ const BasicSettings: React.FC<BasicSettingsProps> = ({ indexVaultToVectorStore }
...
             onChange={(value) => {
+              const selectedModel = settings.activeModels.find(
+                (m) => m.enabled && getModelKeyFromModel(m) === value
+              );
+              if (selectedModel) {
+                const providerKeyName =
+                  ProviderSettingsKeyMap[selectedModel.provider as DisplayKeyProviders];
+                const hasNoApiKey = !selectedModel.apiKey && !settings[providerKeyName];
+                if (hasNoApiKey) {
+                  const notice =
```
logancyang, line 171: This code is duplicated as in the other file. Can be reused as a util. Something to follow up on.

logancyang:
Added some comments, can be addressed in the next PR.

src/components/chat-components/ChatInput.tsx
Emt-lin: Currently, there should be none. only for robustness.

PR #1220, opened 2025-02-08
refactor: Improve source links formatting and rendering.

iinkov:
Here is a native-like view of internal and external links.

The source block is collapsed. It is huge in my case and not used frequently.


logancyang:
Could you attach a screenshot or short video showing how it looks?

iinkov:
image

logancyang:
@iinkov very nice! Would be a bit better if there's extra spacing between the response and sources.

iinkov:
Please note there is a known bug. AI swallows multi-spaces in note titles. I'll need to make a custom search function. It can't open notes with multiple spaces at the moment. It was before my changes.

iinkov:
image

Issue #1235, opened 2025-02-11
Add Custom Model modal always shows OpenAI API Key


i13-ctflora:
- [X] Disable all other plugins besides Copilot **(required)**
- [X] Screenshot of note + Copilot chat pane + dev console added **(required)**

Copilot version: 2.8.4

(Bug report without the above will be closed)

**Describe how to reproduce**
When I go to add a new Custom Model, the "Add Customer Chat Model" popup always shows the OpenAI key defined in the Add Keys/AI Provider Settings window.  The API Key remains even after changing from the "OpenAI" provider to some other provider.

Additionally, it doesn't look like the API Key value in the "Add Custom Chat Model" window is used.  I tried to add a custom Gemini model, entered the Gemini API key there (it was not already entered in the "Add Keys" window), and when I tried to verify it said that I didn't have an API Key defined.  When I went back and added the Gemini key in the "Add Keys" window, it worked, even though the OpenAI key was automatically filled in.

**Expected behavior**
If the API Key is supposed to fill in based on the provider, make sure it changes when changing the provider.  If this field should only be available for certain provider types (like OpenAI format), then make sure it's greyed out for other provider types.  Maybe also add a note/link to the "Add Keys" window if that's where the keys should be defined for certain provider types.

PR #1243, opened 2025-02-12
fix: Fixed the issue where the apikey could not be changed when switc…

Emt-lin:
related：#1235


src/settings/v2/components/ModelAddDialog.tsx
```diff
@@ -165,8 +165,6 @@ export const ModelAddDialog: React.FC<ModelAddDialogProps> = ({
     if (!isEmbeddingModel) {
       return {
         ...baseModel,
-        temperature: 0.1,
-        context: 1000,
```
logancyang, line 169: What is this "context" that was added in the new settings UI in aiParams? Missed it before.

Emt-lin: I will remove it.

PR #1261, opened 2025-02-17
Enhance inclusion/exclusion patterns settings

zeroliu:
New settings UI for inclusions and exclusions. It provides a more structured way to add extensions, tags, note titles, and folders. The new UI will provide suggested values for each category to reduce user errors when setting up inclusion and exclusion settings.

Performance concerns: The suggestion dialogs load all the files of the vault, which can be very slow for users with very large vaults. I added a "custom" option that allows people to add patterns in the old way. This user journey doesn't require loading files in the vault.

Enhanced edge case handling: Note and folder patterns support commas. It can cause issues with our previous way of splitting the patterns. In the latest version, the value is encoded before being stored in data.json.

Backward compatibility: The existing value in user's settings can work directly with this new settings UI. The new settings UI also removes empty value. There is a chance that existing invalid value will fall into "folder" as it catches all the patterns that don't match the other categories and there is no good way to validate folder pattern. I think this is fine because the new grouped view makes this error more clear.


src/components/ui/button.tsx
```diff
@@ -9,19 +9,18 @@ const buttonVariants = cva(
   {
     variants: {
       variant: {
-        default: "bg-interactive-accent text-on-accent shadow hover:bg-interactive-accent/90",
+        default:
+          "mod-cta bg-interactive-accent text-on-accent shadow hover:bg-interactive-accent-hover",
         destructive: "bg-modifier-error text-normal shadow-sm hover:bg-modifier-error/90",
-        outline:
```
zeroliu, line 14: our button styles were confusing. Default looked identical to secondary, and outline doesn't render an outline button but more of a secondary button. I cleaned them up in this PR.

src/components/ui/dialog.tsx
```diff
@@ -45,7 +45,7 @@ const DialogContent = React.forwardRef<
       {...props}
     >
       {children}
-      <DialogPrimitive.Close className="absolute right-4 top-4 rounded-sm opacity-70 ring-offset-ring transition-opacity hover:opacity-100 focus:outline-none focus:ring-2 focus:ring-ring focus:ring-offset-2 disabled:pointer-events-none data-[state=open]:bg-interactive-accent data-[state=open]:text-muted">
+      <DialogPrimitive.Close className="absolute border-none right-4 top-4 text-faint clickable-icon bg-transparent hover:bg-opacity-100 hover:text-normal hover:bg-transparent outline-none focus-visible:outline-none focus-visible:text-normal focus-visible:ring-0">
```
zeroliu, line 48: The dialog close button X had a weird background and border. I made them look more aligned with the obsidian dialog

src/search/indexOperations.ts
```diff
@@ -367,24 +372,25 @@ export class IndexOperations {
   private updateIndexingNoticeMessage(): void {
     if (this.state.indexNoticeMessage) {
       const status = this.state.isIndexingPaused ? " (Paused)" : "";
+      const messages = [
```
zeroliu, line 375: This needs to update as I started to encode the inclusion/exclusion strings

src/search/searchUtils.ts
```diff
@@ -106,55 +114,170 @@ export function shouldIndexFile(file: TFile, inclusions: string[], exclusions: s
...
+
+  // Normalize path separators to forward slashes to ensure cross-platform compatibility
+  const normalizedFilePath = filePath.replace(/\\/g, "/");
+
+  return folderPatterns.some((pattern) => {
+    // Normalize pattern path separators and remove trailing slashes
+    const normalizedPattern = pattern.replace(/\\/g, "/").replace(/\/$/, "");
+
+    // Check if the path starts with the pattern
+    return (
```
zeroliu, line 249: This is slightly different from the previous InPathFileList function behavior, which it not only checked file path starts with the folder pattern, but also includes and ends with the folder pattern. I make it only keeps the startWith because it aligns with obsidian app folder exclusion behavior

src/settings/v2/components/ApiKeyDialog.tsx
```diff
@@ -175,15 +175,12 @@ const ApiKeyDialog: React.FC<ApiKeyDialogProps> = ({
...
+                        variant="secondary"
                         size="sm"
                         className="w-full whitespace-nowrap"
                       >
                         {verifyingProviders.has(item.provider) ? (
-                          <>
-                            <Loader2 className="mr-2 h-4 w-4 animate-spin" />
-                            Verify
-                          </>
+                          <Loader2 className="mr-2 h-4 w-4 animate-spin" />
```
zeroliu, line 164: having loading next to the verify looks pretty odd

src/settings/v2/components/QASettings.tsx
```diff
@@ -49,62 +36,47 @@ const QASettings: React.FC<QASettingsProps> = ({ indexVaultToVectorStore }) => {
...
-                    side="bottom"
-                    align="center"
-                    sideOffset={0}
-                    onMouseEnter={() => handlePopoverOpen("index-help")}
-                    onMouseLeave={() => handlePopoverClose("index-help")}
-                  >
-                    <div className="space-y-2 sm:space-y-2.5">
-                      <div className="rounded bg-callout-warning/10 p-1.5 sm:p-2 ring-ring">
-                        <p className="text-callout-warning text-xs sm:text-sm">
+                <TooltipProvider delayDuration={0}>
```
zeroliu, line 39: This tooltip used the wrong component and looked pretty odd. Cleaned it up as I'm updating QA settings.

src/settings/v2/components/QASettings.tsx
```diff
@@ -186,35 +158,58 @@ const QASettings: React.FC<QASettingsProps> = ({ indexVaultToVectorStore }) => {
...
+              Manage
+            </Button>
+          </SettingItem>
 
           {/* Inclusions */}
           <SettingItem
-            type="textarea"
+            type="custom"
             title="Inclusions"
-            description="When specified, ONLY these paths, tags, or note titles will be indexed (comma separated). Takes precedence over exclusions. Files which were previously indexed will remain in the index unless you force re-index. Format: folder1, folder1/folder2, #tag1, #tag2, [[note1]], [[note2]]"
```
zeroliu, line 213: The original message says inclusion takes precedence over exclusions, but I think it's the opposite. For example, if foo/ folder is in both inclusions and exclusions, foo/ will be excluded.

logancyang:
Generally looking good! 

1. It can benefit from some test cases focusing on edge cases such as folders with spaces `100 some folder name` which I see a lot from our users. Adding some edge cases that we do not support at the moment in the test cases can be helpful too.

2. I notice a significant pause when triggering indexing, there could be performance optimization needed somewhere? For example, here's a suggestion by cursor in case it's partially helpful:

### Current Performance Bottleneck

```typescript:src/search/searchUtils.ts
function matchFilePathWithPatterns(filePath: string, patterns: string[]): boolean {
  if (!patterns || patterns.length === 0) return false;

  const { tagPatterns, extensionPatterns, folderPatterns, notePatterns } =
    categorizePatterns(patterns);  // <-- Main bottleneck

  return (
    matchFilePathWithTags(filePath, tagPatterns) ||
    matchFilePathWithExtensions(filePath, extensionPatterns) ||
    matchFilePathWithFolders(filePath, folderPatterns) ||
    matchFilePathWithNotes(filePath, notePatterns)
  );
}
```

### Optimization Strategy

1. **Optimize Pattern Categorization**:
```typescript:src/search/searchUtils.ts
function matchFilePathWithPatterns(filePath: string, patterns: string[]): boolean {
  if (!patterns || patterns.length === 0) return false;

  // Fast-path for extension matching
  const fileExtension = filePath.split(".").pop()?.toLowerCase() || "";
  for (const pattern of patterns) {
    if (pattern.startsWith("*.") && pattern.slice(2) === fileExtension) {
      return true;
    }
  }

  // Fast-path for folder matching
  const normalizedPath = filePath.replace(/\\/g, "/");
  for (const pattern of patterns) {
    if (!pattern.startsWith("#") && !pattern.startsWith("*.") && !pattern.startsWith("[[")) {
      if (normalizedPath.startsWith(pattern) && 
          (normalizedPath.length === pattern.length || normalizedPath[pattern.length] === "/")) {
        return true;
      }
    }
  }

  // Only categorize remaining patterns if necessary
  const remainingPatterns = patterns.filter(p => 
    p.startsWith("#") || p.startsWith("[[")
  );
  
  if (remainingPatterns.length === 0) return false;

  const { tagPatterns, notePatterns } = categorizePatterns(remainingPatterns);
  
  return (
    matchFilePathWithTags(filePath, tagPatterns) ||
    matchFilePathWithNotes(filePath, notePatterns)
  );
}
```

2. **Optimize Individual Matchers**:
```typescript:src/search/searchUtils.ts
function matchFilePathWithTags(filePath: string, tagPatterns: string[]): boolean {
  if (!tagPatterns || tagPatterns.length === 0) return false;

  const file = app.vault.getAbstractFileByPath(filePath);
  if (!(file instanceof TFile)) return false;

  // Get tags only once
  const tags = getTagsFromNote(file);
  const strippedTags = new Set(tags);
  
  return tagPatterns.some(pattern => strippedTags.has(stripHash(pattern)));
}

function matchFilePathWithNotes(filePath: string, notePatterns: string[]): boolean {
  if (!notePatterns || notePatterns.length === 0) return false;

  const basename = filePath.split("/").pop()?.split(".")[0];
  if (!basename) return false;

  return notePatterns.some(title => title.slice(2, -2) === basename);
}
```

4. **Batch Processing**:
```typescript:src/search/indexOperations.ts
private async getFilesToIndex(overwrite?: boolean): Promise<TFile[]> {
  const { inclusions, exclusions } = getMatchingPatterns();
  const allMarkdownFiles = this.app.vault.getMarkdownFiles();
  const BATCH_SIZE = 100;
  const filesToIndex = new Set<TFile>();

  for (let i = 0; i < allMarkdownFiles.length; i += BATCH_SIZE) {
    const batch = allMarkdownFiles.slice(i, i + BATCH_SIZE);
    
    // Process batch
    for (const file of batch) {
      if (!shouldIndexFile(file, inclusions, exclusions)) continue;
      
      const content = await this.app.vault.cachedRead(file);
      if (!content?.trim()) continue;
      
      filesToIndex.add(file);
    }

    // Let UI breathe between batches
    await new Promise(resolve => setTimeout(resolve, 0));
  }

  return Array.from(filesToIndex);
}
```

These optimizations should significantly improve performance by:
1. Avoiding unnecessary pattern categorization
2. Using fast-path checks for common patterns
3. Optimizing string operations
5. Processing files in batches
6. Allowing UI updates between batches

zeroliu:
> Generally looking good!
> 
> 1. It can benefit from some test cases focusing on edge cases such as folders with spaces `100 some folder name`, `200 some other/300 folder name` which I see a lot from our users. Adding some edge cases that we do not support at the moment in the test cases can be helpful too.

There are already test cases covering space and slashes. I added a few more just in case.

> 2. A UI nit: long items get cutoff and needs horizontal scrolling to reveal the `x`
> 
> SCR-20250218-kwdu

Good catch. Fixed!

> 3. I notice a significant pause when triggering indexing, there could be performance optimization needed somewhere? For example, here's a suggestion by cursor in case it's partially helpful:

I updated the code to only call categorizePatterns once. The rest functions are fairly lightweight (complete within 1ms) and I don't see the suggested fix make much differences. I ran a benchmark on my vault with 10000 notes with inclusion and exclusion rules across all categories. I only see a 20ms delay before the indexing kicks in. Can you elaborate on your significant pause and how to reproduce it?
Screenshot 2025-02-18 at 10 18 18 PM

logancyang:
@zeroliu right now when i run "force reindex" when having nothing in incl/excl filters, it says 0 files to be indexed.

src/search/searchUtils.ts
```diff
@@ -75,86 +89,209 @@ function getInclusionPatterns(): string[] {
...
+export function shouldIndexFile(
+  file: TFile,
+  inclusions: PatternCategory | null,
+  exclusions: PatternCategory | null
+): boolean {
+  if (exclusions && matchFilePathWithPatterns(file.path, exclusions)) {
     return false;
   }
-  if (inclusions.length > 0 && !matchFilePathWithPatterns(file.path, inclusions)) {
+  if (inclusions && !matchFilePathWithPatterns(file.path, inclusions)) {
```
logancyang, line 127: Is this check correct? `exclusions` and `inclusions` now exist even when the properties are all empty arrays. Now running "force re-index" with nothing in inclusion shows "0 file to index".

zeroliu: The 0 file to index issue was caused by another function, which I added test to make sure it returns null. Here the inclusions and exclusions value has changed from the original list to PatternCategory object because of the performance optimization to avoid running categorize function for every file.

zeroliu:
> @zeroliu right now when i run "force reindex" when having nothing in incl/excl filters, it says 0 files to be indexed.

good catch. The performance optimization caused a regression here. Added a new unit test to cover the proper return value. It would be nice to have a more integrated test but it's pretty challenging to set up with the current environment.

logancyang:
Awesome! Thanks for implementing this so quickly. This is a large one!

PR #1262, opened 2025-02-17
Remove language specific prompt in command prompts

zeroliu:


logancyang:
Just as a note, we should communicate to the user somewhere "Prompt effectiveness depends on the model. Make your own custom prompt if this is not working for you."

zeroliu:
> Just as a note, we should communicate to the user somewhere "Prompt effectiveness depends on the model. Make your own custom prompt if this is not working for you."

We haven't made the custom prompt to work like the other system command. We can make the change first and suggest it once it's shipped

Issue #1265, opened 2025-02-18
Using the reasoning model does not return any response for QA


wwjCMP:
Using the reasoning  model does not return any answer; it only returns reference information.

I use  deepseek-r1:14b-qwen-distill-q4_K_M model

logancyang:
Please provide necessary info in the issue template.

Is this Ollama or LM Studio?

wwjCMP:
I use Ollama.

![Image](https://github.com/user-attachments/assets/c0f1007d-d5c6-4676-bf8b-ccfd88587c4c)

![Image](https://github.com/user-attachments/assets/9bf6bcb1-580c-4832-b51d-b6a955d5cf01)

logancyang:
Working for me

https://github.com/user-attachments/assets/c0469ac7-9170-4cfc-bad4-63c2c41ab6cf

wwjCMP:
> Working for me
> 
>  ollamar1.mp4

try vault QA

logancyang:
OK fix up and will be merged soon.

Issue #1113, opened 2025-01-26
Show and save `thinking` block for reasoning models


logancyang:
Reasoning models:

- DeepSeek R1
- Gemini 2.0 flash exp thinking
- OpenAI o1 and o3

There needs to be a common interface for them.

AvijeetPrasad:
Looking forward to this implementation. When using the `deepseek-r1:14b` model with `ollama` as provider, I noticed that for my simple prompt: "quiz me on the {active note}", it provides a nice long summary of the note, but forgets to quiz me in the end. So I guessing these models need to be prompted internally in a different way. I am attaching the output file for your reference.

[20250205_075342_quiz_me_on_the_{activeNote}..md](https://github.com/user-attachments/files/18667890/20250205_075342_quiz_me_on_the_.activeNote.md)

liujuanjuan1984:
补：Gemini 2.0 flash exp thinking 不开放思考内容给 api  https://ai.google.dev/gemini-api/docs/thinking?hl=zh-cn

webguest:
When calling Deepseek for translation, how to hide <think>

----------------------------------------------------------------------------
Hello,world.<think>
Alright, let's tackle this translation task. First, I need to understand the instructions clearly. The user wants me to translate "Hello, world." into Chinese while preserving the meaning and tone. That means I shouldn't just do a direct word-for-word translation; I should capture the essence of the greeting.

Next, maintaining the appropriate cultural context is important. In English, "hello" is a common greeting, but in Chinese, there are several ways to say hello depending on formality and region. However, "你好 (nǐ hǎo)" is widely recognized and polite, making it the best choice for this translation.

I also need to keep the formatting and structure intact. The original text has "Hello," followed by a comma and "world." In Chinese, punctuation works similarly, so I'll follow that structure: "你好，世界。"

Additionally, the user specified to return only the translated text without any additional explanations or markdown formatting. So, after translating, I should ensure there's no extra content attached.

Lastly, considering the simplicity of the sentence, it's crucial to keep the translation straightforward and natural, avoiding overly literal translations that might sound awkward in Chinese.

In summary, by focusing on meaning preservation, cultural appropriateness, and maintaining the original structure, I can provide an accurate and suitable translation for "Hello, world." in Chinese.
</think>

你好，世界。

Jeff-Hugh:
Please make the thinking part and response part different. Now two parts have the same font format.

CiaranG:
[...]

PR #1266, opened 2025-02-18
feat: Add support for rendering `<think>` sections and fix RAG with reasoning models

logancyang:
Related: #1113, #1265 

https://github.com/user-attachments/assets/2abf2fae-e804-443b-9405-a3e2506beee5

### DIFF
--- a/src/chainFactory.ts
+++ b/src/chainFactory.ts
@@ -5,6 +5,7 @@ import { BaseRetriever } from "@langchain/core/retrievers";
 import { RunnablePassthrough, RunnableSequence } from "@langchain/core/runnables";
 import { BaseChatMemory } from "langchain/memory";
 import { formatDocumentsAsString } from "langchain/util/document";
+import { removeThinkTags } from "./utils";
 
 export interface LLMChainInput {
   llm: BaseLanguageModel;
@@ -177,8 +178,9 @@ Question: {question}
       llm,
       new StringOutputParser(),
       (output) => {
-        if (debug) console.log("Standalone Question: ", output);
-        return output;
+        const cleanedOutput = removeThinkTags(output);
+        if (debug) console.log("Standalone Question: ", cleanedOutput);
+        return cleanedOutput;
       },
     ]);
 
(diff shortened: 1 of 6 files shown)


i13-ctflora:
This is a great addition, but I've got one question - in your video above, you show the 'thinking' block come up immediately and stream the output as it's received.  When I tried this with a few thinking models, I got some odd results.  Perplexity's sonar-reasoning models did not stream the thinking or result blocks in real-time, instead showing the results in full at the end.  I just submitted bug #1285 about a ton of console warning messages for the sonar models, so not sure if that has anything to do with it.  

With some local Ollama models (like Deepseek R1 8b), I did see the streaming "thinking" output. Right now, I don't have any other reasoning models to try (don't have access to o3, and the experimental gemini thinking model doesn't return the think block in the API).

logancyang:
@i13-ctflora when it comes to DeepSeek R1 related models, many model providers don't support streaming or have CORS issue with streaming, e.g. pplx reasoner and Azure DS R1. Ollama R1 variants support streaming without issue as you mentioned. There's also an issue with the langchainjs DeepSeek client that prevents getting the thinking tokens from the official DS API at the moment, I will follow up with them.

amanat361:
Any idea if this can be updated to toggle reasoning for a model like Claude 3.7 Sonnet that has two options?

scriptgenerator64:
This is no longer working in the latest update on my setup

ichts:
> This is no longer working in the latest update on my setup

Could you please open an issue?

logancyang:
> This is no longer working in the latest update on my setup

Does your model setting have "reasoning" checked?

scriptgenerator64:
> > This is no longer working in the latest update on my setup
> 
> Does your model setting have "reasoning" checked?

Yes, I don’t have a PC handy to open an official issue right now but basically any of the ollama reasoning models (installed via “ollama run <model>”, specifically deepseek-r1 and gpt-oss) show this kind of message. I even created a brand new fresh vault and was able to replicate the behaviour.

By the way, you’re awesome for responding so quickly, I’ll stay back on the old version for now, but I have no doubt that this will be resolved in time 🙂

IMG_4591

I checked, and the issue is present at least since 3.1.1

scriptgenerator64:
> > This is no longer working in the latest update on my setup
> 
> Could you please open an issue?

Just got back, here's the issue:
https://github.com/logancyang/obsidian-copilot/issues/2030

PR #1269, opened 2025-02-18
implement model switching for custom prompts

### LINKED ISSUE #1146: Have an optional model field in custom prompt
A prompt that works for one model may not work for another.

Add an optional model field in Copilot Custom Prompt and selecting it should automatically set the model.

aviatesk:
Add support for both permanent and temporary model switching in custom prompts using frontmatter configuration. Users can now specify models using either 'model:' for permanent switches or 'temp-model:' for temporary switches.

This change improves the flexibility of custom prompts by allowing users to switch models contextually while maintaining their current model setting.

- closes #1146

### DIFF
--- a/src/LLMProviders/chatModelManager.ts
+++ b/src/LLMProviders/chatModelManager.ts
@@ -390,4 +390,10 @@ export default class ChatModelManager {
     const settings = getSettings();
     return settings.activeModels.find((model) => model.name === modelName);
   }
+
+  getCurrentModel(): CustomModel | null {
+    const currentModelKey = getModelKey();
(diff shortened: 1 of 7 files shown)


logancyang:
Thanks for the PR! There are some major changes, sorry this didn't make it in because it should be part of the redesign. If you still have time and interest, could you check if this can be merged into the `2.9.0-preview` branch?

swoh816:
I'm so much looking forward to having this feature implemented!!

logancyang:
Custom prompt is now "custom command" with a new window. Please let me know if it doesn't solve your issue. Closing now.

PR #1306, opened 2025-02-26
fix: replacing Node's Buffer with npm's buffer package.

Emt-lin:
All platform both use buffer package.

npm's buffer package: https://www.npmjs.com/package/buffer


logancyang:
So you've tested with mobile and this works? Nice!

Emt-lin:
> So you've tested with mobile and this works? Nice!

Perhaps you should test whether the old encryption key can work normally.

Issue #1128, opened 2025-01-28
[PLUS] Copilot Composer


logancyang:
Some relevant use cases

> The primary use case emerges when starting work on an idea. The process involves online research, LLM conversations, and documenting thoughts. Information is typically documented in bullet points, including images and links discovered during research. Once the data is gathered and processed, Composer helps rewrite it in a structured format. The content undergoes continuous refinement until it meets the standards of a sapling note.
> 
> Composer assists in populating various YAML fields and tags based on values already present in the vault. Cursorrules contains template information for different types of notes and permissible YAML field values. Using this information, Composer suggests appropriate YAML for the current note under development.
> 
> The tool also recommends suitable folders within the vault structure for filing the note.
> 
> Interlinking often gets overlooked, but Composer helps identify similar ideas in the vault and adjusts the current note's content to enable linking to those notes. It also analyzes whether other notes can or should link back to the current note, making relevant edits accordingly.
> 
> In cases where manual research and external webpage linking seems tedious (such as when writing about Dataview and needing links to its documentation), Cursor's composer (using Perplexity's API) automatically discovers URLs and incorporates them in appropriate locations. This applies to multiple concepts within a single note.

logancyang:
Created this master task to curate ideas, feel free to add more ideas to guide our design later. cc @zeroliu @Emt-lin @wenzhengjiang

jonashaefele:
I posted this in Discord, but it might be more useful here. It's a simplified use case, but it feels very related to this idea.
It's not the same, but feels very overlapping.  From how I understand Composer it might reformat a note, I was looking at creating a new note from a source note. Both approaches would need some sort of context gathering, importing of frontmatter structure, tone of voice etc..

---

I just tried to update a bit of my content creation workflow. I set it up and almost finished it last summer using a combination of custom scripts and SmartConnections prompts, but it's been sitting there unused, since I got sidetracked with work and uni.

Basically, I'm trying to create social media content drafts for my side projects from my voice notes. The prompts I created last year included things like context, brand voice, etc based on tags on the voice note. (e.g. tag #brandA on the voice note pulls in tone of voice and brand descriptions for that brand). That needs some sort of customised, on-the-fly prompt.

I tried to think about how to achieve that with Copilot. I'd like to limit complex scripts that can break. I was thinking something like:
- open source note (voice note) in editor
- trigger custom copilot prompt on active note, maybe read {selection} to steer the prompt/focus on a part
[...]

PR #1315, opened 2025-03-01
Implement composer

wenzhengjiang:
#1128 
This is quite large PR containing below major changes:
1. Appending composer system prompt for copilot-plus mode.
2. Rendering composer markdown blocks by adding a note path title and the apply button.
3. Implement apply function by calling the composer apply endpoint and creating the Apply view.


src/plusUtils.ts
```diff
@@ -80,3 +93,28 @@ export function turnOffPlus(): void {
...
+  if (cachedComposerPrompt) {
+    return `${currentSystemPrompt}\n${cachedComposerPrompt}`;
+  }
+
+  // Otherwise, fetch it from the API
+  const brevilabsClient = BrevilabsClient.getInstance();
+  try {
+    // Get the composer prompt from the API
+    const composerPromptResponse = await brevilabsClient.composerPrompt();
+    cachedComposerPrompt = composerPromptResponse.prompt;
```
logancyang: Should we persist this on disk just like how `/pdf4llm` cache works? Just need the simplest invalidation mechanism for now. Ideally we need every way possible to make it faster.

logancyang (commented):
Awesome job 🔥 Congrats! 🎉

Some quick observations:

- The full path and the buttons in this top section looks pretty big overall, might be better to make them smaller?
SCR-20250303-ppzt

- After clicking "Apply", there's a noticeable lag before the diff view is opened. It's fine to have a lag but it would be a nice UX improvement to **add a spinner** on the Apply button so that it doesn't look like nothing is happening for a second after clicking it.

- When there's nothing accepted, I click "Apply Accepted Changes" it just closes the diff view without anything changed. This is reasonable function-wise but a bit awkward on the UX side - user might think it's "Apply All" or they didn't realize they didn't accept anything. Is it better to have an "Apply all / or accepted changes", if nothing is accepted explicitly, it just acts like "Apply all"; if anything accepted, it applies those. Open to other ideas.

- When there's only one diff block with Accept / Reject, is it better to make the block-level Accept / Reject act like the global ones and close the diff view? It seems unnecessary to click the top ones again for a single block.

WDYT? @wenzhengjiang

src/plusUtils.ts
wenzhengjiang: I won't worry about the extra latency of call. /broca plus /chat/completion take ~5000ms, while this prompt fetch will be no more than 50ms.

zeroliu:
> * user might think it's "Apply All" or they didn't realize they didn't accept anything. Is it better to have an "Apply all / or accepted changes", if nothing is accepted explicitly, it just acts like "Apply all"; if anything accepted, it applies those. Open to other ideas.

I was thinking the exact same thing when I first saw the demo. The UX of accepting diff has been pretty standardized. The same pattern has been used not only for the latest AI editing, but also has been like that since the git diff experience. I would highly recommend us to mimic a similar experience.

package.json
```diff
@@ -98,12 +98,14 @@
     "@radix-ui/react-switch": "^1.1.1",
     "@radix-ui/react-tooltip": "^1.1.6",
     "@tabler/icons-react": "^2.14.0",
+    "@types/diff": "^7.0.1",
```
zeroliu: this should be a dev dependency

src/LLMProviders/brevilabsClient.ts
```diff
@@ -194,4 +215,12 @@ export class BrevilabsClient {
...
     return this.makeRequest<Youtube4llmResponse>("/youtube4llm", { url });
   }
+
+  async composerPrompt(): Promise<ComposerPromptResponse> {
+    return this.makeRequest<ComposerPromptResponse>("/composer/prompt", {}, "GET");
+  }
+
+  async composerApply(request: ComposerApplyRequest): Promise<ComposerApplyResponse> {
+    return this.makeRequest<ComposerApplyResponse>("/composer/apply", request);
+  }
```
zeroliu: The latest brevilabs client changed makeRequest to return both data and error. The wrapper functions here need to handle the error too

src/styles/tailwind.css
```diff
@@ -872,3 +896,125 @@ If your plugin does not need CSS, delete this file.
   color: var(--text-error);
   opacity: 1;
 }
+
+/* Apply View Styles */
+.apply-view-container {
```
zeroliu: We already standardized styling to tailwindcss. All existing CSS class names only exist for legacy reasons. No changes should be introduced to tailwind.css in this PR.

src/plusUtils.ts
zeroliu: Is there any reason why this prompt needs to come from the server? I assume the server doesn't do any extra processing as it doesn't take any argument from the client?

src/components/chat-components/ChatSingleMessage.tsx
```diff
@@ -203,6 +296,85 @@ const ChatSingleMessage: React.FC<ChatSingleMessageProps> = ({
         "", // Empty string for sourcePath as we don't have a specific source file
         componentRef.current
       );
+
+      // Function to add apply buttons to code blocks
+      const addApplyButtonsToCodeBlocks = () => {
+        if (!contentRef.current) return;
+
+        // Find all pre elements (code blocks) in the rendered content
+        const codeBlocks = contentRef.current.querySelectorAll("pre");
```
zeroliu: interacting with the raw DOM element in the react component can be very error-prone. Is there any reason why it needed to be implemented this instead of relying on React functions?

wenzhengjiang:
Thanks for the good UX feedbacks! 
* For the top section, I also plan to remove the status bar. I feel like it's not adding much value.
* For the top apply button,  let me rename the "Apply Accepted Changes" button to "Apply changes"? And it will apply accepted or unselected changes.

> When there's only one diff block with Accept / Reject, is it better to make the block-level Accept / Reject act like the global ones and close the diff view? It seems unnecessary to click the top ones again for a single block.
Or they could just click the top ones since there is only one block? Not sure we want to change the behavior of the buttons for this specific case.

src/components/ApplyView.tsx
```diff
@@ -0,0 +1,351 @@
...
+          const isAccepted = hasChanges && block.every((change) => change.accepted);
+          const isRejected = hasChanges && block.every((change) => change.rejected);
+
+          return (
+            <div
+              key={blockIndex}
+              className={cn("diff-block mb-4 border rounded-md overflow-hidden", {
+                "border-green-500 accepted": isAccepted,
+                "border-red-500 rejected": isRejected,
+                "border-gray-300": !isAccepted && !isRejected,
```
zeroliu: tokens like green-500, red-500, gray-300 are not valid tokens in our design system. We should use the ones configured in the design system.

src/styles/tailwind.css
```diff
@@ -872,3 +896,125 @@ If your plugin does not need CSS, delete this file.
...
+
+/* Add styles for the summary section */
+.apply-view-summary {
+  background-color: var(--background-secondary-alt);
+  padding: 8px 12px;
+  border-bottom: 1px solid var(--background-modifier-border);
+  margin-bottom: 12px;
+}
+
+.apply-view-summary .text-green-600 {
```
zeroliu: We do not allow adding classnames like this. This breaks how we work with tailwindcss. There are existing tokens that work for the color that you need. I highly recommend take a close look at tailwind.config.js and follow tailwindcss styles.

zeroliu (changes requested):

package.json
wenzhengjiang: Done.

src/LLMProviders/brevilabsClient.ts
wenzhengjiang: Done.

src/plusUtils.ts
wenzhengjiang: The main benefit is faster deployment. At least in the pre-release stage, I expect there will be a lot of prompt-tuning and deploying changes from the backend is a lot easier than asking users to upgrade their code.

After the official release, I don't have strong opinion whether putting the prompt in the backend or client side.

src/styles/tailwind.css
wenzhengjiang: Got it. Sounds good! Removed all custom CSS class below.

src/styles/tailwind.css
wenzhengjiang: Got it. Migrates the tokens to tailwindcss.

src/components/chat-components/ChatSingleMessage.tsx
wenzhengjiang: No specific reason, simply never did web frontend before :). Updated the code by creating a custom React component. Hope that makes more sense to you.

src/components/ApplyView.tsx
wenzhengjiang: I see. Done.

wenzhengjiang:
@logancyang @zeroliu I think I have addressed all your concerns. I'm pretty new to web frontend, so I could be still missing something...

Currently I'm using the default variant for accept buttons and secondary variant for reject buttons. They look like below:

Screenshot 2025-03-04 at 22 39 22

Open to making the green and red changes if you think they’re worth it.

src/components/ApplyView.tsx
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  getDisplayText(): string {
+    return "Apply Changes";
+  }
+
+  async setState(state: ApplyViewState) {
+    this.state = state;
+    this.render();
+  }
```
zeroliu: is this function used somewhere?

src/components/ApplyView.tsx
```diff
@@ -0,0 +1,297 @@
...
+      this.root = null;
+    }
+  }
+
+  private render() {
+    if (!this.state) return;
+
+    const container = this.containerEl.children[1];
+    container.empty();
+    container.addClass("apply-view-container");
```
zeroliu: do we need this class?

src/components/ApplyView.tsx
```diff
@@ -0,0 +1,297 @@
...
+    if (this.root) {
+      this.root.unmount();
+      this.root = null;
+    }
+  }
+
+  private render() {
+    if (!this.state) return;
+
+    const container = this.containerEl.children[1];
```
zeroliu: why `this.containerEl.children[1];`? Is this deterministic? what other children does the container element have?

src/components/ApplyView.tsx
```diff
@@ -0,0 +1,297 @@
...
+    container.empty();
+    container.addClass("apply-view-container");
+
+    const rootEl = container.createDiv();
+    if (!this.root) {
+      this.root = createRoot(rootEl);
+    }
+
+    this.root.render(
+      <ApplyViewRoot app={this.app} state={this.state} close={() => this.leaf.detach()} />
```
zeroliu: I don't see how this.state is set. `setState` is not called in this PR and it's not coming from constructor.

src/components/ApplyView.tsx
```diff
@@ -0,0 +1,297 @@
...
+    });
+
+    if (currentBlock.length > 0) {
+      blocks.push(currentBlock);
+    }
+
+    setChangeBlocks(blocks);
+  }, [diff]);
+
+  // Add defensive check for state after hooks
```
zeroliu: How would the UI enter this state?

src/components/ApplyView.tsx
```diff
@@ -0,0 +1,297 @@
...
+      const block = changeBlocks[blockIndex];
+
+      // Find the indices of the changes in this block
+      block.forEach((blockChange) => {
+        const index = newDiff.findIndex((change) => change === blockChange);
+        if (index !== -1) {
+          newDiff[index] = {
+            ...newDiff[index],
+            accepted: true,
+            rejected: false,
```
zeroliu: why do we need two flags to represent this value? is there a case where accepted and rejected can be same values?

src/components/ApplyView.tsx
```diff
@@ -0,0 +1,297 @@
...
+                    <XIcon className="h-4 w-4" />
+                    Reject
+                  </Button>
+                  <Button size="sm" onClick={() => acceptBlock(blockIndex)}>
+                    <Check className="h-4 w-4" />
+                    Accept
+                  </Button>
+                </div>
+              )}
+            </div>
```
zeroliu: can we apply the change directly in this view when a user accept or reject a block? It's not very clear after a user accepts and rejects the change and what's going to happen

src/plusUtils.ts
```diff
@@ -80,3 +93,28 @@ export function turnOffPlus(): void {
     new CopilotPlusExpiredModal(app).open();
   }
 }
+
+export async function getComposerSystemPrompt(): Promise<string> {
```
zeroliu: I suggest move anything composer related to its own file. `plusUtil` should only take care of plus user status checks

src/components/chat-components/ChatSingleMessage.tsx
```diff
@@ -79,6 +85,100 @@ const ChatSingleMessage: React.FC<ChatSingleMessageProps> = ({
     });
   };
 
+  const handleApplyCode = useCallback(
```
zeroliu: I would also suggest create a custom hook and move this chunk of code to a separate module under composer. Making composer code more isolated from the core chat.

src/components/chat-components/ChatSingleMessage.tsx
```diff
@@ -79,6 +85,100 @@ const ChatSingleMessage: React.FC<ChatSingleMessageProps> = ({
...
+          // Open the Apply View in a new leaf with the processed content
+          const leaf = app.workspace.getLeaf(true);
+          await leaf.setViewState({
+            type: "obsidian-copilot-apply-view",
+            active: true,
+            state: {
+              file: file,
+              originalContent: originalContent,
+              newContent: newContent,
+              path: path,
```
zeroliu:
```suggestion
              file,
              originalContent,
              newContent,
              path,
```

src/components/chat-components/ChatSingleMessage.tsx
```diff
@@ -203,6 +303,32 @@ const ChatSingleMessage: React.FC<ChatSingleMessageProps> = ({
         "", // Empty string for sourcePath as we don't have a specific source file
         componentRef.current
       );
+
+      // Process code blocks after rendering
+      const codeBlocks = contentRef.current.querySelectorAll("pre");
```
zeroliu, line 215: is there any reason why we need to escape from react component tree here?

src/styles/tailwind.css
```diff
@@ -432,26 +432,50 @@ If your plugin does not need CSS, delete this file.
 .message-content pre .copy-code-button {
```
zeroliu: can you please help me understand why we need the changes in this file still?

zeroliu (commented):
looks much better. Left more comment as I'm digging into more implementation details

zeroliu:
> @logancyang @zeroliu I think I have addressed all your concerns. I'm pretty new to web frontend, so I could be still missing something...
> 
> Currently I'm using the default variant for accept/apply buttons and secondary variant for reject buttons. They look like below:
> 
> Screenshot 2025-03-04 at 22 39 22
> Open to making the green and red changes if you think they’re worth it.

We will probably need new tokens for this since it's a new type of UI. I can help you create them once we get the main body of the code ready 🥂

src/components/ApplyView.tsx
wenzhengjiang: Ah, you are right. We don't need this anymore.

src/components/ApplyView.tsx
wenzhengjiang: Didn't find any official documentation, but it seems like a standard pattern to access the second child and the content view while the first child as the header view. See https://github.com/search?q=this.containerEl.children%5B1%5D&type=code

src/components/ApplyView.tsx
wenzhengjiang: It's set in https://github.com/logancyang/obsidian-copilot/pull/1315/files#diff-eee12cc355c7066cd49d0c5b30e8b3ec2ad7c87acc76702adae3da7a9cd6b54dR150 via setViewState of the WorkspaceLeaf object. I think internally WorkspaceLeaf called the setState to pass the state object. 

Again, not official document, but found something related in https://liamca.in/Obsidian/API+FAQ/views/persisting+your+view+state.

wenzhengjiang:
> looks much better. Left more comment as I'm digging into more implementation details

Thanks for the detailed review! Latest updates:
1. Isolate composer code to their own files
2. Update the UI to use checkbox to select change blocks.
3. Revert all changes on classical css classes.


Screenshot 2025-03-07 at 23 18 45

zeroliu (approved):

src/components/composer/ApplyView.tsx
```diff
@@ -0,0 +1,437 @@
...
+    if (currentBlock.length > 0) {
+      blocks.push(currentBlock);
+    }
+
+    return blocks;
+  }, [diff]);
+
+  // Add refs to track change blocks
+  const blockRefs = useRef<(HTMLDivElement | null)[]>([]);
+  console.log(changeBlocks);
```
logancyang, line 134: Should this be removed

logancyang:
Another observation is that I always get a single empty line diff at the end every time I hit Apply, might worth removing it programmatically? @wenzhengjiang @zeroliu 

SCR-20250311-kdoe

logancyang (approved):
LGTM!! This is a huge milestone, congrats everyone! 🚀🚀🚀

wenzhengjiang:
Plan for improving the current implementation:
1. Trigger composer with the `@composer` tool
2. The `@composer` tool:
  * Input: user query, chat history, original content of target notes
  * Output: diff in md block, full new content.
  * Can call the current chat model, but should be in a new chat session.
  * diff md only needs to contain the added lines. (for now)
3. diff md block is added to chat context to be displayed to users.
4. The full new content is added to a shared state to be fetched in apply view.

@logancyang @zeroliu @Emt-lin

PR #1316, opened 2025-03-01
Enhance custom command

zeroliu:
Major update to the inline edit command

1. All built-in inline edit commands are now surfaced to settings, allowing customization.
2. New custom commands can be created to fit specific user use cases.
3. All inline edit commands can have customizable names, prompts, options to show in the context menu, and choices of which LLM model to use.

Note that the legacy translate and change tone commands are removed because we no longer support subCommandEvent. However, users can create their own version of them as shown in the gif above.


logancyang (commented):
Looks like some commands like `translation` and `tone` are removed? Translation is a big one and has a 2-level structure, can it be supported in this new scheme?

zeroliu:
I left a note in the PR description about this decision.

I don't think we need to support two level command any more. It will add complexity to the settings, making it harder to use. For translation, users can create their own new custom command easily with the new functionality as shown in the attached gif. This is easier to use than going through a list of languages to trigger translation.

logancyang:
We should address the user question "where did the translation command go" since they were mentioned as useful among all commands.

We could

1. create an example translation command
2. have some words about how they can create one in the new command setting tab

Any other ideas for a better transition?

zeroliu:
> We should address the user question "where did the translation command go" since they were mentioned as useful among all commands.
> 
> We could
> 
> 1. create an example translation command (this is useful because a simple "translate to language X" usually returns a bunch of fluff. We can set an example to save them some prompt engineering)
> 2. have some words about how they can create one in the new command setting tab
> 
> Any other ideas for a better transition?

Agree with this direction. Creating a simplified translation command by reusing the existing prompt can be very helpful.
About sharing the transition with users, are you thinking of including it in the release note / announcement or in the app?

zeroliu:
Translation example added @logancyang

src/commands/constants.ts
```diff
@@ -0,0 +1,137 @@
...
+export const DEFAULT_INLINE_EDIT_COMMANDS: InlineEditCommandSettings[] = [
+  {
+    name: "Fix grammar and spelling",
+    prompt:
+      `<instruction>Fix the grammar and spelling of the text below. Preserve all formatting, line breaks, and special characters. Do not add or remove any content. Return only the corrected text.</instruction>\n\n` +
+      `<text>${SELECTED_TEXT_PLACEHOLDER}</text>`,
+    showInContextMenu: true,
+  },
+  {
+    name: "Translate to Chinese",
```
logancyang, line 15: This is great. However, any default settings change must make sure it gets reflected without a setting reset. Right now, the prev setting didn't have this entry in the `inlineEditCommands`, so any change here won't show up without a setting reset.

SCR-20250303-kvum

Should we add a refresh button in the Commands setting tab to perform the necessary merging? It could be reused in the Model tab too.

logancyang:
> > We should address the user question "where did the translation command go" since they were mentioned as useful among all commands.
> > We could
> > 
> > 1. create an example translation command (this is useful because a simple "translate to language X" usually returns a bunch of fluff. We can set an example to save them some prompt engineering)
> > 2. have some words about how they can create one in the new command setting tab
> > 
> > Any other ideas for a better transition?
> 
> Agree with this direction. Creating a simplified translation command by reusing the existing prompt can be very helpful. About sharing the transition with users, are you thinking of including it in the release note / announcement or in the app?

Release note will have the announcement but I'm afraid most people don't read it. Any words in the settings tab like "Builtin commands serve as an example, create your own commands here" or something like that.

src/commands/constants.ts
zeroliu: This is being addressed. No settings reset is needed. If you test it today with your current settings data without inline edit comments, you will see the default commands display correctly. This is because when we read inline edit commands, we always use the get inline edit commands function, which defaults to the default commands. This allows us to show default built-in commands if the value doesn't exist. I have also set the type of inline edit commands to be possibley undefined, ensuring that throughout the code, when referring to inline edit commands from settings, we address the default when the value is undefined.

logancyang:
> > Release note will have the announcement but I'm afraid most people don't read it. Any words in the settings tab like "Builtin commands serve as an example, create your own commands here" or something like that.
> 
> I believe this can be done. One thing to avoid is displaying this instruction all the time, even after a user has already created custom commands in the past. Additionally, it's important not to store too many variables in the settings to prevent cluttering the already complex settings data.json file. A middle ground could be tracking whether a user has created any custom commands by comparing what's in data.json with the default, and if they have, stop displaying this instruction. I also want to avoid overwhelming the UI with instructions and aim for a design that's intuitive on its own. If we need to explain the UI, it indicates that the design may not be optimal.

I agree with this philosophy. In this case it's just a one-liner description for the purpose of the "commands" tab, I think it's fine to have it there always. Other sections have descriptions too.

zeroliu:
@logancyang updated

zeroliu:
> > > Release note will have the announcement but I'm afraid most people don't read it. Any words in the settings tab like "Builtin commands serve as an example, create your own commands here" or something like that.
> > 
> > 
> > I believe this can be done. One thing to avoid is displaying this instruction all the time, even after a user has already created custom commands in the past. Additionally, it's important not to store too many variables in the settings to prevent cluttering the already complex settings data.json file. A middle ground could be tracking whether a user has created any custom commands by comparing what's in data.json with the default, and if they have, stop displaying this instruction. I also want to avoid overwhelming the UI with instructions and aim for a design that's intuitive on its own. If we need to explain the UI, it indicates that the design may not be optimal.
> 
> I agree with this philosophy. In this case it's just a one-liner description for the purpose of the "commands" tab, I think it's fine to have it there always. Other sections have descriptions too.

Oh, I added a banner before I saw your message. Can you take a look and let me know if you prefer a description instead.?

zeroliu:
I think adding a description to explain what "command" is and how to trigger that in the settings will help. The message about using the built-in command as an example to create your own is not timeless and may not apply after the user creates their own set of commands.

logancyang:
The description can simply be "Edit builtin commands or create your own". The banner and a state flag for it is a bit much imo.

src/commands/constants.ts
logancyang: Sorry I don't follow. When the data.json is already filled with a prev default (like in my case), any code update with new defaults in future versions doesn't get reflected, right? The same thing happens with the default model updates. There's no migration from old defaults to new defaults.

zeroliu:
Screenshot 2025-03-03 at 6 46 23 PM

This is the description I am thinking. 

"Edit builtin commands or create your own" honestly feels redundant. We don't have "Edit model or create your own" on the model settings page.

zeroliu:
> The banner and a state flag for it is a bit much IMO. 

I didn't create a new flag. I just check whether the inlineEditCommand === DEFAULT. Once anything changes, it will disappear.

src/commands/constants.ts
zeroliu: I see what you are saying. I was talking about when user didn't have any inlineEditCommands and there are code to handle it without resetting settings. A few questions:

1. How did we handle chat model default value update today?
2. In what scenario will we introduce new built-in command and want it to show up in users' panel, especially after they already update it to match their use case. Imagine a scenario where I delete all the built-in commands and added 3 of my own. Now a new copilot version introduces a new builtin command. IMO, the expected behavior will be that it doesn't show up for me because it shouldn't interrupt the custom commands that I set up. This is different from chat model where we have defaults that users cannot remove. WDYT?

zeroliu: That's a good callout though. The way that it is implemented today is that we will give the control of the command prompt entirely to the users. Future prompt optimization on the commands will only reflect on the new users or users who reset the settings. I personally think this is acceptable but if you think of any edge cases that can cause issues, please let me know. This is a one-way door so we should be careful.

logancyang: 1. We used to have a merge util that merges any existing models from data.json with the builtin models of the latest version. That logic got lost during setting v2 migration and last time when I added those o-series to builtin, the users couldn't see them without a setting reset. Some were not happy.
2. I rarely updated the builtin commands. Now that we have user commands, so probably never 😂

I think we should have a generic util for migrating an old Copilot setting data.json to a new one, and assign it to a "refresh settings" button or something. But it's out-of-scope for this work. I think this looks good now and we can defer that setting refresh stuff to a future PR.

logancyang (approved):
LGTM! Awesome stuff, will merge in a bit.

Issue #1336, opened 2025-03-06
Azure OpenAI Embedding models fail with 401


roqman97:
- [x] Disable all other plugins besides Copilot **(required)**
- [x] Screenshot of note + Copilot chat pane + dev console added **(required)**

Copilot version: 2.8.7

(Bug report without the above will be closed)

**Describe how to reproduce**
A clear and concise description of what the bug is. Clear steps to reproduce the behavior

1)  Add an Azure OpenAI Configuration option (with key and base URL) to the copilot.
2) Use those settings to add a custom embedding model, using Azure OpenAI as the provider
3) Review 401 unauthorized error

**Expected behavior**

Using the same settings to add an embedding model and a chat model should work both times.

**Screenshots**
Add screenshots to help explain your problem. Please turn on debug mode in Copilot settings, turn off other plugins to leave only Copilot dev messages as necessary.

Model verifies with GPT-4o-mini
![Image](https://github.com/user-attachments/assets/07c7f1d0-ddd5-442d-9fe7-8629cc6a0c3b)

Model gets 401 when trying to use embedding model with identical settings
![Image](https://github.com/user-attachments/assets/49feb121-d635-4d68-b4fe-fcfd49448c40)

**Additional context**

The model was working in previous versions of the copilot.

PR #1337, opened 2025-03-06
Update embeddingManager.ts to use the updated AzureOpenAIEmbeddings c…

roqman97:
Updated the embeddingManager to use the native AzureOpenAIEmbeddings class when using the Azure OpenAI provider.

Documentation for this integration is available here: https://js.langchain.com/docs/integrations/text_embedding/azure_openai/#integration-details

This should not require any change for how users interact with adding Azure OpenAI models.

This should resolve issue #1336

### DIFF
--- a/src/LLMProviders/embeddingManager.ts
+++ b/src/LLMProviders/embeddingManager.ts
@@ -10,6 +10,7 @@ import { Embeddings } from "@langchain/core/embeddings";
 import { GoogleGenerativeAIEmbeddings } from "@langchain/google-genai";
 import { OllamaEmbeddings } from "@langchain/ollama";
 import { OpenAIEmbeddings } from "@langchain/openai";
+import { AzureOpenAIEmbeddings } from "@langchain/openai";
 import { Notice } from "obsidian";
 import { CustomJinaEmbeddings } from "./CustomJinaEmbeddings";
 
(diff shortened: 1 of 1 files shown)


logancyang:
Thanks for the PR. Have you tested this change locally?

roqman97:
> Thanks for the PR. Have you tested this change locally?

Yes, I ran `npm build`, And copied the new main.js into my obsidian plug-in (overwriting the existing one) and it resolved my embedding issues. I know that's a bit of a janky local test, but it was the fastest way I could think to try it out.

infinitelyloopy-bt:
> > Thanks for the PR. Have you tested this change locally?
> 
> Yes, I ran `npm build`, And copied the new main.js into my obsidian plug-in (overwriting the existing one) and it resolved my embedding issues. I know that's a bit of a janky local test, but it was the fastest way I could think to try it out.

Hi @roqman97 I am confused here. Did you mean to download the plugin and install it into your obsidian folder manually? If yes, where and how and what npm build command did you enter?

I have updated the embeddingManager.ts locally, but then I am confused as to what the next step is or should be.

Do I delete the copilot plugin that was install via Obsidian? and then install this local version? If yes, how?

infinitelyloopy-bt:
> Sorry for the confusion.
> 
> To clarify, i was doing the coding outside of my plug-ins folder (I wanted to test this in multiple vaults). So, after updating the .ts file in the repo, I ran `npm install` and then `npm build` on the updated repo. Then, i took the generated main.js and replaced the existing one in the copilot folder inside my vault's plugins.

Great! I tried that and it worked but the options page of the plugin seemed like the CSS or layout was complete formatted incorrectly.

I'll just wait for the developer to correct the bug.

logancyang:
How did you guys add the azure custom model? I can't make it work. Azure is confusing as hell right now, there used to be just Azure OpenAI, now I can only find AI Foundry and can't find any CURL example from their page. 

Can you share some screenshots how you made it work? Mask sensitive parts as needed. @roqman97 @infinitelyloopy-bt

logancyang:
Ok i got it to work. Still, the Azure AI Foundry pages are quite the maze to navigate.

roqman97:
Sorry I never replied! I was out of the country for a short vacation. Glad you got it sorted - and I agree, Foundry can be a bit confusing.

PR #1354, opened 2025-03-11
Label and mask licence key when debug logging is enabled

johannrichard:
The plus licence key is printed verbatim when logging is enabled. In the interest of privacy and security, I would suggest masking it slightly, to avoid that people accidentally disclosing it e.g. when posting console logs. (Apologies for the previous, now closed PR, I messed up w/ my branches).


logancyang:
Thanks for the PR. Actually this line should be removed entirely.

PR #1359, opened 2025-03-14
feat: Add project-based UI.

Emt-lin:


src/LLMProviders/chatModelManager.ts
```diff
@@ -103,6 +103,7 @@ export default class ChatModelManager {
           fetch: customModel.enableCors ? safeFetch : undefined,
           organization: await getDecryptedKey(customModel.openAIOrgId || settings.openAIOrgId),
         },
+        topP: customModel.TopP,
```
logancyang: Why are we adding TopP for them individually here? Also, temperature and TopP shouldn't be present at the same time. We should add Frequency Penalty instead, but to the base config, not individually.

logancyang:
The init problem still exists
SCR-20250316-otzt

logancyang:
Since we will rely on context caching, it's essential to not allow some models in Project mode or the user will face huge cost. We should allow only models with the Google Gemini provider in the early iterations where we specifically implement context caching for them on the client side.

Just putting it here, not really a requirement for this PR but it's relevant for "Projects" mode overall. 

BTW we should name it "Plus Projects (alpha)" mode.

logancyang:
Thanks for the PR!

Here are some observations

1. The text inputs in Project don't allow space?
2. Prompt should be "Project System Prompt" and it should be optional.
3. The models should be hardcoded to `gemini-1.5-flash-001` and `gemini-1.5-pro-001` in this iteration since they are the only ones with 1M context and context caching.
4. When no inclusion/exclusion but only youtube URL, the Project "save" button is disabled, but it should not be.
5. When there's an inclusion folder and a youtube URL, the final prompt only has the youtube transcript but not the included notes. (I like the spinner "project context processing"!)

I'm going to attempt to fix some of these myself and aim for a pre-release for tmr.

logancyang:
- [ ] The text inputs in Project don't allow space?
- [ ] Prompt should be "Project System Prompt" and it should be optional.
- [ ] After updating the project system prompt, it is not actually updated during chat.
- [ ] The instruction for <ProjectContext> should be outside of the tag.
- [ ] The models should be hardcoded to gemini-1.5-flash-001 and gemini-1.5-pro-001 in this iteration since they are the only ones with 1M context and context caching.
- [ ] When no inclusion/exclusion but only youtube URL, the Project "save" button is disabled, but it should not be.
- [ ] When there's an inclusion folder and a youtube URL, the final prompt only has the youtube transcript but not the included notes. (I like the spinner "project context processing"!)
- [ ] Project context cache is not working

logancyang:
This PR has been merged into https://github.com/logancyang/obsidian-copilot/tree/2.9.0-preview

This PR is now for **review only**. Further changes should be made to a **NEW PR against that branch**.

PR #1380, opened 2025-03-20
Add plus-exclusive and believer-exclusive checks in embeddingManager

logancyang:
This is to avoid unnecessary 403 calls to the backend.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,31 @@
+# CLAUDE.md - Copilot for Obsidian Plugin
+
+## Development Commands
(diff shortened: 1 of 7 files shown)


logancyang:
@zeroliu We have UI logic in `turnOffPlus` that is imported to brevilabsClient, which is in turn used at various places. This created a dependency on this UI logic and broke the tests. How should we best break this out?

logancyang:
Using brevilabsClient directly in embeddingManager to avoid plusUtils's circular dependency for now.

PR #1385, opened 2025-03-22
Fix Azure OpenAI chat model baseURL construction logic.

doin4:
Hi, I noticed that when using Azure OpenAI as the chat model provider, the generated URL is incorrect. I modified the code to use the user-provided endpoint URL, and it worked correctly for me.

However, there is still an issue when using Azure OpenAI as the embedding model provider. Unfortunately, I am not familiar enough with this part of the script to rewrite the URL logic. I can, however, provide some error details: the script generates a default URL in the format `https://{model_name}xx.openai.xx`, which is not the correct format for the Azure OpenAI API. I strongly recommend using the user-provided endpoint URL instead of constructing the URL programmatically based on rules.

### DIFF
--- a/src/LLMProviders/chatModelManager.ts
+++ b/src/LLMProviders/chatModelManager.ts
@@ -120,7 +120,9 @@ export default class ChatModelManager {
           customModel.azureOpenAIApiDeploymentName || settings.azureOpenAIApiDeploymentName,
         openAIApiKey: await getDecryptedKey(customModel.apiKey || settings.azureOpenAIApiKey),
         configuration: {
-          baseURL: customModel.baseUrl,
+          baseURL:
+            customModel.baseUrl ||
+            `https://${customModel.azureOpenAIApiInstanceName || settings.azureOpenAIApiInstanceName}.openai.azure.com/openai/deployments/${customModel.azureOpenAIApiDeploymentName || settings.azureOpenAIApiDeploymentName}`,
           defaultQuery: {
             "api-version": customModel.azureOpenAIApiVersion || settings.azureOpenAIApiVersion,
           },
--- a/src/LLMProviders/chatModelManager.ts
+++ b/src/LLMProviders/chatModelManager.ts
@@ -120,7 +120,7 @@ export default class ChatModelManager {
           customModel.azureOpenAIApiDeploymentName || settings.azureOpenAIApiDeploymentName,
         openAIApiKey: await getDecryptedKey(customModel.apiKey || settings.azureOpenAIApiKey),
         configuration: {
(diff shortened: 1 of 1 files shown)


logancyang:
Your change discarded the existing instance name and deployment name params entirely. Since Azure now has many different structures available, it's better to have custom model URL as an override rather than the only way.

logancyang (changes requested):
`customModel.baseUrl` should be an override instead of the only way to set it.

doin4:
I still can't find a right way to change logic for Azure embedding models. I tried to add a `baseURL` in config, but it seems not work.

doin4:
Hope you can solve this embedding URL problem.

doin4:
The problem is that I met the same URL issue when trying to get Azure OpenAI embedding models. The generated URL is not the end point that I provided, thus I cannot verify.

I tried to fix the new bug the way I did for chat model, but not working.

tacticsiege:
Azure OpenAI model deployments follow the url format:
`POST https://YOUR_RESOURCE_NAME.openai.azure.com/openai/deployments/YOUR_DEPLOYMENT_NAME/chat/completions?api-version=2024-06-01`

https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#rest-api-versioning

doin4:
@tacticsiege Sorry for the late response, and THANKS for your information!

src/LLMProviders/chatModelManager.ts
```diff
@@ -120,7 +120,8 @@ export default class ChatModelManager {
           customModel.azureOpenAIApiDeploymentName || settings.azureOpenAIApiDeploymentName,
         openAIApiKey: await getDecryptedKey(customModel.apiKey || settings.azureOpenAIApiKey),
         configuration: {
-          baseURL: `https://${customModel.azureOpenAIApiInstanceName || settings.azureOpenAIApiInstanceName}.openai.azure.com/openai/deployments/${customModel.azureOpenAIApiDeploymentName || settings.azureOpenAIApiDeploymentName}`,
+          baseURL:
+            customModel.baseUrl || `https://${customModel.azureOpenAIApiInstanceName || settings.azureOpenAIApiInstanceName}.openai.azure.com/openai/deployments/${customModel.azureOpenAIApiDeploymentName || settings.azureOpenAIApiDeploymentName}/chat/completions`,
```
logancyang: `/chat/completions` shouldn't be in the baseURL, right?

doin4: I have tested both ways, and they point to the same address. Since it is the base URL, I agree with you and removed the `/chat/completions`.

logancyang (approved):

PR #1399, opened 2025-03-27
Refactor Composer: Implement @composer tool.

wenzhengjiang:
#1315 
NOTE: `@composer` is not implemented as a typical copilot tool as it needs to access the output of other tools. 
Flow:
1. In chainRunner, check if user message contains `@composer`.
2. If true, get all the context and then send the message with context to `composer`.
3. `composer` calls the current chat model and gets the new note content in a certain JSON format.
4. `composer` creates the diff against the original content and store it in a global map.
5. Lastly `composer` returns a user-friendly diff to chainRunner and rewrite the chat context with only the diff.
6. When apply button is clicked, ApplyView show the diff view stored in step 4.

Test cases for diff generation:
* Generate changes containing ADD, REMOVE
* Generate changes for multiple notes
* Request with missing target note
* Multi-turn edits
* Combine with `@youtube` and URL context

Test cases for apply view:
* `accept` and `reject` buttons at the bottom
* `accept` and `reject` button for individual blocks.


logancyang:
Observation: when i ask it to create a new note it says fine, but apply fails
SCR-20250402-ogio

```
Error creating file: Error: ENOENT: no such file or directory, open '/Users/.../logan-dev-vault/Summaries/AI Safety Alignment Landscape.md'
    at async open (node:internal/original-fs/promises:636:25)
    at async Object.writeFile (node:internal/original-fs/promises:1216:14)
```

src/LLMProviders/chainRunner.ts
```diff
@@ -632,13 +627,29 @@ class CopilotPlusChainRunner extends BaseChainRunner {
...
         }
 
-        fullAIResponse = await this.streamMultimodalResponse(
-          enhancedUserMessage,
-          userMessage,
-          abortController,
-          updateCurrentAiMessage,
-          debug
-        );
+        if (messageForAnalysis.includes("@composer")) {
```
logancyang, line 629: `@composer` is in the `else` block of local search right now, so we can't do things like `@vault @composer ...`, right? I think this is ok now, but we might need to refactor and make this more modular, and revisit the relationship between these tools at some point.

Since `@composer @youtube ...` works, users may wonder why other combinations don't work.

wenzhengjiang: Both `@vault` and `@web` take the whole user message as input so currently there is no way to make them work with `@composer`

I can refactor the code here a bit to make it cleaner, and we can create a separate ticket for redesigning above two commands.

wenzhengjiang:
> Observation: when i ask it to create a new note it says fine, but apply fails SCR-20250402-ogio
> 
> ```
> Error creating file: Error: ENOENT: no such file or directory, open '/Users/.../logan-dev-vault/Summaries/AI Safety Alignment Landscape.md'
>     at async open (node:internal/original-fs/promises:636:25)
>     at async Object.writeFile (node:internal/original-fs/promises:1216:14)
> ```
> 
> The issue seems to be any generated note title with `/` not being handled correclty.

Thanks catching this! It should be fixed now.

PR #1414, opened 2025-03-31
Update broca requests

zeroliu:


src/LLMProviders/brevilabsClient.ts
```diff
@@ -104,9 +104,14 @@ export class BrevilabsClient {
     endpoint: string,
     body: any,
     method = "POST",
-    excludeAuthHeader = false
+    excludeAuthHeader = false,
+    skipLicenseCheck = false
   ): Promise<{ data: T | null; error?: Error }> {
-    this.checkLicenseKey();
+    if (!skipLicenseCheck) {
```
logancyang, line 110: Can this be more explicit without this flag, like

```
if (endpoint !== '/license') {
    this.checkLicenseKey();
}
```

logancyang, line 110: Can this be more explicit and avoid adding this flag?

```
if (endpoint !== '/license') {
    this.checkLicenseKey();
}
```

src/constants.ts
```diff
@@ -469,6 +470,7 @@ export const COMMAND_NAMES: Record<CommandId, string> = {
 export type CommandId = (typeof COMMAND_IDS)[keyof typeof COMMAND_IDS];
 
 export const DEFAULT_SETTINGS: CopilotSettings = {
+  userId: uuidv4(),
```
logancyang, line 473: So this essentially is an ID per-vault iiuc?

logancyang (approved):
Left some minor comments and question. LGTM.

src/LLMProviders/brevilabsClient.ts
zeroliu: My personal preference is to let the license check method pass in the flag instead of verifying the license route in the shared method. In case we change license check route in the future, it won't impact this logic.

src/constants.ts
zeroliu: that's right!

PR #1416, opened 2025-03-31
feat: Supports opening and entering the license key via a browser.

Emt-lin:


src/main.ts
```diff
@@ -328,4 +338,45 @@ export default class CopilotPlugin extends Plugin {
...
     }));
   }
+
+  registerLicenseKeyHandler(params: ObsidianProtocolData) {
+    const encodedLicenseKey = params["licenseKey"];
+
+    if (encodedLicenseKey) {
+      try {
+        const licenseKey = decodeURIComponent(encodedLicenseKey);
+        logInfo(`==== licenseKey ====:`, licenseKey);
```
logancyang, line 348: We should not log license key.

src/typings.d.ts
```diff
@@ -0,0 +1,41 @@
...
+          api: any;
+        };
+        ["obsidian-hover-editor"]?: Plugin & {
+          activePopovers: (HoverPopover & {
+            toggleMinimized(): void;
+            togglePin(value?: boolean): void;
+          })[];
+          spawnPopover(initiatingEl?: HTMLElement, onShowCallback?: () => unknown): WorkspaceLeaf;
+        };
+        ["obsidian-tts"]?: Plugin & {
```
logancyang, line 31: What's this and why is this needed?

logancyang:
I have concerns about this approach because apparently this URL action with the license key can be intercepted.

```
- URL Parameters in Browser History
When a user clicks a link with the license key, it gets stored in their browser history
Anyone with access to the browser history can see the license key
Example URL: obsidian://copilot?licenseKey=abc123xyz

- Network Traffic
If the user is on an unsecured network (public WiFi, etc.)
If the URL is accessed over HTTP instead of HTTPS
Network administrators or malicious actors could intercept the request

- Referrer Headers
If the user clicks the link from another website, the full URL (including license key) might be sent in the referrer header

- Browser Extensions
Malicious browser extensions could read the URL parameters
Extensions with permission to read browser history could access it
```

Unless it's a one-time access token, this seems an unnecessary risk. 

wdyt? @Emt-lin @zeroliu @wenzhengjiang

zeroliu:
That's a valid concern. We haven't received feedback from users asking about how to apply the license key. It's probably better to take some precautions here.

logancyang:
Putting this one on hold until prioritized.

PR #1420, opened 2025-04-01
some project-based feature bug fix 

Emt-lin:
## Bug
1. the old project context cache file isn't deleted when cache updated.
2. the context cache doesn't be updated when the source file updated.
3. Since All those API(url、pdf api) calls are slow and expensive, so need to refactor project caching functionality.

## Fixed

1. Fixed issues with item 1 and item 2.
2. Refactor project caching functionality.
  - 1.new cache structure.
 
     ```ts
       export interface ContextCache {
          markdownContext: string;
          webContexts: Record<string, string>; // URL -> context
          youtubeContexts: Record<string, string>; // URL -> context
          timestamp: number;
          markdownNeedsReload: boolean;
       }
    ```

 - 2.When the project configuration changed, markdownContext、webContexts/youtubeContexts will be processed differently
   - markdownContext: all markdownContext cache will be deleted.
   - webContexts/youtubeContexts: Remove context for URLs that no longer exist.
 - 3.Load project context steps
   - markdownContext: Only process if `markdownNeedsReload` flag is true or there is no existing content
   - webContexts/youtubeContexts: For each url, fetch from cache if available; otherwise, make API call load context.

### DIFF
--- a/src/LLMProviders/projectManager.ts
+++ b/src/LLMProviders/projectManager.ts
@@ -39,7 +39,7 @@ export default class ProjectManager {
     this.plugin = plugin;
     this.currentProjectId = null;
     this.chainMangerInstance = new ChainManager(app, vectorStoreManager);
-    this.projectContextCache = ProjectContextCache.getInstance();
+    this.projectContextCache = ProjectContextCache.getInstance(app);
     this.chatMessageCache = new Map();
 
     // Set up subscriptions
@@ -251,14 +251,6 @@ ${contextParts.join("\n\n")}
     return this.projectContextCache.getSync(project);
   }
 
-  public async clearContextCache(projectId: string): Promise<void> {
-    const project = getSettings().projectList.find((p) => p.id === projectId);
-    if (project) {
-      await this.projectContextCache.clearForProject(project);
-      logInfo(`Context cache cleared for project: ${projectId}`);
-    }
-  }
-
   private async processMarkdownContext(inclusions?: string, exclusions?: string): Promise<string> {
     if (!inclusions && !exclusions) {
       return "";
@@ -335,5 +327,9 @@ ${content}`;
 
     const results = await Promise.all(processPromises);
     return results.join("");
+  }
+
+  public onunload(): void {
+    this.projectContextCache.cleanup();
   }
 }
--- a/src/LLMProviders/projectManager.ts
+++ b/src/LLMProviders/projectManager.ts
@@ -7,7 +7,7 @@ import {
   subscribeToModelKeyChange,
   subscribeToProjectChange,
 } from "@/aiParams";
-import { ProjectContextCache } from "@/cache/projectContextCache";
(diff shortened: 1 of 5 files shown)


logancyang:
You should create the PR against 2.9.0-preview, not master. Check https://github.com/logancyang/obsidian-copilot/pull/1399

Emt-lin:
@logancyang wait review

logancyang:
Could you list what's been done in the PR in the description?

src/LLMProviders/projectManager.ts
```diff
@@ -39,7 +39,7 @@ export default class ProjectManager {
     this.plugin = plugin;
     this.currentProjectId = null;
     this.chainMangerInstance = new ChainManager(app, vectorStoreManager);
-    this.projectContextCache = ProjectContextCache.getInstance();
+    this.projectContextCache = ProjectContextCache.getInstance(app);
```
logancyang: Why do we need this `app` here? It's a global in this codebase everywhere.

src/cache/projectContextCache.ts
```diff
@@ -1,38 +1,104 @@
...
+  private debounceTimer: number | null = null;
 
-  private constructor() {}
+  private constructor(private app: App) {
+    this.vault = app.vault;
+    this.initializeEventListeners();
+  }
 
-  static getInstance(): ProjectContextCache {
+  static getInstance(app: App): ProjectContextCache {
```
logancyang: Ditto, shouldn't need this `app` in the constructor.

src/cache/projectContextCache.ts
```diff
@@ -1,38 +1,104 @@
...
+      // check if the project needs to clear cache
+      await Promise.all([
+        projects.map(async (project) => {
+          const { inclusions, exclusions } = getMatchingPatterns({
+            inclusions: project.contextSource.inclusions,
+            exclusions: project.contextSource.exclusions,
+            isProject: true,
+          });
+
+          if (shouldIndexFile(file, inclusions, exclusions)) {
```
logancyang, line 81: Does this mean if any note changes, all project's cache is cleared including static things like youtube, URLs and PDFs (coming next)? All those API calls are slow and expensive. The main purpose of the cache is actually for those.

src/cache/projectContextCache.ts
Emt-lin: i wil remove it.

src/cache/projectContextCache.ts
Emt-lin: so, we should refactor the context cache struct.

Emt-lin: @logancyang The project caching feature has been refactored.

Emt-lin:
@logancyang wait review.

logancyang:
@Emt-lin sorry for the late reply. Can you submit this PR to the private repo?

Emt-lin:
will resubmit this PR in the preview repo

PR #1446, opened 2025-04-13
Improve custom command (v3)

zeroliu:
Improve custom command UX based on recent feedback

- Trimmed AI responses. Some models such as gemini 2.5 include a line break at the end, causing replacing annoying to use.
- Made custom command AI response auto scroll while generating. Previously it's hard to know whether the AI output has completed when it's long.
- Supported custom prompt syntax in custom command. The legacy selected text syntax is made backward compatible.
- Updated custom command default template to use the custom prompt syntax.
- Supported followup instructions in the custom command.
- Added keyboard shortcuts to make it keyboard hardcore user friendly.
- Fixed a bug in the custom prompt process where invalid variable will cause variable index to mismatch.


logancyang:
It's great that it's backward compatible. One risk of having things inside user's setting is migration. This diff doesn't have any issue around that.

PR #1450, opened 2025-04-15
fix: Add a new line when press the Enter key on mobile.

Emt-lin:
## Bug

When typing in the chat interface on mobile, the return key seems to randomly alternates between:
- Adding a new line to the message
- Submitting the message to chat

This is frustrating as you can't predict which will happen, leading to accidentally sent incomplete messages.

## fixed

Provide consistent behavior: Add a new line when press the Enter key on mobile.


Emt-lin:
> For each PR, can you describe:
> 
> 1. what the issue is
> 2. how this PR fixes the issue

Added some description.

Issue #1445, opened 2025-04-12
Cannot See Model List in Model Tab


TomSpoct:
Copilot version:2.8.8

I cannot see the Model list in the model tab.

![Image](https://github.com/user-attachments/assets/0e75de21-dc0a-44da-b11a-ad59f1bf1933)

logancyang:
Can you try resetting your Copilot settings, looks like it's corrupted somehow.

TomSpoct:
I tried. I reinstalled everything. I can see models from version 2.7 but not 2.8.

Emt-lin:
assigned me @logancyang

PxlImp:
I have the same issue, seems to be a conflict with the Media Notes plugin: https://github.com/jemstelos/obsidian-media-notes

Emt-lin:
> I have the same issue, seems to be a conflict with the Media Notes plugin: https://github.com/jemstelos/obsidian-media-notes

Thanks for  providing the reproduction direction.

Issue #1408, opened 2025-03-28
The added models table is not shown in the Copilot Settings


ivannin:
- [x] Disable all other plugins besides Copilot **(required)**
- [x] Screenshot of note + Copilot chat pane + dev console added **(required)**

Copilot version: v2.8.8

The added models table is not shown in the Copilot Settings -- Model tab.

![Image](https://github.com/user-attachments/assets/d44986d1-4c68-4dc1-98f8-5ec71b04c78f)

logancyang:
Can you try a setting reset and see if it helps? Or reinstall the plugin.

ivannin:
> Can you try a setting reset and see if it helps? Or reinstall the plugin.

Yes, of course! I forgot to clarify that the screenshot was taken after a complete reset of the plugin settings and after its update today.

ivannin:
> Can you try a setting reset and see if it helps? Or reinstall the plugin.

Oh! I found it! This class hides the table! This is clearly programmatically set. If this class is removed, the table appears.

![Image](https://github.com/user-attachments/assets/a7bc756a-fbdf-432a-8f2a-8e3cd88e0d7f)

logancyang:
Is it on desktop or mobile? cc @Emt-lin

ivannin:
> Is it on desktop or mobile? cc [@Emt-lin](https://github.com/Emt-lin)

This is desktop

Emt-lin:
> > Is it on desktop or mobile? cc [@Emt-lin](https://github.com/Emt-lin)
> 
> This is desktop

@ivannin @logancyang there is css media query. when the screen width > 768px, it will automatically display. 

solution: I will try to reduce this number.

see: 

![Image](https://github.com/user-attachments/assets/33a6f468-3ffe-4127-8fae-2405b40f057f)

TomSpoct:
> > Can you try a setting reset and see if it helps? Or reinstall the plugin.
> 
> Oh! I found it! This class hides the table! This is clearly programmatically set. If this class is removed, the table appears.
> 
> ![Image](https://github.com/user-attachments/assets/a7bc756a-fbdf-432a-8f2a-8e3cd88e0d7f)

Can you tell me where to find and delete this ? Thanks

ivannin:
> > > Can you try a setting reset and see if it helps? Or reinstall the plugin.

[...]

PR #1467, opened 2025-04-23
fix: Use custom CSS naming to avoid being affected by other plugins or theme styles.

Emt-lin:
## Bug
The added models table is not shown in the Copilot Settings

Related: #1408、#1445

## Fixed

Use custom CSS name style instead common tailwind css name.

CleanShot 2025-04-23 at 20 07 17

### DIFF
--- a/src/components/chat-components/ChatInput.tsx
+++ b/src/components/chat-components/ChatInput.tsx
@@ -253,10 +253,14 @@ const ChatInput = forwardRef<{ focus: () => void }, ChatInputProps>(
       }
 
       if (e.key === "Enter" && !e.shiftKey) {
-        e.preventDefault();
-        onSendMessage(false);
-        setHistoryIndex(-1);
-        setTempInput("");
+        // Only prevent default and send message on non-mobile platforms
+        if (!Platform.isMobile) {
+          e.preventDefault();
+          onSendMessage(false);
+          setHistoryIndex(-1);
+          setTempInput("");
+        }
+        // On mobile, do nothing here, allowing the default newline behavior
       } else if (e.key === "ArrowUp") {
         if (currentLineIndex > 0 || selectionStart > 0) {
           // Allow normal cursor movement within multi-line input
--- a/src/settings/v2/components/ModelTable.tsx
+++ b/src/settings/v2/components/ModelTable.tsx
@@ -439,7 +439,7 @@ export const ModelTable: React.FC<ModelTableProps> = ({
 
   // Mobile view rendering
   const renderMobileView = () => (
-    <div className="md:hidden relative">
+    <div className="copilot-model-table-mobile relative">
       <DndContext
         sensors={sensors}
         collisionDetection={closestCenter}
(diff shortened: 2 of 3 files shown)


zeroliu:
Thanks for the context. That's super helpful.

My main concern is that tokens such as `hidden` are very common. While this PR removes one use case of hidden, there is no guarantee that we will use it elsewhere in our code base. As we confirmed that a generic token like `hidden` should be banned, we need to do something on the system level. A few possible solutions can be

1. Add a custom ESLint rule to ban `hidden` and point it at the new classname that we created.
2. Explore the feasibility of auto-adding the copilot prefix to the tailwind tokens during compilation. I remember there was a tailwind plugin for this.

Option 2 would be ideal because we often hear people's complaints about copilot conflicting with other themes. This will solve it once and for all

Emt-lin:
> Thanks for the context. That's super helpful.
> 
> My main concern is that tokens such as `hidden` are very common. While this PR removes one use case of hidden, there is no guarantee that we will not use it elsewhere in our code base. As we confirmed that a generic token like `hidden` should be banned, we need to do something on the system level. A few possible solutions can be
> 
> 1. Add a custom ESLint rule to ban `hidden` and point it at the new classname that we created.
> 2. Explore the feasibility of auto-adding the copilot prefix to the tailwind tokens during compilation. I remember there was a tailwind plugin for this.
> 
> Option 2 would be ideal because we often hear people's complaints about copilot conflicting with other themes. This will solve it once and for all

yeah, I will explore option 2.

Emt-lin:
> Thanks for the context. That's super helpful.
> 
> My main concern is that tokens such as `hidden` are very common. While this PR removes one use case of hidden, there is no guarantee that we will not use it elsewhere in our code base. As we confirmed that a generic token like `hidden` should be banned, we need to do something on the system level. A few possible solutions can be
> 
> 1. Add a custom ESLint rule to ban `hidden` and point it at the new classname that we created.
> 2. Explore the feasibility of auto-adding the copilot prefix to the tailwind tokens during compilation. I remember there was a tailwind plugin for this.
> 
> Option 2 would be ideal because we often hear people's complaints about copilot conflicting with other themes. This will solve it once and for all

For option 2: If we want to use tailwind's the prefix feature, It requires manually adding  prefix. 

To fix this bug, I think we can adopt the approach from this PR. cc @zeroliu  @logancyang

zeroliu (commented):
Can we rename the classname as tw-hidden instead of a new custom name knowing that we will move to the new prefix pattern?

Emt-lin:
> Can we rename the classname as tw-hidden instead of a new custom name knowing that we will move to the new prefix pattern?

md:tw-hidden not support, here must be use diffrent classNames. (here `tw-hidden` is customize className)

To quickly fixe this bug, we can do this for now until we move to the new prefix pattern. @zeroliu 

CleanShot 2025-05-15 at 09 32 12

zeroliu:
> > Can we rename the classname as tw-hidden instead of a new custom name knowing that we will move to the new prefix pattern?
> 
> 
> 
> md:tw-hidden not support, here must be use diffrent classNames. (here `tw-hidden` is customize className)
> 
> 
> 
> To quickly fixe this bug, we can do this for now until we move to the new prefix pattern. @zeroliu 
> 
> 
> 
> CleanShot 2025-05-15 at 09 32 12
> 
> 

Got it. Lgtm. Thanks for exploring the other options

Emt-lin:
@logancyang  this pr can be merge.

Issue #1540, opened 2025-06-12
Project Mode: Support project context preview


Emt-lin:
**Is your feature request related to a problem? Please describe.**
Currently, I don't know which files to include in the project, when i add the inclusion pattern.

see:

![Image](https://github.com/user-attachments/assets/d17c0c2b-3dde-40eb-bd17-ba5436e5880a)

**Describe the solution you'd like**
Support files preview when i add the inclusion pattern.

PR #1551, opened 2025-06-16
feat: Support project context preview.

Emt-lin:
## Issue

#1540

## Feature

- UI design: [V0](https://v0.dev/chat/note-and-file-ui-lqFJUoWoF6Z)
- actual effect：

https://github.com/user-attachments/assets/3baff94b-6ea6-40f0-bfa4-9fa592287cd1


Emt-lin:
> Can we remove the "extension" part? No need to pop "not supported", just remove it completely is fine.

If the user previously set up an extension, then removing the extension will hide this part of the context.  @logancyang

logancyang:
> > Can we remove the "extension" part? No need to pop "not supported", just remove it completely is fine.
> 
> If the user previously set up an extension, then removing the extension will hide this part of the context. @logancyang

It's ok, let's remove it. Having it doesn't make sense.

logancyang:
bugbot run

logancyang:
Let's prioritize merging this one by fixing those bugbot comments first @Emt-lin

Emt-lin:
> Let's prioritize merging this one by fixing those bugbot comments first @Emt-lin

fixed it. @logancyang

logancyang:
bugbot run

Emt-lin:
How can i run this `cursor bugbot` @logancyang

logancyang:
@Emt-lin only I can trigger it at the moment, the auto trigger isn't working. Let's see if it will allow proper auto triggers next.

Issue #1374, opened 2025-03-18
I hope each model can have its own temperature setting and include top-p.


Wooden-Gear:
希望每个模型都能设置单独的temperature，并加上top-p。
I hope each model can have its own temperature setting and include top-p.

logancyang:
@Emt-lin is doing it. Top-p fights with temperature. We will add frequency penalty instead.

Wooden-Gear:
> [@Emt-lin](https://github.com/Emt-lin) is doing it. Top-p fights with temperature. We will add frequency penalty instead.

还有输出token 长度，模型之间也不同，应该每个模型单独设置。

The output token length also varies between models, so each model should be configured individually.

this-is-sebs:
Hey Logan, 

I don't know that Top-P actually "fights" with temperature. They do interact but not to a severe degree. I agree that a frequency penalty would reduce randomness (dependent upon how it is used), but it does not inherently prevent the model from selecting a low-probability, nonsensical token if that token hasn't appeared frequently yet. The Top-P contains the sampling pool to the most probable candidates, which enhances the factual consistency of its output. The Top-P improves the coherence and reduces hallucinations by getting rid of the long tail of improbable next tokens. These are starkly different, with frequency penalty reducing  or increasing frequency of words, whereas Top-P actually changes the distribution of probable tokens. It would not hurt to add Top-P and configure it to default unless an individual wants to change it.

logancyang:
@this-is-sebs I agree, these are different and complementary parameters that are nice to have. My goal is not to make this plugin too complex for the avg user, but your point is valid, it's better to have them somewhere that's not cluttering the settings, more of a UIUX design question.

Wooden-Gear:
> [@this-is-sebs](https://github.com/this-is-sebs) I agree, these are different and complementary parameters that are nice to have. My goal is not to make this plugin too complex for the avg user, but your point is valid, it's better to have them somewhere that's not cluttering the settings, more of a UIUX design question.

top_p 和 min_p 这两个参数结合会是比较好的方案。
Combining the top_p and min_p parameters will provide a better solution.

Issue #1139, opened 2025-01-31
Add frequency penalty


misaka-20002:
- [x] Disable all other plugins besides Copilot **(required)**
- [x] Screenshot of note + Copilot chat pane + dev console added **(required)**

Copilot version: v2.8.2

(Bug report without the above will be closed)

**Describe how to reproduce**
use local ollama nomic embedding model and config OBCopilot CHAT OpenAI Format model supported by open-webui's openai compatable api
ask it a question

**Expected behavior**
Give normal answers

**Screenshots**

![Image](https://github.com/user-attachments/assets/1d397a9f-5cff-4d23-9456-8b8f95160847)

**Additional context**
None

misaka-20002:
the models are same qwen2.5:1.5b(btw with other models on open-webui it still gives this fxxking response)

misaka-20002:
It could related to prompt "system" role being passed wrongly

misaka-20002:
I manually copied the prompts that OBCopilot passed to the api and asked it in webchat. it works fine.

![Image](https://github.com/user-attachments/assets/a10af404-0bcd-41b7-b036-28bea8793b4c)

misaka-20002:
the frequency penality is 0 in OBCopilot requests and that could cause the problem,

misaka-20002:
After changing frequency penality to 1.1 the model goes well.

![Image](https://github.com/user-attachments/assets/92ec5bcb-faab-4672-944d-cab14d7ad0e1)
i suggest that adding frequency_penality config is necessary
@logancyang

misaka-20002:
Any other models shows same error including deepseekr1:14b as well as qwq-32b.
The local ollama works fine because it didnt pass the frequency penality argument to ollama but with openai format it did pass 0.
The problem solved after i manually modified the main.js, setting frequency penality to 1.1 instead of 0

Issue #1083, opened 2025-01-22
Allow editing of models


i13-ctflora:
**Is your feature request related to a problem? Please describe.**
I cannot see all of the details about my configured models from the UI.  I can't edit a model configuration through the UI if I discover a typo, but must instead delete it and re-enter it.  Also, many models from the same provider use the same configuration except for the model name - it would be much easier to enter these if I could duplicate the first one entered.

**Describe the solution you'd like**
In the 'Actions' column, add an Edit button and a Duplicate button.  
Edit opens the Add Custom Embedding Model dialog with the selected model's configuration pre-populated, allowing you to see & edit the settings.
Duplicate should do the same thing, but if you change the Model field, it saves as a new Model.

**Describe alternatives you've considered**
The only alternative I can see is to manually edit the Copilot data.json file, which is prone to errors.

**Additional context**

i13-ctflora:
Looks like this has been partially implemented, but I'm fine with what's here.  It looks like we can't edit the Provider, Base URL, or API Key fields, but can edit the rest.  And there's no 'duplicate' button.  But simply being able to easily edit the model & display name is a big win!

logancyang:
Editing of other fields will be added next! 

cc @Emt-lin this is in your plan right?

Emt-lin:
> Editing of other fields will be added next!
> 
> cc [@Emt-lin](https://github.com/Emt-lin) this is in your plan right?

yeah

PR #1562, opened 2025-06-20
feat: Support editing all parameters individually for each model

Emt-lin:
## Overview

Now, you can edit all  all parameters individually for each model and new models by duplicating existing ones.

actual effect: 

CleanShot 2025-06-20 at 18 09 08

## issues

- #1083
- #1374
- #1139

## Key changes

- support copy model for quicklly add.
- support edit maxToken、temperature、topP args for each model

### DIFF
--- a/src/aiParams.ts
+++ b/src/aiParams.ts
@@ -105,8 +105,9 @@ export interface CustomModel {
   stream?: boolean;
   temperature?: number;
   maxTokens?: number;
+  topP?: number;
+  frequencyPenalty?: number;
 
-  context?: number;
   projectEnabled?: boolean;
   plusExclusive?: boolean;
   believerExclusive?: boolean;
--- a/src/components/modals/project/AddProjectModal.tsx
+++ b/src/components/modals/project/AddProjectModal.tsx
@@ -16,6 +16,7 @@ import { HelpCircle } from "lucide-react";
 import { App, Modal, Notice } from "obsidian";
 import React, { useState } from "react";
 import { createRoot, Root } from "react-dom/client";
+import { DEFAULT_MODEL_SETTING } from "@/constants";
 
 interface AddProjectModalContentProps {
   initialProject?: ProjectConfig;
@@ -41,8 +42,8 @@ function AddProjectModalContent({ initialProject, onSave, onCancel }: AddProject
       systemPrompt: "",
       projectModelKey: "",
       modelConfigs: {
-        temperature: 1.0,
-        maxTokens: 6000,
+        temperature: DEFAULT_MODEL_SETTING.TEMPERATURE,
+        maxTokens: DEFAULT_MODEL_SETTING.MAX_TOKENS,
       },
       contextSource: {
         inclusions: "",
@@ -231,7 +232,7 @@ function AddProjectModalContent({ initialProject, onSave, onCancel }: AddProject
           <div className="tw-grid tw-grid-cols-1 tw-gap-4">
             <FormField label="Temperature">
               <SettingSlider
-                value={formData.modelConfigs?.temperature ?? 1}
+                value={formData.modelConfigs?.temperature ?? DEFAULT_MODEL_SETTING.TEMPERATURE}
                 onChange={(value) => handleInputChange("modelConfigs.temperature", value)}
                 min={0}
                 max={2}
@@ -241,7 +242,7 @@ function AddProjectModalContent({ initialProject, onSave, onCancel }: AddProject
             </FormField>
             <FormField label="Token Limit">
               <SettingSlider
-                value={formData.modelConfigs?.maxTokens ?? 6000}
+                value={formData.modelConfigs?.maxTokens ?? DEFAULT_MODEL_SETTING.MAX_TOKENS}
                 onChange={(value) => handleInputChange("modelConfigs.maxTokens", value)}
                 min={1}
                 max={65000}
--- a/src/components/modals/project/AddProjectModal.tsx
+++ b/src/components/modals/project/AddProjectModal.tsx
@@ -42,7 +42,7 @@ function AddProjectModalContent({ initialProject, onSave, onCancel }: AddProject
       projectModelKey: "",
       modelConfigs: {
         temperature: 1.0,
-        maxTokens: 1000,
+        maxTokens: 6000,
       },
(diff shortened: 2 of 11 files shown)


logancyang:
bugbot run

logancyang:
With these individual parameter settings, should the global one still exist?

Emt-lin:
> With these individual parameter settings, should the global one still exist?

Now, temperature and maxTokens is global parameters. For non-technical personnel, default values for these parameters are needed, and it should be convenient to quickly adjust all models. 

Once they understand the meaning of these parameters, they can independently configure the parameters for each model.

This is a step-by-step process.

logancyang:
According to our previous chat, we should either remove the global settings or do something else, the idea is to have only 1 source of truth.

logancyang (changes requested):
According to our previous chat, we should either remove the global settings or do something else, the idea is to have only 1 source of truth.

Emt-lin:
> According to our previous chat, we should either remove the global settings or do something else, the idea is to have only 1 source of truth.

I have removed the global settings for temperature and maxTokens

logancyang:
bugbot run

logancyang:
I just realized that embedding models are not editable like the chat models. We should make them editable too same as chat models.

src/settings/v2/components/ModelEditDialog.tsx
```diff
@@ -111,207 +100,293 @@ export const ModelEditDialog: React.FC<ModelEditDialogProps> = ({
...
+                      Al&#39;s vocabulary, the more diverse
+                    </div>
+                  </TooltipContent>
+                </Tooltip>
+              </TooltipProvider>
+            </div>
+          }
+        >
+          <SettingSlider
+            value={localModel.topP ?? 0.9}
```
logancyang: Why does the above say default value is 1 and this is showing 0.9? Also, should we have default values in our constant same as temp and max tokens?

logancyang:
Also don't forget to rebase on the new master.

src/settings/v2/components/ModelEditDialog.tsx
Emt-lin: > Why does the above say default value is 1 and this is showing 0.9?

already changed to 0.9.

> Also, should we have default values in our constant same as temp and max tokens?

we should not set default values formodel, because maybe some models are incompatiable with these attributes(like OpenAI o serise model).

The more default values are set, the higher the probability of errors. When the users only know about these parameters, it would be easier to set up correctly.

logancyang:
bugbot run

src/settings/v2/components/ModelEditDialog.tsx
logancyang: No I meant defining them as a constant rather than a magic number here and there, it makes maintaining the logic harder over time.

src/settings/v2/components/ModelEditDialog.tsx
```diff
@@ -109,209 +104,320 @@ export const ModelEditDialog: React.FC<ModelEditDialogProps> = ({
...
+                        </div>
                       </TooltipContent>
                     </Tooltip>
                   </TooltipProvider>
                 </div>
               }
             >
               <SettingSlider
-                value={localModel.temperature ?? 0.1}
+                value={localModel.temperature ?? settings.temperature ?? 1}
```
logancyang: Why 1? It's another magic number, if there's a reason we should name it something obvious.

Emt-lin: it's the result of autocompelete by ai. I forget to keep it consistent with the setting default value.(and also maxTokens )

Emt-lin:
@logancyang  Can you test the perplexity.ai API? It has supported the frequency_penalty value of 0.

docs: https://docs.perplexity.ai/api-reference/chat-completions-post#body-frequency-penalty

Google Chrome 2025-06-25 14 53 56

logancyang (approved):
Tested all models including perplexity, everything works

logancyang:
bugbot run

logancyang:
@Emt-lin can you address the bugbot comments above. And please also document the edge cases where some providers support one thing and not others.

Overall this is a welcomed change, we just need to be careful not to brick some models and make this clean and maintainable.

I'll merge once these are addressed.

Emt-lin:
> And please also document the edge cases where some providers support one thing and not others.

Add `method getProviderSpecificParams` to identify which parameters are supported by which Providers. @logancyang

src/aiParams.ts
```diff
@@ -102,8 +102,9 @@ export interface CustomModel {
   stream?: boolean;
   temperature?: number;
   maxTokens?: number;
+  topP?: number;
+  frequencyPenalty?: number;
 
-  context?: number;
```
logancyang, line 109: What's this? Are we using it anywhere

logancyang:
bugbot run

src/aiParams.ts
Emt-lin: @logancyang noting, so remove it

logancyang:
bugbot run

src/components/modals/project/AddProjectModal.tsx
```diff
@@ -42,7 +42,7 @@ function AddProjectModalContent({ initialProject, onSave, onCancel }: AddProject
       projectModelKey: "",
       modelConfigs: {
         temperature: 1.0,
-        maxTokens: 1000,
+        maxTokens: 6000,
```
logancyang: As a general comment for this PR and all future PRs, let's avoid having numbers in the logic directly wherever we can and define human readable constants `PROJECT_DEFAULT_MAX_TOKENS` etc. Without it we have to remember every hardcoded place or see future bugs and unexpected behavior pop up everywhere.

logancyang:
LGTM!

Could you rebase and also fix the "The arbitrary class 'tw-w-[1px]' could be replaced by 'tw-w-px'" stuff, it's popping up in every PR. Thanks!

Emt-lin:
> LGTM!
> 
> Could you rebase and also fix the "The arbitrary class 'tw-w-[1px]' could be replaced by 'tw-w-px'" stuff, it's popping up in every PR. Thanks!

fixed

logancyang:
bugbot run

PR #1566, opened 2025-06-21
Enhance abort signal in chains

logancyang:
Move from chain to chat model for better abort signal handling. Being less reliant on "chain" is the strategic move next for agents.


logancyang:
bugbot run

logancyang:
bugbot run

PR #1568, opened 2025-06-24
Fix word completion triggers

logancyang:
- Prevent suggestions for incomplete Obsidian wiki links.
- Allow native note completion to manage links in progress.
- Introduce checks for complete words with no significant completions to avoid unnecessary suggestions.


logancyang:
bugbot run

PR #1575, opened 2025-06-25
feat: Implement slash command detection and replacement in ChatInput

logancyang:
- Added unit tests for slash command detection and replacement logic.
- Enhanced ChatInput component to show a modal for slash commands and replace the slash with the corresponding command content.
- Updated showCustomPromptModal to handle cursor position and current input value for better user experience.


logancyang:
bugbot run

PR #1576, opened 2025-06-25
feat: Add support for selected text context in chat component

logancyang:
- Introduced a new command to add selected text to the chat context.
- Implemented functionality to manage selected text contexts within the plugin.
- Enhanced Chat and ChatContextMenu components to display and remove selected text contexts.
- Updated shared state to include selected text context structure.
- Improved user experience with notifications for added selections.

### DIFF
--- a/src/commands/contextMenu.ts
+++ b/src/commands/contextMenu.ts
@@ -1,10 +1,26 @@
 import { getCommandId, sortCommandsByOrder } from "@/commands/customCommandUtils";
+import { getCachedCustomCommands } from "@/commands/state";
+import { COMMAND_IDS } from "@/constants";
 import { Menu } from "obsidian";
 import { CustomCommand } from "./type";
-import { getCachedCustomCommands } from "@/commands/state";
 
 export function registerContextMenu(menu: Menu) {
+  menu.addItem((item) => {
+    item.setTitle("Copilot: Add selection to chat context").onClick(() => {
+      (app as any).commands.executeCommandById(
+        `copilot:${COMMAND_IDS.ADD_SELECTION_TO_CHAT_CONTEXT}`
+      );
+    });
+  });
+
+  // Add separator if there are custom commands too
   const commands = getCachedCustomCommands();
+  const visibleCustomCommands = commands.filter(
+    (command: CustomCommand) => command.showInContextMenu
+  );
+  if (visibleCustomCommands.length > 0) {
+    menu.addSeparator();
+  }
 
   sortCommandsByOrder(
     commands.filter((command: CustomCommand) => command.showInContextMenu)
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -1,4 +1,4 @@
-import { getChainType } from "@/aiParams";
+import { addSelectedTextContext, getChainType } from "@/aiParams";
 import { FileCache } from "@/cache/fileCache";
 import { ProjectContextCache } from "@/cache/projectContextCache";
 import { ChainType } from "@/chainFactory";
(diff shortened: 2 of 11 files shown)


logancyang:
bugbot run

logancyang:
bugbot run

src/commands/index.ts
```diff
@@ -285,4 +289,64 @@ export function registerCommands(
...
+  addEditorCommand(plugin, COMMAND_IDS.ADD_SELECTION_TO_CHAT_CONTEXT, async (editor: Editor) => {
+    // Check if we're in Copilot Plus mode
+    const currentChainType = getChainType();
+    if (
+      currentChainType !== ChainType.COPILOT_PLUS_CHAIN &&
+      currentChainType !== ChainType.PROJECT_CHAIN
+    ) {
+      new Notice("Selected text context is only available in Copilot Plus and Project modes");
+      return;
+    }
```
zeroliu, line 303: shall we only add this command if the user is in plus or project mode?

src/components/Chat.tsx
```diff
@@ -92,17 +93,34 @@ const Chat: React.FC<ChatProps> = ({
...
+  useEffect(() => {
+    const updateSelectedTextContexts = () => {
+      setSelectedTextContexts([...plugin.selectedTextContexts]);
+    };
+
+    // Initialize with current state
+    updateSelectedTextContexts();
+
+    // Subscribe to changes
+    const unsubscribe = plugin.subscribeToSelectedTextContextsChange(updateSelectedTextContexts);
```
zeroliu: We can create an atom to communicate the selected text with the component. I find the subscription model overkill and tedious to maintain. With atom, you can have the command set the value, and have this component use the atom value.

src/commands/index.ts
logancyang: @zeroliu yes but is there a better way to do this? I'm also thinking about adding it to the right-click menu, and it is (should be) independent of the mode selection.

logancyang:
bugbot run

logancyang:
@zeroliu I added it to the right-click menu as well. Not sure what's the best way / does it make sense to only have it for Plus and Projects mode, any suggestions?

src/commands/contextMenu.ts
```diff
@@ -1,10 +1,26 @@
 import { getCommandId, sortCommandsByOrder } from "@/commands/customCommandUtils";
+import { getCachedCustomCommands } from "@/commands/state";
+import { COMMAND_IDS } from "@/constants";
 import { Menu } from "obsidian";
 import { CustomCommand } from "./type";
-import { getCachedCustomCommands } from "@/commands/state";
 
 export function registerContextMenu(menu: Menu) {
+  menu.addItem((item) => {
```
zeroliu, line 8: You can conditionally register this command by checking the current mode.

src/commands/index.ts
zeroliu: you can move the chain check condition outside and conditionally add the editor command. Different from context menu where the items are added when triggering the right click, the command here is persisted. You will have to listen to the chain type change and add/remove this command accordingly. I think this one is less visible so what you have here is fine if we don't want to introduce this complexity.

src/commands/contextMenu.ts
logancyang: That's a bit weird imo, once we hide chat and vault qa modes to plus users this will be unnecessary.

src/commands/index.ts
logancyang: I'll keep this for now and we will probably 

- open up the chat context menu to free users next as our main free vs. paid strategy shifts to free vs paid MCP tools
- hide the free modes to plus users

PR #1581, opened 2025-06-27
Composer: fix compose block for empty note

wenzhengjiang:
Screenshot 2025-06-27 at 11 20 54

### DIFF
--- a/src/components/chat-components/ChatSingleMessage.tsx
+++ b/src/components/chat-components/ChatSingleMessage.tsx
@@ -279,10 +279,12 @@ const ChatSingleMessage: React.FC<ChatSingleMessageProps> = ({
                   if (
                     composerData.type === "composer" &&
                     composerData.path &&
-                    (composerData.content !== undefined || composerData.canvas_json)
+                    // `content` and `canvas_json` should never exist together
+                    (typeof composerData.content === "string" ||
+                      typeof composerData.canvas_json === "string")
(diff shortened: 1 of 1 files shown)


logancyang:
bugbot run

wenzhengjiang:
> The change from a truthy check (`composerData.content || ...`) to `composerData.content !== undefined` allows falsy values (e.g., `null`, `false`, `0`) to be assigned to `newContent`. Previously, these values would have triggered a fallback to `JSON.stringify(composerData.canvas_json)`. This now causes runtime errors because downstream operations (e.g., `split()`, `diffTrimmedLines()`) expect `newContent` to be a string.

`composerData.content` and `composerData.canvas_json` should never exist together, so I think this is fine.

wenzhengjiang:
bugbot run

wenzhengjiang:
> > The change from a truthy check (`composerData.content || ...`) to `composerData.content !== undefined` allows falsy values (e.g., `null`, `false`, `0`) to be assigned to `newContent`. Previously, these values would have triggered a fallback to `JSON.stringify(composerData.canvas_json)`. This now causes runtime errors because downstream operations (e.g., `split()`, `diffTrimmedLines()`) expect `newContent` to be a string.
> 
> `composerData.content` and `composerData.canvas_json` should never exist together, so I think this is fine.

Make some improvement on code clarify

PR #1582, opened 2025-06-27
feat(chat): Implement autosave for chat after messages and responses

Mathieu2301:
This PR introduces a more robust autosave feature for the chat.

Previously, autosave was only triggered when creating a new chat or reloading the plugin. This could lead to data loss if the user closed Obsidian or if the plugin crashed.

The new implementation saves the chat automatically at several key moments, ensuring the conversation is always up-to-date:

- After a user sends a message.
- After the AI finishes generating a response.
- After a user edits a message.
- After an AI response is regenerated.

This is controlled by the "Autosave Chat" setting, which has been updated to reflect the new behavior. The dependencies of the `useCallback` hooks for `handleEdit` and `handleRegenerate` have also been updated to include the new logic.

_I was inspired to make this change after losing several very important conversations multiple times in the same week. I was getting quite frustrated that the "Autosave Chat" setting didn't work in a way that felt logical, only saving on a new chat. I decided to fix it to create the behavior I, and likely many others, would expect, ensuring no more conversations are lost unexpectedly._

### DIFF
--- a/src/components/Chat.tsx
+++ b/src/components/Chat.tsx
@@ -248,7 +248,7 @@ const Chat: React.FC<ChatProps> = ({
 
     // Autosave the chat if the setting is enabled
     if (settings.autosaveChat) {
-      await handleSaveAsNote();
+      handleSaveAsNote();
     }
 
     await getAIResponse(
@@ -261,7 +261,7 @@ const Chat: React.FC<ChatProps> = ({
     );
     // Autosave the chat if the setting is enabled
     if (settings.autosaveChat) {
-      await handleSaveAsNote();
+      handleSaveAsNote();
     }
     setLoading(false);
     setLoadingMessage(LOADING_MESSAGES.DEFAULT);
@@ -433,7 +433,7 @@ ${chatContent}`;
 
       // Autosave the chat if the setting is enabled
       if (settings.autosaveChat) {
-        await handleSaveAsNote();
+        handleSaveAsNote();
       }
     },
     [
@@ -483,7 +483,7 @@ ${chatContent}`;
 
       // Autosave the chat if the setting is enabled
       if (settings.autosaveChat) {
-        await handleSaveAsNote();
+        handleSaveAsNote();
       }
     },
     [


logancyang:
bugbot run

Mathieu2301:
Ah I see, that makes sense! We should remove the await to make it non-blocking. That gives us a quick solution that keeps the crash protection without the performance impact.

src/components/Chat.tsx
```diff
@@ -246,6 +246,11 @@ const Chat: React.FC<ChatProps> = ({
       updateUserMessageHistory(inputMessage);
     }
 
+    // Autosave the chat if the setting is enabled
+    if (settings.autosaveChat) {
+      await handleSaveAsNote();
```
Mathieu2301:
```suggestion
     handleSaveAsNote();
```

src/components/Chat.tsx
```diff
@@ -254,6 +259,10 @@ const Chat: React.FC<ChatProps> = ({
       setAbortController,
       { debug: settings.debug, updateLoadingMessage: setLoadingMessage }
     );
+    // Autosave the chat if the setting is enabled
+    if (settings.autosaveChat) {
+      await handleSaveAsNote();
```
Mathieu2301:
```suggestion
     handleSaveAsNote();
```

src/components/Chat.tsx
```diff
@@ -422,8 +431,21 @@ ${chatContent}`;
       } finally {
         setLoading(false);
       }
+
+      // Autosave the chat if the setting is enabled
+      if (settings.autosaveChat) {
+        await handleSaveAsNote();
```
Mathieu2301:
```suggestion
       handleSaveAsNote();
```

src/components/Chat.tsx
```diff
@@ -459,8 +481,21 @@ ${chatContent}`;
       ) {
         handleRegenerate(messageIndex + 1);
       }
+
+      // Autosave the chat if the setting is enabled
+      if (settings.autosaveChat) {
+        await handleSaveAsNote();
```
Mathieu2301:
```suggestion
       handleSaveAsNote();
```

Mathieu2301:
During testing of the autosave feature, I discovered that the saved chat files were missing the last 2 messages (user message and AI response).
This was caused by the handleSaveAsNote function using the React state chatHistory instead of the up-to-date sharedState.

Fixed by:

- Modified handleSaveAsNote to use sharedState.getMessages() directly instead of chatHistory
- Updated useCallback dependencies to include sharedState instead of chatHistory

Also removed the "Chat updated..." toast notifications that appeared on every autosave, as they were becoming intrusive during regular usage. The "Chat saved..." notification for new files is preserved.

logancyang:
bugbot run

logancyang:
This is functionally fine for now but there's a significant UX issue that is - with autosave on, the user will see a notice "chat saved" on every message. We have to make it much more subtle before we introduce this change.

Mathieu2301:
> with autosave on, the user will see a notice "chat saved" on every message

I removed the "Chat updated..." toast notifications that appeared on every autosave

logancyang:
> > with autosave on, the user will see a notice "chat saved" on every message
> 
> I removed the "Chat updated..." toast notifications that appeared on every autosave

Oh I see, and the "chat saved as note ..." notice only pops up at the first message and not subsequent ones. That's fine I think. Thanks for the PR, merging now and let's see what the users say 🙏

PR #1589, opened 2025-06-28
fix(chat): Improved message editing UX by adding Escape key cancellation and removing auto-save on blur

Mathieu2301:
## Problem

I was frequently frustrated while editing messages I had sent to Copilot.
When looking up information in my notes, I would accidentally click outside the message being edited, which would cause it to lose focus and automatically save/send the half-modified message to Copilot. This generated unnecessary AI responses and disrupted my workflow. This happened multiple times and made message editing feel unreliable.

## Solution

Implemented a better message editing flow:

- Escape key: Cancels editing and restores the original message
- Enter key: Saves and sends the modified message
- Lost focus: No action taken (no more accidental saves)

This provides explicit user control over when edits are saved or cancelled, preventing accidental message submissions.

## Changes

- Added Escape key handler to cancel message editing
- Removed onBlur save behavior to prevent accidental saves
- Message editing now requires intentional user action (Enter or Escape)

## Testing

Tested in real conditions within my Obsidian vault.


logancyang:
bugbot run

PR #1592, opened 2025-06-30
fix: Disable autocomplete by default

logancyang:
And update activation logic based on Plus user status in AutocompleteService


logancyang:
bugbot run

PR #1594, opened 2025-07-01
Implement auto youtube tool

logancyang:

PR #1593, opened 2025-07-01
Fix auto index trigger

logancyang:
Fix a bug where auto indexing triggered 3 times at "Apply".

PR #1595, opened 2025-07-01
2.9.2

logancyang:
# Copilot for Obsidian - Release v2.9.2
A quick patch on top of v2.9.1. Now you don't need to manually `@youtube` to get the transcript, simply include the youtube url(s) in your chat message and their transcripts will be available in the context. (`@youtube <url>` for the transcript still works). Another critical fix is for free users - no more license key check popup if you happen to have autocomplete on.

Small UX improvement from our community contributor: improved message editing; autosave on current chat at every message to avoid loss of data in case of an app crash.
### PRs
- #1594 Implement auto youtube tool @logancyang
- #1589 Improved message editing UX by adding Escape key cancellation and removing auto-save on blur @Mathieu2301
- #1593 Fix auto index trigger @logancyang
- #1592 Disable autocomplete by default and prevent license key popup for free user @logancyang


logancyang:
bugbot run

PR #1600, opened 2025-07-02
Fix slash trigger

logancyang:
If chat input is not empty, `/` should only trigger after a space (e.g. avoid false trigger from pasted URL)


logancyang:
bugbot run

PR #1619, opened 2025-07-10
Fix copilot-plus-small

logancyang:
`copilot-plus-small` is not working because the dimension is truncated silently to 128 and it doesn't exist anymore on Voyage. Use 3.5 lite with dim 256.


logancyang:
bugbot run

logancyang:
bugbot run

PR #1620, opened 2025-07-10
Disallow file types in context

logancyang:
- Added `isAllowedFileForContext` utility function to check if a file is of allowed types (markdown, PDF, or canvas).
- Updated `ChatInput` to use this function for setting the current active note.
- Refactored `BaseNoteModal` to utilize the new utility for filtering files and managing the active note state.


logancyang:
bugbot run

PR #1634, opened 2025-07-19
Fix context processing blocking user message UI update

logancyang:
- Introduced a new method `setOnMessageCreatedCallback` in ChatManager to allow setting a callback that triggers when a message is created.
- Updated ChatUIState to utilize this callback for notifying listeners, ensuring the UI updates immediately upon message creation.
- This enhancement improves the responsiveness of the chat interface by synchronizing message creation events with UI state updates.


logancyang:
bugbot run

PR #1640, opened 2025-07-20
Quick Command (cmd+k)

zeroliu:
Replace the legacy "apply adhoc custom prompt" with the brand new "quick command". It allows users to insert a quick prompt and use it with the custom command workflow.

The quick command is implemented by introducing a new `copilotquickcommand` code block processor. In the editor, we will add the code block when the quick command is triggered and the processor will render it like cursor cmd+k popover above the selected text.

This PR also improves a few things:
- Move context menu items to "copilot" submenu
Screenshot 2025-07-20 at 1 35 18 AM

- Use XML tags for prompt processor. It makes the variables more recognizable by models.
- Clean up legacy code
- Extract model selectors component to a reusable component.


zeroliu:
bugbot run

PR #1641, opened 2025-07-20
Fix embedded image wikilinks

logancyang:
- Updated the `extractEmbeddedImages` method to accept an optional `sourcePath` parameter for resolving wikilinks.
- Implemented logic to determine the source path from context notes or fallback to the active file.
- Enhanced logging for unresolved images to improve debugging and user feedback.
- Ensured that embedded images are processed correctly based on the new source path handling.


logancyang:
bugbot run

PR #1654, opened 2025-07-23
fix: Fix the shouldIndexFile method to return all files when inclusions is empty in project mode.

Emt-lin:


logancyang:
Why is the test failing in CI?

Emt-lin:
> Why is the test failing in CI?

I don't know, it appears to be related to `composer test file`. It might be due to unstable model responses(`gemini-2.0-flash`).

logancyang:
But why only this PR fails and not others?

logancyang:
can we also make this pr against full-agent branch

Emt-lin:
will reopen in `full-agent` branch

PR #1667, opened 2025-07-28
fix: Fix the shouldIndexFile method to return all files when inclusions is empty in project mode

Emt-lin:


Emt-lin:
@logancyang I don't know why it failded again.

logancyang:
@wenzhengjiang any idea why only this PR keeps failing that composer test?

PR #1694, opened 2025-08-05
Add tool call marker for the tool call that is being generated

wenzhengjiang:

Add tool call markers within the `AutonomousAgentChainRunner` and enhance XML parsing utilities to support extracting tool names and managing partial XML tool call content.

These changes enhance the system's ability to handle dynamic tool calls by identifying and marking tool calls during runtime, providing context and facilitating better tracking of execution states. The refactored XML parsing functions now better manage partial tool call content, handling edge cases like empty names or whitespace, improving the robustness and accuracy of tool name extraction.

### DIFF
--- a/.cursor/rules/test-rule.mdc
+++ b/.cursor/rules/test-rule.mdc
@@ -1,0 +1,4 @@
+---
+alwaysApply: true
+---
+Use "npx" command to run test
--- a/src/LLMProviders/chainRunner/AutonomousAgentChainRunner.ts
+++ b/src/LLMProviders/chainRunner/AutonomousAgentChainRunner.ts
@@ -24,7 +24,11 @@ import {
   ToolExecutionResult,
 } from "./utils/toolExecution";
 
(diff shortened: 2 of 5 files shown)


.cursor/rules/test-rule.mdc
```diff
@@ -0,0 +1,4 @@
+---
```
logancyang, line 1: remove?

logancyang:
I observe a gap between "Calling" and "Called", after the first disappears there can be a several-second gap where no banner is displayed. It happens with Claude 4 but not with gemini 2.5 flash / pro. Interesting. Could it be caused by the think block somehow?

wenzhengjiang:
> I observe a gap between "Calling" and "Called", after the first disappears there can be a several-second gap where no banner is displayed. It happens with Claude 4 but not with gemini 2.5 flash / pro. Interesting. Could it be caused by the think block somehow?

Hmm.. I couldn't see the gap between "Calling" and "Called" with Claude 4 (open router). Do you have a recording?

logancyang:
Video shared in Discord

PR #1746, opened 2025-08-25
Support image and chat context menu in free modes

logancyang:
- Support image and chat context menu in free modes
- Restrict url and non-markdown processing
- Update chat input button tooltips

Enable support for image and chat context menus in both free and Plus modes by updating the message handling and chain type checks, while including additional restrictions and feedback for non-Plus users.

The intention is to expand the context menu capabilities to all users, while maintaining certain premium features exclusive to Plus users, such as URL and non-markdown file processing. This approach ensures a broader feature set available to free users while preserving the value of the Plus subscription. Notably, adjustments include better handling of multimodal content in message processing and improved restriction notices for non-Plus users when using unsupported features.


logancyang:
bugbot run

PR #1759, opened 2025-08-27
Update 'replaceInFile' tool description to indicate experimental status and remove it from default tool list

wenzhengjiang:
'replaceInFile' has been reported being unreliable on copilot-flash. Disable it by default and mark it as experimental until we figure out a better approach.

Update the 'replaceInFile' tool description to indicate its experimental status and remove it from the default tool list.

The 'replaceInFile' tool is currently experimental and may only function properly with very advanced models, thus it's being removed from the default settings to prevent potential issues. Updating the description to reflect its experimental nature provides clarity to users and sets proper expectations regarding its current capabilities.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -745,7 +745,6 @@ export const DEFAULT_SETTINGS: CopilotSettings = {
     "pomodoro",
     "youtubeTranscription",
     "writeToFile",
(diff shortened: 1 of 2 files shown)


wenzhengjiang:
Actually, let me try me improve it first.

PR #1766, opened 2025-08-27
Fix context not being sent to LLM for `[[note]]` references

### LINKED ISSUE #1763: The new version cannot read the local note file
Android，copilot Does not read local note files.Whether I use [[]] or just click the input box to add a file, the plugin cannot read file.It used to be normal, but it will be like this after updating to 3.0.
Includes some built-in reference symbols, which also fail.
The screenshot is as follows:
Image

aviatesk:
- Update `LLMChainRunner` to use processed message text (with context) for multimodal content
- This ensures that when using `[[note title]]` syntax, the note content is properly included in LLM requests

Closes #1763.


logancyang:
bugbot run

logancyang:
Thanks for the catch and PR! Great stuff!

PR #1794, opened 2025-09-08
Zero/new chat input

zeroliu:


logancyang:
bugbot run

zeroliu:
> bugbot run

@logancyang please hold on bugbot for now. The PR is not done yet

logancyang:
bugbot run

zeroliu:
@logancyang I realize the pdf context support is broken. Let me fix it

zeroliu:
@logancyang it's good now

logancyang:
@zeroliu This is a HUGE upgrade to our chat input, thanks for the hard work!! 🔥🔥🔥 

I clicked around a bit and here are some initial findings:

1. Seems I can't add a url as plain text. When i paste, it becomes a badge; when i backspace, it gets removed completely as a badge.
SCR-20250924-okjv

2. Typing `@` can enter the tool in the input, but clicking by mouse doesn't work.
SCR-20250924-okcj

3. Adding notes or urls makes the "@ Add Context" button collapse to just "@", but Folders or Tags don't

SCR-20250924-ojkx
SCR-20250924-ojrc

4. When i type `[[` the popup shows 4-5 candidates with previews, this search works differently from the `@ Add Context` button. For example, if i type `[[pdf` it won't find pdf files, but `@ Add Context` can find them.

5. If i open the settings UI while having the context picker and preview on, they stay there like this
SCR-20250924-opic

6. Could you remind me what we have settled on for the old `@youtube <url>` behavior - getting full raw transcript? How do we achieve it now?

logancyang:
These are the only problems i can find now. Overall superb stuff!! 🔥🔥🔥

PS there's a conflict to be resolve since the memory change has been merged to that 3.1.0 branch.

zeroliu:
> @zeroliu This is a HUGE upgrade to our chat input, thanks for the hard work!! 🔥🔥🔥
> 
> I clicked around a bit and here are some initial findings:
> 
> 1. Seems I can't add a url as plain text. When i paste, it becomes a badge; when i backspace, it gets removed completely as a badge.
> 
> SCR-20250924-okjv

This one is intentional. I checked the latest cursor and it doesn't contain the old behavior of converting a URL pill to plain text any more. I suspect that if a URL is removed from the context, most likely you want to remove it from the context too because the URL itself doesn't add much value to the context. If we decide in the future to process the chat content and fetch the plain text URL, the current behavior is future-proof, too. WDYT?


> 2. Typing `@` can enter the tool in the input, but clicking by mouse doesn't work.
> 
> SCR-20250924-okcj

It's actually working. The tool is toggled and the tool button is updated. This is intentional because I don't want the tool pill to show up in the context menu since the bottom tool button already shows that. I understand why it's misleading though. In the latest version, when you click on a tool option in the context menu, it will add a tool pill in the chat. This would work for @memory where there is no tool button. It's still a bit different from other type of context but I think it's the best option so far.

> 3. Adding notes or urls makes the "@ Add Context" button collapse to just "@", but Folders or Tags don't
> 
> SCR-20250924-ojkx SCR-20250924-ojrc

Fixed!

> 4. When i type `[[` the popup shows 4-5 candidates with previews, this search works differently from the `@ Add Context` button. For example, if i type `[[pdf` it won't find pdf files, but `@ Add Context` can find them.
> 5. If i open the settings UI while having the context picker and preview on, they stay there like this
> 
> SCR-20250924-opic

Fixed!

> 6. Could you remind me what we have settled on for the old `@youtube <url>` behavior - getting full raw transcript? How do we achieve it now?

We decided to introduce a new download youtube video script command. 

Screenshot 2025-09-25 at 1 01 30 AM

Screenshot 2025-09-25 at 1 01 49 AM

logancyang:
@zeroliu Merged! Made it to the 3.1.0 preview 🔥

About the literal url, Cursor has the "unlink" option to pass url as plain text. But let's ship this for now and see if people actually complain about it.
SCR-20250925-llno

Emt-lin:
1. This should be a display bug. When I hover over an item and then use the arrow keys to move up or down, the hover state of the item under the mouse remains.
Obsidian 2025-09-26 23 37 42

2. on mobile,  there are some UI  issues for the `@` portal.

CleanShot 2025-09-27 at 00 02 49

3. for the `@` context portal, maybe It is a little wide. Is it to previe the note content?

CleanShot 2025-09-27 at 00 10 37

PR #1796, opened 2025-09-08
FIX: Supporting creating new folders in composer tools

wenzhengjiang:
TESTED: 
* note in new folder 
* note in new nested folder
* note in existing folder

Refactor `show_preview` function to create missing directories while obtaining a file, enabling folder creation before file creation in the `ComposerTools`.

The changes address the issue of folder paths not being handled correctly when creating new files, which led to file creation failures. By introducing a `getFile` function that ensures folders exist before creating a file, the refactor provides a more robust solution to file operations in the `ComposerTools` module.

### DIFF
--- a/src/tools/ComposerTools.ts
+++ b/src/tools/ComposerTools.ts
@@ -4,21 +4,40 @@ import { diffTrimmedLines } from "diff";
 import { ApplyViewResult } from "@/types";
 import { z } from "zod";
 import { createTool } from "./SimpleTool";
+import { ensureFolderExists } from "@/utils";
 
 async function getFile(file_path: string): Promise<TFile> {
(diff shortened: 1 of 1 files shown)


logancyang:
bugbot run

logancyang:
@wenzhengjiang there's a new `ensureFolderExists` util in master that ensures successful folder creation on different platforms (win and linux used to fail in some cases). Probably needed in this code too.

wenzhengjiang:
> @wenzhengjiang there's a new `ensureFolderExists` util in master that ensures successful folder creation on different platforms (win and linux used to fail in some cases). Probably needed in this code too.

Nice! I updated the code to use this new function.

PR #1828, opened 2025-09-20
Disable "@ for tools" in agent mode

wenzhengjiang:
"@ for tools" conflicts with agent mode. We should disable them to provide a more consistent experience.

Tested manually


logancyang:
Why do we want to disable `@ for tools` in agent mode again? I think we should support manual trigger in agent mode as well.

wenzhengjiang:
Close this PR as discussed in the chat group.

PR #1829, opened 2025-09-21
Agent Mode: Map copilot @command to builtin agent tools

wenzhengjiang:
Implement a consistent user experience where users can use `@command` to explicitly trigger a tool call in both agent mode and normal plus mode.

Tested manually with '@vault', '@websearch' and '@composer'. 

NOTE: This does not change the current toggle restriction.


wenzhengjiang:
bugbot run

Issue #409, opened 2024-04-09
Question about the parameter num_ctx setttings for Ollama models


luc99hen:
As the doc explains "AFAIK ollama serve doesn't have a consolidated way to configure context window for all the models at a single place. The current best way is to run ollama run <modelname> and then /set parameter num_ctx 32768 (this is the max for Mistral, set it based on your model requirement), and don't forget to /save <modelname> for each model individually."

32768 works perfectly for Mistral, but not for other models like Qwen and llama. I want to know how can I find a proper num_ctx for other models.

logancyang:
@luc99hen you can find the model specifications online from their providers, e.g. huggingface model card or whatever platform they publish them.

For future reference, pure questions are better posted in the discussion section. Thanks!

PR #1849, opened 2025-09-25
Fix chat crash React 409

logancyang:
# React Invariant `#409` on Tool Call Messages

## Summary

- **Symptom:** Submitting a chat message containing historical tool-call banners triggered `Minified React error #409` in the Obsidian developer console. The error appeared whenever the renderer attempted to update banners that were created in a previous session.
- **Root Cause:** `ChatSingleMessage` cached React roots for each tool-call banner in a global map. The hourly cleanup routine removed references to entries that were older than one hour but left the underlying DOM nodes intact. When the message re-rendered, the component retrieved the stale root from the map and invoked `root.render(...)`. React throws invariant `#409` when you render into a root that has already been unmounted.
- **Impact:** Any assistant response that referenced a stale tool-call banner crashed the renderer, preventing the chat view from updating until the pane was reloaded.

## Diagnosis

1. Reproduced the failure by re-opening a note that contained an earlier assistant response with tool-call markers. The console displayed invariant `#409` each time the message rendered.
2. The stack trace pointed at `ChatSingleMessage`, the component responsible for streaming assistant output and tool-call banners.
3. Inspection of `rootsRef.current` revealed entries whose `_internalRoot` was already `null`. The cleanup job had unmounted the React roots without removing their DOM containers.

## Fix

| File | Description |
| ---- | ----------- |
| `src/components/chat-components/toolCallRootManager.tsx` | New shared manager that stores `{ root, isUnmounting }` records, encapsulates the React root lifecycle, and exposes helpers to render, remove, and clean up tool-call banners. |
| `src/components/chat-components/ChatSingleMessage.tsx` | Tool-call rendering now delegates to the manager helper to finish pending disposals and only spin up new roots when the cache has none. |
| `src/components/chat-components/ChatSingleMessage.tsx` | Cleanup paths (per-tool, hourly purge, component unmount) call into the manager so entries are marked unmounting, disposed on the next tick, and removed from the registry only after disposal completes, preventing re-renders from touching half-disposed roots.

The new logic relies on strict cache hygiene—roots are marked as unmounting, disposed, and only then removed from the map—so future renders either reuse a healthy root or build a fresh one without touching private React internals. Logging remains centralized through `logWarn`, eliminating the invariant crash while keeping the streaming UX intact.

## Verification

- `npm test -- src/components/chat-components/ChatSingleMessage.test.tsx`
- Manual chat submission with historical tool-call banners (pending).

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -318,64 +318,26 @@ function updateMcpToolSetting(toolId: string, enabled: boolean) {
 3. **Settings UI**: `ToolSettingsSection` component reads from the registry to generate UI
 4. **Tool Execution**: `AutonomousAgentChainRunner.getAvailableTools()` filters tools based on settings
 
-### Tool Execution Flow
+## Tool Call Rendering Roots
 
-1. Agent calls `getAvailableTools()` which:
+React invariant #409 surfaced when tool-call banners attempted to render into React roots that had already been unmounted. To prevent this regression, the plugin routes all banner rendering through the shared manager in `src/components/chat-components/toolCallRootManager.tsx`.
 
-   - Gets enabled tool IDs from settings array (`autonomousAgentEnabledToolIds`)
-   - Calls `registry.getEnabledTools()` to get actual tool implementations
-   - Filters based on vault availability and user preferences
+### Manager Responsibilities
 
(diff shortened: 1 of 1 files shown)


logancyang:
bugbot run

PR #1873, opened 2025-10-01
Rebuild "Active Note" and more chat input improvement

zeroliu:
- Active note logic is rebuilt. It becomes its own context type. It's now added to the new @ mention typeahead menu.
- Custom command notes are ranked lower when there is no search query.

Screenshot 2025-09-30 at 11 19 02 PM

### DIFF
--- a/src/components/chat-components/AtMentionTypeahead.tsx
+++ b/src/components/chat-components/AtMentionTypeahead.tsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useEffect, useState } from "react";
+import { TFile } from "obsidian";
 import { TypeaheadMenuPopover } from "./TypeaheadMenuPopover";
 import {
   useAtMentionCategories,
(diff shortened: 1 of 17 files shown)


logancyang:
bugbot run

logancyang:
Nice step forward! Another similar thing missing is - if active note isn't in the context menu yet, typing `[[` should have the active note at the top of the list. Recently i constantly wanted to add it this way but the list given by `[[` doesn't have any "UI context awareness". Ideally, `[[` may have a section at the top that is every note in the active tabs. WDYT? @zeroliu

zeroliu:
> Nice step forward! Another similar thing missing is - if active note isn't in the context menu yet, typing `[[` should have the active note at the top of the list. Recently i constantly wanted to add it this way but the list given by `[[` doesn't have any "UI context awareness". Ideally, `[[` may have a section at the top that is every note in the active tabs. WDYT? @zeroliu

That makes sense. I can add the same active note option to the [[ typeahead menu. Alternatively, we can keep [[ to only add static note, but as you pointed out, many users already have the habit of using [[ to find notes.

zeroliu:
can I do this as a separate PR? Don't want to expand the scope of this PR any more

PR #1914, opened 2025-10-11
Turn on memory feature by default

wenzhengjiang:


logancyang:
The `sanitizeSettings` in model.ts should also be updated in case the setting is missing

wenzhengjiang:
> The `sanitizeSettings` in model.ts should also be updated in case the setting is missing

Done.

PR #1966, opened 2025-10-25
Reduce images size in half in README without losing quality

ichts:


images/AI-Model-Setup.png
logancyang, line 1: hmm, are we still using this old one?

images/Copilot-Plus-Setup.png
logancyang, line 1: ditto, i don't see this one used in readme?

images/AI-Model-Setup.png
ichts: No, should I delete the non used images?

logancyang: yes please

PR #2026, opened 2025-11-12
Fix auto scrolling

logancyang:
Before this change: auto scrolls to the bottom when AI is streaming, and when it finishes it goes back to the user message.

After: user message moves to the top and there's no further auto scrolling as the AI is streaming.

This prevents crazy oscillations with long AI streaming responses, especially when images are in the stream.

### DIFF
--- a/src/hooks/useChatScrolling.ts
+++ b/src/hooks/useChatScrolling.ts
@@ -140,6 +140,11 @@ export const useChatScrolling = ({
 
   // Scroll to bottom only when user messages are added
   useEffect(() => {
+    // Reset ref if chat history shrinks (new chat, deleted messages)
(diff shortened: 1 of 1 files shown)


zeroliu:
Can you please share some context about this bug? I'm not aware that we have new chat scrolling issues

logancyang:
> Can you please share some context about this bug? I'm not aware that we have new chat scrolling issues

i don't know how it happened either, if you test the current master it's as described in the pr description, not sure what made it like that?

zeroliu:
Given the discovery here: https://discord.com/channels/1182847556032659546/1332201585899732992/1438281103285817436

Shall we revert the culprit PR instead?

logancyang:
@zeroliu #2015 is reverted (#2038)

Issue #2040, opened 2025-11-22
Thinking Blocks reasoning text is flattened into normal assistant reply after cross-device sync


vuhe:
- [x] Disable all other plugins besides Copilot **(required)**
- [x] Log file generated via "Copilot: Create Log File" command or Settings -> Advanced -> Create Log File **(required)**
- [ ] Screenshot of note + Copilot chat pane + dev console added **(optional)**

Copilot version: 3.1.3
Model used: deepseek-reasoner

[copilot-log.md](https://github.com/user-attachments/files/23687355/copilot-log.md)

**Describe how to reproduce**

1. On device A, open a note, open the Copilot chat pane, enable Thinking Blocks, and ask a question.
2. Wait for the AI response to finish; you can see the Thinking Blocks UI plus the normal answer body.
3. Sync the vault and then open the same note and Copilot chat on device B (with the same vault synced via a cloud/sync service).
4. On device B, reload the chat: the previous Thinking Blocks content is now rendered as part of the assistant’s normal reply text, without any separation or special styling, making it indistinguishable from the main answer.

**Expected behavior**

When reopening the same conversation on another device, Thinking Blocks should remain as a separate, clearly marked/hidden reasoning area (or be omitted entirely if disabled), and only the main answer body should appear as the assistant’s reply. The chat view and separation between Thinking Blocks and answer text should be consistent across devices.

**Screenshots**

Not available on my current setup, so I cannot provide screenshots.

**Additional context**

This only seems to happen when viewing the same conversation on a different device from the one where the response was originally generated, so it may be related to how Thinking Blocks metadata is serialized into the note and then parsed after sync. The fact that the local conversation record itself already has thinking content and answer text mixed together suggests the flattening might be happening at the time of saving or syncing, rather than purely in the UI.

logancyang:
@Emt-lin can you reproduce this one?

Emt-lin:
> [@Emt-lin](https://github.com/Emt-lin) can you reproduce this one?

yeah, i can reproduce. i will try to fix it.

logancyang:
Thanks for reporting this, and sorry for the slow reply.

Copilot v3 is in code freeze, so we're closing this as not planned. v4 is a ground-up rewrite and is launching soon, with a very different architecture around agent mode.

If this is still an issue for you in v4 once it lands, please open a new issue and we'll take a fresh look. 🙏

PR #2099, opened 2026-01-07
fix: always process think blocks regardless of current model selection.

Emt-lin:
## issues
This issue was introduce by this PR(#2024).
- #2040

### Summary
- Removes the logic that conditionally processes `<think>` blocks based on the currently selected model's REASONING capability
- Think blocks are now always processed, ensuring consistent rendering regardless of which model is currently selected

### Problem
Previously, think blocks were only rendered if the currently selected model had the `REASONING` capability. This was incorrect because:
1. Messages may have been generated by a different model than the one currently selected
2. Switching models would cause think blocks to appear/disappear unexpectedly
3. The rendering of historical messages should not depend on current model selection

### Changes
- `ChatSingleMessage.tsx`: Remove `shouldProcessThinkBlocks` logic and related imports (`ModelCapability`, `findCustomModel`, `useModelKey`, `useMemo`)
- Think blocks are now unconditionally processed via `processThinkSection()`

### Test Plan
- [x] Send a message with a reasoning model (e.g., Claude with extended thinking) to generate think blocks
- [x] Switch to a non-reasoning model
- [x] Verify think blocks still render correctly in the chat history

### DIFF
--- a/src/components/chat-components/ChatSingleMessage.tsx
+++ b/src/components/chat-components/ChatSingleMessage.tsx
@@ -26,15 +26,14 @@ import {
   renderToolCallBanner,
   type ToolCallRootRecord,
 } from "@/components/chat-components/toolCallRootManager";
-import { ModelCapability, USER_SENDER } from "@/constants";
+import { USER_SENDER } from "@/constants";
 import { cn } from "@/lib/utils";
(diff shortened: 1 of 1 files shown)


Emt-lin:
@codex review

PR #2102, opened 2026-01-11
Self host alpha

logancyang:
- Add self-host toggle and logic
- Implement retriever factory for miyo support next and proper fallback logic


logancyang:
@codex review

PR #2121, opened 2026-01-19
Dev anna1124

AnnaHsu-Phison:


logancyang:
This plugin has the AGPL-3.0 license, you cannot fork and just call it your own.

PR #2147, opened 2026-02-04
Update ollama

logancyang:
Update ollama chat client to fix thinking block, native tool calls, context length setting and repetition penalty


logancyang:
@codex review

PR #2148, opened 2026-02-04
feat: add streamUsage config for 3rd party OpenAI-format providers

### LINKED ISSUE #2046: Add advanced config for models for streaming and streamUsage
Right now if you want to use a custom LLM provider that supports the OpenAI format, it works provided that provider does 2 things:  Supports streaming and supports the stream options config.

obsidian-copilot is using langchain/openai package that defaults the config `streamUsage: true` in the constructor.  I have found that if your llm provider does not support stream options you will get the following error:

`{"message":"stream_options: Extra inputs are not permitted"}`

The quick fix is to edit the data.json file in the plugin directory and set the value of "stream" to false example:

```json
{
      "name": "databricks-claude-sonnet-4",
      "provider": "3rd party (openai-format)",
      "enabled": true,
      "isBuiltIn": false,
      "baseUrl": "https://<your-workspace-id>.azuredatabricks.net/serving-endpoints",
      "apiKey": "<wouldn't you like to know>,
      "isEmbeddingModel": false,
      "capabilities": [],
      "stream": false,
      "displayName": "DBX Claude Sonnet 4",
      "enableCors": true
    }
```

This allows a custom llm that doesn't support stream_options to work, however the real error is how this plugin is using the langchainjs library as noted by this issue:  https://github.com/langchain-ai/langchainjs/issues/6533

There should be some advanced configuration for OpenAI format providers to allow the option to turn off streaming and allow for the user to configure if stream_options should be passed as suggested by this fix in the issue above:  https://github.com/langchain-ai/langchainjs/issues/6533#issuecomment-2290025841

To get around this limitation I currently do the following:
1. Quick solution is to edit the data.json file and find my model and set the value of `stream` to `false`.
2. To get streaming to work I edited the minified plugin js `main.js` and found the `getModelConfig` function and manually added the following to that function `streamUsage:false,` after the `maxRetries:3,` line in the code.  Essentually adding it this line of the code:  https://github.com/logancyang/obsidian-copilot/blob/e05e3fafefbddc2aab03d91abb30828d4da27289/src/LLMProviders/chatModelManager.ts#L150

**Additional context**
I think with all of the options for hosting your own llm's this would be very handy and this is how under the hood MLFlow allows you to run your own models from hugging face.

wotan-allfather:
Add optional streamUsage setting to CustomModel interface that allows users to control whether stream_options is sent to the API. This fixes compatibility with 3rd party providers (e.g., Databricks, MLFlow) that do not support the stream_options parameter.

- Add streamUsage?: boolean to CustomModel interface in aiParams.ts
- Pass streamUsage to ChatOpenAI config for OPENAI_FORMAT and LM_STUDIO providers
- Add UI toggle in ModelAddDialog and ModelEditDialog with helpful tooltip
- Default to false to maintain compatibility with providers that error on stream_options

Closes #2046

### DIFF
--- a/src/LLMProviders/chatModelManager.ts
+++ b/src/LLMProviders/chatModelManager.ts
@@ -298,6 +298,7 @@ export default class ChatModelManager {
       [ChatModelProviders.LM_STUDIO]: {
         modelName: modelName,
         apiKey: customModel.apiKey || "default-key",
+        streamUsage: customModel.streamUsage ?? false,
         configuration: {
           baseURL: customModel.baseUrl || "http://localhost:1234/v1",
           fetch: customModel.enableCors ? safeFetch : undefined,
@@ -306,6 +307,7 @@ export default class ChatModelManager {
       [ChatModelProviders.OPENAI_FORMAT]: {
         modelName: modelName,
         apiKey: await getDecryptedKey(customModel.apiKey || settings.openAIApiKey),
+        streamUsage: customModel.streamUsage ?? false,
         configuration: {
           baseURL: customModel.baseUrl,
           fetch: customModel.enableCors ? safeFetch : undefined,
--- a/src/aiParams.ts
+++ b/src/aiParams.ts
@@ -124,6 +124,7 @@ export interface CustomModel {
   enableCors?: boolean;
   core?: boolean;
   stream?: boolean;
+  streamUsage?: boolean;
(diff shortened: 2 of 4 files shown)


Emt-lin:
@codex review

logancyang:
Thanks for the PR!

Issue #11, opened 2023-05-12
Struggling to get set up - open api 429 error


TheMany172:
Hello,

Apologies that this may well be a 'me setting it up wrong' thing

But I cannot get the plugin to work

I have downloaded the files and moved them into the directory specified - I also tried using Obsidian BRAT when it didnt work

I have logged into chatGPT online and made an API key and pasted it in the settings for copilot and have it enabled etc

when I open it and enter a chat I get an error popup saying open api error: null pls see console etc

the console for it is:

```
Error in streamSSE: {
    "error": {
        "message": "You exceeded your current quota, please check your plan and billing details.",
        "type": "insufficient_quota",
        "param": null,
        "code": null
    }
}

getAIResponse @ plugin:copilot:37823
```

the error is quite clear, but I can log in and use chat gpt just fine, i am on a normal free plan and don't really use it much yet, so shouldn't be beyond whatever their free amount is

Is there some setting somewhere I am failing to see/change?

logancyang:
No worries! I just googled this error message and this is the page I was looking at https://help.openai.com/en/articles/6891831-error-code-429-you-exceeded-your-current-quota-please-check-your-plan-and-billing-details

My guess is that you have not set up the payment in your OpenAI account? The API is not free, but OpenAI gives you a small credit to get started (it was $120 for me since I signed up early, but it expired the next month). Could you check if you have set up the payment correctly in OpenAI, and then test here: https://platform.openai.com/playground?mode=chat

If that works, come back to the plugin and try again. Also, the plugin is currently available in Obsidian Community Plugin, you no longer need to do the manual installation. Let me know if this helps!

TheMany172:
Ah yes, that is it, I have not set up payment - the free amount prior to setting up (at least for me) is $0
So no wonder it wasn't working...

Thanks Logan - will close the issue

![openapi issue](https://github.com/logancyang/obsidian-copilot/assets/110791742/6e165b45-efae-4997-8a2e-ba295ae585ee)

logancyang:
@TheMany172  Glad it helped! This is actually a common issue, I will add a FAQ section in the readme soon to include it. Thanks for reporting!

Also, I'm thinking about adding a completely local and free solution so that there is no need to call OpenAI. Please stay tuned!

TheMany172:
Will do, good luck with it all 🫠

Issue #10, opened 2023-05-12
Embedded CSS conflicts with default styling


luckman212:
> Tested w. Copilot v1.2.1

The following CSS overrides default `p` and `ul` styles causing a problem with their margins:

https://github.com/logancyang/obsidian-copilot/blob/5ababd8e122c5f5748fb1554294e289952f88fa2/styles.css#L163-L172

### Example

<img width="540" alt="screenshot_fgEEFKa6_gh" src="https://github.com/logancyang/obsidian-copilot/assets/1992842/ad19859a-263f-4291-9936-c701a17a98c1">

### I edited the CSS to this, seems to fix it

```css
.message-content pre, .message-content p {
  margin: 0;
  padding: 0;
}

.message-content ol, .message-content ul {
  list-style: none;
  padding-left: 20px;
  margin: 0;
}
```

logancyang:
Thanks for the catch! Posting a fix soon!

logancyang:
Pushed 1.2.2, let me know if there's still an issue. Appreciate the feedback!

PR #2156, opened 2026-02-06
Miyo integration: backend + retriever

wenzhengjiang:
## Summary
- add Miyo service discovery + API client
- implement Miyo index backend + semantic retriever
- wire backend selection, embedding override, and settings toggle
- update Miyo API/search docs

## Testing
- eslint . --ext .js,.jsx,.ts,.tsx

### DIFF
--- a/src/components/chat-components/RelevantNotes.tsx
+++ b/src/components/chat-components/RelevantNotes.tsx
@@ -9,6 +9,7 @@ import { useActiveFile } from "@/hooks/useActiveFile";
 import { cn } from "@/lib/utils";
 import { logWarn } from "@/logger";
 import { SemanticSearchToggleModal } from "@/components/modals/SemanticSearchToggleModal";
+import type { CopilotSettings } from "@/settings/model";
 import {
   findRelevantNotes,
   getSimilarityCategory,
@@ -27,6 +28,30 @@ import {
 } from "lucide-react";
 import { Notice, TFile } from "obsidian";
 import React, { memo, useCallback, useEffect, useState } from "react";
+
+const SELF_HOST_GRACE_PERIOD_MS = 15 * 24 * 60 * 60 * 1000;
+
+/**
+ * Return true when Miyo-backed semantic index is expected to be active.
+ *
+ * @param settings - Current Copilot settings object.
+ * @returns True when Miyo mode and self-host validation are active.
+ */
+function shouldUseMiyoIndex(settings: CopilotSettings): boolean {
+  if (!settings.enableMiyoSearch || !settings.enableSemanticSearchV3) {
+    return false;
+  }
+
+  if (settings.selfHostModeValidatedAt == null) {
+    return false;
+  }
+
+  if ((settings.selfHostValidationCount ?? 0) >= 3) {
+    return true;
+  }
+
+  return Date.now() - settings.selfHostModeValidatedAt < SELF_HOST_GRACE_PERIOD_MS;
+}
 
 function useRelevantNotes(refresher: number) {
   const [relevantNotes, setRelevantNotes] = useState<RelevantNoteEntry[]>([]);
@@ -65,6 +90,16 @@ function useHasIndex(notePath: string, refresher: number) {
     async function fetchHasIndex() {
       try {
         const VectorStoreManager = (await import("@/search/vectorStoreManager")).default;
+        const { getSettings } = await import("@/settings/model");
+        const settings = getSettings();
+        const shouldUseMiyo = shouldUseMiyoIndex(settings);
+
+        if (shouldUseMiyo) {
+          const isEmpty = await VectorStoreManager.getInstance().isIndexEmpty();
+          setHasIndex(!isEmpty);
+          return;
+        }
+
         const has = await VectorStoreManager.getInstance().hasIndex(notePath);
         setHasIndex(has);
       } catch {
--- a/src/miyo/MiyoClient.ts
+++ b/src/miyo/MiyoClient.ts
@@ -177,7 +177,7 @@ export class MiyoClient {
     sourceId: string,
     documents: MiyoUpsertDocument[]
   ): Promise<number> {
-    const payload = { source_id: sourceId, documents };
+    const payload = { vault: sourceId, documents };
     const response = await this.requestJson<{ upserted: number }>(baseUrl, "/v0/index/upsert", {
       method: "POST",
       body: payload,
@@ -194,7 +194,7 @@ export class MiyoClient {
    * @returns Count of documents deleted.
    */
   public async deleteByPath(baseUrl: string, sourceId: string, path: string): Promise<number> {
-    const payload = { source_id: sourceId, path };
+    const payload = { vault: sourceId, path };
     const response = await this.requestJson<{ deleted?: number }>(baseUrl, "/v0/index/by_path", {
       method: "DELETE",
       body: payload,
@@ -209,7 +209,7 @@ export class MiyoClient {
    * @param sourceId - Vault-specific source id.
    */
   public async clearIndex(baseUrl: string, sourceId: string): Promise<void> {
-    const payload = { source_id: sourceId };
+    const payload = { vault: sourceId };
     await this.requestJson(baseUrl, "/v0/index/clear", { method: "POST", body: payload });
   }
 
@@ -230,7 +230,7 @@ export class MiyoClient {
   ): Promise<MiyoIndexedFilesResponse> {
     return this.requestJson<MiyoIndexedFilesResponse>(baseUrl, "/v0/index/files", {
       method: "GET",
-      query: { source_id: sourceId, offset, limit },
+      query: { vault: sourceId, offset, limit },
     });
   }
 
@@ -244,7 +244,7 @@ export class MiyoClient {
   public async getStats(baseUrl: string, sourceId: string): Promise<MiyoIndexStatsResponse> {
     return this.requestJson<MiyoIndexStatsResponse>(baseUrl, "/v0/index/stats", {
       method: "GET",
-      query: { source_id: sourceId },
+      query: { vault: sourceId },
     });
   }
 
@@ -263,7 +263,7 @@ export class MiyoClient {
   ): Promise<MiyoDocumentsResponse> {
     return this.requestJson<MiyoDocumentsResponse>(baseUrl, "/v0/index/documents", {
       method: "GET",
-      query: { source_id: sourceId, path },
+      query: { vault: sourceId, path },
     });
   }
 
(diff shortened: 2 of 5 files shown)


logancyang:
# Code Review: Miyo Integration (dc729fd)

**Commit**: `dc729fd Miyo integration: backend + retriever (#2156)`
**Reviewers**: Claude Opus 4.6 + OpenAI Codex (gpt-5.3-codex)
**Date**: 2026-02-18
**Verdict**: Merge-worthy with follow-up items. No blockers found. Several improvements recommended.

---

## Summary

The commit introduces Miyo as an alternative semantic search backend alongside Orama. It adds:
- `MiyoClient` HTTP client with service discovery
- `MiyoIndexBackend` implementing `SemanticIndexBackend`
- `MiyoSemanticRetriever` extending LangChain's `BaseRetriever`
- `RetrieverFactory` updates for Miyo priority routing
- `VectorStoreManager` dual-backend switching
- `findRelevantNotes` Miyo-backed scoring path
- Settings UI for Miyo Search toggle

Build passes. All existing tests pass. No regressions detected.

---

## Findings

### P1 — Privacy / Correctness

#### 1. Unconditional request logging leaks vault content to file log
**File**: `src/miyo/MiyoClient.ts:377-382`
**Severity**: High
**Status**: Valid

```typescript
logInfo("Miyo request:", {
  method: options.method,
  url: url.toString(),
  hasBody: Boolean(body),
  ...(options.method === "POST" ? { postBody: options.body } : {}),
});
```

`logInfo()` always writes to the rolling file log (`copilot/copilot-log.md`) regardless of the `debug` setting — only the `console.log` call is gated. Every POST body (containing full document content during upsert) is persisted to the vault log file. This is a privacy concern for users who don't expect their note content in log files.

**Fix**: Wrap in `if (getSettings().debug)` or remove `postBody` from the unconditional log.

---

#### 2. Grace period logic duplicated between `findRelevantNotes` and `plusUtils`
**File**: `src/search/findRelevantNotes.ts:15, 22-37`
**Severity**: High (maintenance risk)
**Status**: Valid

`shouldUseMiyoForRelevantNotes()` reimplements the exact same validation logic as `isSelfHostAccessValid()` from `plusUtils.ts`, including its own copy of `SELF_HOST_GRACE_PERIOD_MS`. If the grace period constant, threshold count (3), or validation logic ever changes in one place, the other won't follow, silently breaking Miyo for relevant notes.

**Fix**: Replace `shouldUseMiyoForRelevantNotes()` body with:
```typescript
function shouldUseMiyoForRelevantNotes(): boolean {
  const settings = getSettings();
  return settings.enableMiyoSearch && settings.enableSemanticSearchV3 && isSelfHostAccessValid();
}
```
This is exactly what `RetrieverFactory.shouldUseMiyo()` and `VectorStoreManager.shouldUseMiyo()` already do.

---

#### 3. `fromMiyoDocument` mutates the API response metadata object in place
**File**: `src/search/indexBackend/MiyoIndexBackend.ts:391-397`
**Severity**: Medium
**Status**: Valid

```typescript
const metadata = doc.metadata ?? {};
if (!metadata.chunkId && doc.chunk_index !== undefined) {
  metadata.chunkId = `${doc.path || fallbackPath}#${doc.chunk_index}`;
}
```

`metadata` is a reference to `doc.metadata`. Mutating it changes the original API response object. While the practical impact is low (responses are typically consumed once), this violates the expectation that a conversion function produces a new object.

**Fix**: Spread-copy: `const metadata = { ...(doc.metadata ?? {}) };`

---

### P2 — Code Quality / Dead Code

#### 4. Dead `if (enabled)` branch in Miyo Search toggle handler
**File**: `src/settings/v2/components/CopilotPlusSettings.tsx:47-69`
**Severity**: Low (correctness OK, readability issue)
**Status**: Valid

```typescript
if (!enabled) {
  updateSetting("enableMiyoSearch", false);
  return;                        // ← already returned for !enabled
}

if (enabled) {                   // ← always true here
  setIsValidatingSelfHost(true);
  // ...
}
```

The second `if (enabled)` block is redundant since `!enabled` already returned. The code works correctly but the dead branch adds confusion.

**Fix**: Remove the `if (enabled)` wrapper; the code after the early return is already guarded.

---

#### 5. Verbose per-document upsert logging without debug guard
**File**: `src/search/indexBackend/MiyoIndexBackend.ts:305-311`
**Severity**: Medium (perf + noise)
**Status**: Valid

```typescript
logInfo(`Miyo upsert batch: ${docs.length} documents`, { baseUrl, sourceId });
docs.forEach((doc) => {
  logInfo(`Miyo upsert document ${doc.id} @ ${doc.path}${chunkSuffix}`);
});
```

Every individual document upsert is logged unconditionally. `logInfo` always writes to the rolling log file. For a vault with thousands of notes, a full reindex produces thousands of log lines, potentially causing the log file to churn rapidly and lose older entries.

**Fix**: Wrap per-document logging in `if (getSettings().debug)`.

---

#### 6. Test name describes wrong code path
**File**: `src/search/findRelevantNotes.test.ts:172`
**Severity**: Low (test coverage gap)
**Status**: Valid

Test is named: *"uses Miyo related-note endpoint when source note chunks do not include embeddings"*

With `enableMiyoSearch: true` and valid `selfHostModeValidatedAt`, `shouldUseMiyoForRelevantNotes()` returns `true`, so `calculateSimilarityScore` takes the Miyo-primary path (line 181) and never inspects embeddings at all. The test actually validates the **Miyo-primary** path, not the **no-embeddings fallback** path.

The actual no-embeddings fallback (lines 190-212 of `findRelevantNotes.ts` — where Orama docs exist but have empty embeddings, and the code falls through to Miyo) is **untested**.

**Fix**: Rename the test to reflect what it tests ("uses Miyo when shouldUseMiyoForRelevantNotes is true") and add a separate test for the no-embeddings fallback path with `enableMiyoSearch: false`.

---

### P3 — Platform / Architecture

#### 7. Service discovery hardcoded to macOS paths
**File**: `src/miyo/MiyoServiceDiscovery.ts:94-100`
**Severity**: Medium (platform limitation)
**Status**: Valid (intentional for now)

```typescript
return pathModule.join(
  osModule.homedir(),
  "Library",
  "Application Support",
  "Miyo",
  "service.json"
);
```

The `Library/Application Support` path is macOS-specific. Windows would need `AppData/Local/Miyo` and Linux `~/.config/Miyo` or `$XDG_CONFIG_HOME/Miyo`.

Since Miyo is currently macOS-only, this is acceptable. However, it should be documented or a TODO added for cross-platform support when Miyo ships on other platforms. The `Platform.isDesktopApp` guard at line 57 prevents crashes on mobile but doesn't distinguish between Windows/Linux/macOS desktop.

---

#### 8. `MiyoSemanticRetriever.getExplicitChunks` uses `VectorStoreManager` singleton instead of `this.client`
**File**: `src/search/miyo/MiyoSemanticRetriever.ts:193`
**Severity**: Medium (architectural inconsistency)
**Status**: Valid

```typescript
const docs = await VectorStoreManager.getInstance().getDocumentsByPath(noteFile.path);
```

The retriever has `this.client` (a `MiyoClient`) but fetches explicit chunks through `VectorStoreManager`, which delegates to whichever backend is currently active. When the active backend is Orama (e.g., during a transition or if `shouldUseMiyo` evaluates differently), explicit chunks come from Orama while search results come from Miyo — creating a data source mismatch.

**Fix**: Use `this.client.getDocumentsByPath()` directly, or accept the VectorStoreManager indirection and document the intentional cross-backend behavior.

---

#### 9. Dual auth headers sent on every request
**File**: `src/miyo/MiyoClient.ts:340-347`
**Severity**: Low
**Status**: Valid

```typescript
private buildHeaders(apiKeyOverride?: string): Record<string, string> {
  const apiKey = apiKeyOverride ?? getSettings().selfHostApiKey;
  if (apiKey) {
    headers.Authorization = `Bearer ${apiKey}`;
    headers["X-API-Key"] = apiKey;
  }
  return headers;
}
```

Two issues:
1. The `apiKeyOverride` parameter is never passed by any caller (line 387 calls `this.buildHeaders()` without arguments). It's dead code.
2. Both `Authorization: Bearer` and `X-API-Key` headers are sent simultaneously. This is likely for compatibility with different server configurations, but should be documented.

---

### P4 — Performance (Low Priority)

#### 10. `garbageCollect` makes N sequential HTTP DELETE requests
**File**: `src/search/indexBackend/MiyoIndexBackend.ts:226-228`
**Severity**: Low
**Status**: Valid

```typescript
for (const path of pathsToRemove) {
  deleted += await this.client.deleteByPath(baseUrl, sourceId, path);
}
```

Each stale path triggers a separate HTTP round-trip. For vaults with many renamed/deleted files, this could be slow. Could batch with `Promise.all` (with concurrency limit) or a bulk delete endpoint.

Low priority since garbage collection runs infrequently and typically handles few files.

---

#### 11. `getIndexedFiles` pagination reassigns `total` on every page
**File**: `src/search/indexBackend/MiyoIndexBackend.ts:119`
**Severity**: Low
**Status**: Valid

```typescript
total = response.total ?? paths.size;
```

If the server-side document count changes between pages (e.g., concurrent indexing adds files), the loop termination condition shifts. Reading `total` once from the first response would be more reliable.

---

## Items Reviewed and Dismissed

| Candidate Finding | Reason Dismissed |
|---|---|
| Backend switching race in constructor | `setupSettingsSubscription` runs synchronously after field init; timing window is negligible |
| Missing error boundary in `searchMiyo` | Already has try/catch returning `[]` on failure (line 112-115) |
| `getDocumentKey` fallback chain is fragile | Defensive; worst case is duplicate-but-correct entries |
| Miyo toggle doesn't auto-enable `enableSelfHostMode` | Intentional — Miyo and Self-Host are independent features sharing validation |
| `parseResponseJson` double-parse attempt | Defensive against Obsidian's `requestUrl` which may return pre-parsed JSON or raw text |

---

## Recommendations Summary

| Priority | Finding | Action |
|----------|---------|--------|
| P1 | #1 Unconditional request body logging | Gate behind `debug` or remove body from log |
| P1 | #2 Duplicated grace period logic | Use `isSelfHostAccessValid()` from `plusUtils` |
| P2 | #3 Metadata mutation | Spread-copy in `fromMiyoDocument` |
| P2 | #4 Dead if-branch | Remove redundant `if (enabled)` |
| P2 | #5 Verbose upsert logging | Gate per-doc logging behind `debug` |
| P2 | #6 Misleading test name + gap | Rename test, add fallback path test |
| P3 | #7 macOS-only discovery | Document or add platform switch |
| P3 | #8 Explicit chunks via singleton | Use `this.client` directly |
| Low | #9 Dual auth headers / dead param | Clean up `apiKeyOverride`, document dual headers |
| Low | #10 Sequential GC deletes | Batch or parallelize |
| Low | #11 Pagination total drift | Read total from first response only |

PR #2189, opened 2026-02-16
Fix unwanted indexing + editor lag during indexing

logancyang:
## Summary
- **Guard against unwanted indexing**: Added `enableSemanticSearchV3` check at the top of `indexVaultToVectorStore()` in both `VectorStoreManager` and `IndexOperations` (defense-in-depth). This prevents indexing when semantic search is disabled, fixing the `applyPlusSettings()` bypass in `plusUtils.ts`.
- **Throttle progress updates**: Added `throttledUpdateIndexingCount()` in `aiParams.ts` that limits Jotai atom writes to once per 500ms (down from ~60+ per run), dramatically reducing cascading React re-renders of `Chat.tsx`, `IndexingProgressCard`, and `ChatContextMenu`.
- **Yield main thread**: Added `yieldToMainThread()` calls between embedding batches and chunk preparation batches so the browser can process editor input events during indexing.

## Test plan
- [x] `npm run build` — clean compile
- [x] `npm run test` — all 1642 tests pass
- [x] Manual: Set `enableSemanticSearchV3: false`, trigger indexing via command palette → should be blocked with log warning
- [x] Manual: Index a large vault (500+ notes), type in editor during indexing → should remain responsive
- [x] Manual: Verify progress card still updates visually during indexing (within ~500ms intervals)

### DIFF
--- a/src/aiParams.ts
+++ b/src/aiParams.ts
@@ -347,22 +347,6 @@ export function setIndexingProgressState(state: IndexingProgressState) {
 }
 
 /**
- * Resets the indexing progress state to the default (idle) state.
- * Use when indexing completes with nothing to do (e.g. index already up to date).
- */
-export function resetIndexingProgressState() {
-  settingsStore.set(indexingProgressAtom, {
-    isActive: false,
-    isPaused: false,
-    isCancelled: false,
-    indexedCount: 0,
-    totalFiles: 0,
-    errors: [],
-    completionStatus: "none",
-  });
-}
-
-/**
  * Updates specific fields in the indexing progress state.
  */
 export function updateIndexingProgressState(partial: Partial<IndexingProgressState>) {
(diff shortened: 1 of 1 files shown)


src/search/indexOperations.ts
```diff
@@ -314,11 +324,21 @@ export class IndexOperations {
...
     }
 
     return allChunks;
   }
 
+  /**
+   * Yields to the main thread so the browser can process pending input events and re-renders.
+   */
+  private yieldToMainThread(): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, 0));
```
zeroliu: I don't quite get how this can help with the performance. Can you double check with and without this function, it will make a user-perceived difference?

logancyang: removed

logancyang:
@codex review

logancyang:
@codex review

PR #2191, opened 2026-02-19
fix: address Miyo PR review findings (privacy, dedup, architecture)

wenzhengjiang:
## Summary

Follow-up fixes from the code review of the merged Miyo integration PR (#2156).

- **P1 — Privacy**: Gate POST body from request logs behind `debug` flag — note content was being written to vault log files unconditionally on every upsert
- **P1 — Dedup**: Replace duplicated grace-period validation logic in `findRelevantNotes` with `isSelfHostAccessValid()` from `plusUtils` (single source of truth)
- **P2 — Correctness**: Spread-copy `doc.metadata` in `fromMiyoDocument` instead of mutating the API response object in place
- **P2 — Dead code**: Remove redundant `if (enabled)` wrapper in `CopilotPlusSettings` toggle handler (unreachable after early return)
- **P2 — Logging**: Gate per-document upsert log lines behind `debug` to prevent log churn on full vault reindex (thousands of lines)
- **P2 — Tests**: Rename misleading test ("no embeddings" was actually testing Miyo-primary path); add a proper no-embeddings fallback path test
- **P3 — Platform**: Add TODO comment in `MiyoServiceDiscovery` for Windows/Linux paths when Miyo ships on those platforms
- **P3 — Architecture**: Use `this.client.getDocumentsByPath()` directly in `MiyoSemanticRetriever.getExplicitChunks` instead of `VectorStoreManager` singleton, preventing data-source mismatch if backends differ
- **Low**: Remove unused `apiKeyOverride` dead parameter from `buildHeaders`; document why both `Authorization: Bearer` and `X-API-Key` headers are sent
- **Low**: Read pagination `total` only from first response in `getIndexedFiles` to avoid drift from concurrent indexing

## Test plan

- [x] All 1646 existing tests pass (`npm run test`)
- [x] ESLint clean (`npm run lint`)
- [x] New test: "falls back to Miyo when Orama docs exist but have no embeddings and have content" covers previously untested fallback path
- [x] Renamed test accurately describes Miyo-primary path behavior

### DIFF
--- a/src/search/findRelevantNotes.test.ts
+++ b/src/search/findRelevantNotes.test.ts
@@ -178,7 +178,7 @@ describe("findRelevantNotes", () => {
     expect(mockSearchRelated).not.toHaveBeenCalled();
   });
 
-  it("uses Miyo when shouldUseMiyoForRelevantNotes is true (enableMiyoSearch=true and valid self-host)", async () => {
+  it("uses Miyo when shouldUseMiyoForRelevantNotes is true (enableMiyo=true and valid self-host)", async () => {
     mockedIsSelfHostAccessValid.mockReturnValue(true);
     mockedGetSettings.mockReturnValue({
       debug: false,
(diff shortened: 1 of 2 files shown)


src/search/indexBackend/MiyoIndexBackend.ts
```diff
@@ -303,12 +306,14 @@ export class MiyoIndexBackend implements SemanticIndexBackend {
...
-      const chunkSuffix =
-        typeof chunkId === "string" && chunkId.length > 0 ? ` (chunkId: ${chunkId})` : "";
-      logInfo(`Miyo upsert document ${doc.id} @ ${doc.path}${chunkSuffix}`);
-    });
+    if (getSettings().debug) {
+      docs.forEach((doc) => {
+        const chunkId = doc.metadata?.chunkId;
+        const chunkSuffix =
+          typeof chunkId === "string" && chunkId.length > 0 ? ` (chunkId: ${chunkId})` : "";
+        logInfo(`Miyo upsert document ${doc.id} @ ${doc.path}${chunkSuffix}`);
```
logancyang: logInfo logWarn logError utils are already aware of the debug flag. This is unnecessary. That review was a false positive

wenzhengjiang: Let me add this to AGENTS.md.

logancyang:
@codex review

logancyang:
@wenzhengjiang we should deprecate explicit chunks and use filterRetriever for path reads. In the future the acp agent won't need this either.

PR #2198, opened 2026-02-20
Miyo indexing: reset batch size on enable and apply updates after resume

wenzhengjiang:
## Summary
- reset `embeddingBatchSize` to `DEFAULT_SETTINGS.embeddingBatchSize` when users enable Miyo Search
- update the Miyo enable confirmation copy to explain the reset and that users with strong hardware can increase batch size later in QA settings
- fix pause/resume indexing flow so updated batch-size config is applied after resume
- switch indexing batch iteration to dynamic stepping (`i += batch.length`) to avoid skipped chunks when batch size changes mid-run

## Validation
- npm run format
- npm run lint
- npm run lint -- src/search/indexOperations.ts

### DIFF
--- a/src/search/indexOperations.ts
+++ b/src/search/indexOperations.ts
@@ -475,8 +475,11 @@ export class IndexOperations {
           this.cancelIndexing();
           return;
         }
-        this.state.totalFilesToIndex = files.length;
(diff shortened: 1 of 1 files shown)


wenzhengjiang:
@codex review

PR #2210, opened 2026-02-24
feat: add configurable output folder for converted document markdown

logancyang:
## Summary

- **Configurable output folder for converted docs**: New "Store converted markdown at" setting under Document Processor (Copilot Plus). When set, converted markdown from pdf4llm/docs4llm is saved to the specified vault folder with readable flat naming (`{basename}.md`). Empty = disabled. Suggested path: `copilot/converteddocs` (already excluded from indexing by default via `qaExclusions`).
- **Miyo PDF routing in project mode**: PDFs are now routed through Miyo's `parse-doc` endpoint in project mode when self-host is enabled, not just in non-project mode.
- **Self-host privacy preservation**: When Miyo fails to parse a PDF in self-host mode, we never fall back to the cloud API. In project mode (`Docs4LLMParser`) this throws so `executeWithProcessTracking` correctly marks the file as failed/retriable. In non-project mode (`PDFParser`) it returns an error string to the chat.
- **MiyoParseResult type**: Replaces the previous `string | null` return with a discriminated union (`null | {content} | {error}`) so Miyo error details (empty text, network failure, path issues) propagate to the user instead of generic messages.
- **Content-dedup on write**: `saveConvertedDocOutput` compares existing file content before writing, preventing mtime churn that triggers re-indexing/sync on every project refresh.
- **Cache-hit export**: All cache-hit paths (PDFParser, Docs4LLMParser, ProjectManager) now trigger `saveConvertedDocOutput`, so enabling the setting retroactively works without clearing caches.
- **Collision handling**: Double-underscore (`__`) path separator for disambiguation, with source-header verification and a final guard against overwrites.
- **TODO comments**: Request timeout in MiyoClient, batch progress feedback in ProjectManager.

## Test plan

- [x] All 1678 unit tests pass
- [x] Enable "Store converted markdown at" with `copilot/converteddocs`, process a PDF — verify `.md` file appears with `<!-- source: ... -->` header
- [x] Change the setting path, reprocess — verify output appears in new location
- [x] Leave setting empty — verify no output files created
- [x] Enable self-host + Miyo, process a normal PDF — verify Miyo is used
- [ ] Enable self-host + Miyo, process a failing PDF — verify error thrown (no cloud fallback), file marked as failed in project load tracker
- [x] Disable self-host, process a PDF — verify pdf4llm cloud API is used as before
- [ ] Process two PDFs with same basename from different folders — verify disambiguation naming

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -241,6 +241,17 @@ For detailed architecture diagrams and documentation, see [`MESSAGE_ARCHITECTURE
 - Test files adjacent to implementation (`.test.ts`)
 - Use `@testing-library/react` for component testing
 
+### Avoiding Deep Dependency Chains in Tests
+
+This codebase has deep transitive import chains (e.g. a utility → cache → searchUtils → embeddingManager → brevilabsClient → plusUtils → Modal). Importing any module in this chain from a test requires mocking the entire tree, which is brittle and verbose.
+
+**Rules for new code:**
+
+1. **Pass data, not services** — If a function only needs a string (like `outputFolder`), accept it as a parameter. Don't give it access to the entire settings singleton.
+2. **Singletons at the edges only** — `getSettings()`, `PDFCache.getInstance()`, `BrevilabsClient.getInstance()` should only be called in top-level orchestration (constructors, main entry points). Inner functions receive what they need as parameters.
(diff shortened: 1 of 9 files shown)


logancyang:
@codex review

logancyang:
@codex review

logancyang:
@codex review

Issue #2218, opened 2026-02-25
The mouse selection content is not being updated


hubuque:
The mouse selection content is not being updated; it's still recognizing the previous mouse selection.

![Image](https://github.com/user-attachments/assets/16871ef5-1d26-4650-865c-311dcfcfe8b4)

logancyang:
cc @Emt-lin can you reproduce this? I can't

hubuque:
This bug started appearing from version 3.2.0.

Emt-lin:
> cc [@Emt-lin](https://github.com/Emt-lin) can you reproduce this? I can't

yeah, i can reproduce

hubuque:
希望今天更新一下插件，谢谢。

PR #2222, opened 2026-02-27
fix: exclude non-recoverable segments from L2 promotion to prevent stale selected text

Emt-lin:
## issues
- #2218

### Problem                                                                             
                                                                                        
After #2164 (8cea1a6) made `parseContextIntoSegments` registry-driven and introduced    
per-artifact segment IDs, `selected_text` and `web_selected_text` blocks from previous
turns started leaking into L2 context. This caused stale selected text from earlier
turns to shadow the current turn's fresh selection, leading to the LLM receiving
outdated context.

Unlike `note_context` or `url_content`, selected text is ephemeral — it reflects a
momentary editor selection and cannot be re-fetched. Promoting it to L2 is incorrect.

### Fix

Refactored `compactSegmentForL2()` to use a single-pass `replace()` that matches
top-level XML blocks and decides per-block whether to strip, compact, or pass through:

- **Strip**: non-recoverable blocks (`selected_text`, `web_selected_text`) and
`prior_context_note`
- **Compact**: registered recoverable blocks (e.g. `note_context`, `url_content`)
- **Pass through**: `prior_context` blocks (already compacted) and unknown tags

This replaces the previous "extract-and-reconstruct" approach (from the initial fix)
which silently dropped unknown tags and inter-block text. The single-pass approach also
avoids penetrating into nested content (e.g. a literal `<selected_text>` inside a note's
 `<content>`).

Additionally, tightened `hasCompactedContent` detection from `/<prior_context[\s>]/` to
`/<prior_context\s+source=/`, requiring the `source=` attribute to avoid false-positives
 from literal XML in user note content.

### DIFF
--- a/src/core/ContextManager.l2Promotion.test.ts
+++ b/src/core/ContextManager.l2Promotion.test.ts
@@ -1,0 +1,517 @@
+/**
+ * Tests for L2 promotion filtering in ContextManager.buildL2ContextFromPreviousTurns.
+ *
+ * Verifies that non-recoverable segments (selected_text, web_selected_text) are
+ * excluded from L2 promotion to prevent stale context from previous turns
+ * shadowing the current turn's fresh context.
+ */
+
+import { PromptContextEnvelope, PromptLayerSegment } from "@/context/PromptContextTypes";
+
+// Minimal mocks to avoid deep dependency chains
+jest.mock("@/chainFactory", () => ({
+  ChainType: {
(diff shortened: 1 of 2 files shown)


Emt-lin:
@codex review

Emt-lin:
@codex review

Emt-lin:
@codex review

Emt-lin:
@codex review

PR #2223, opened 2026-02-27
fix: close Radix portaled layers when mobile drawer hides

Emt-lin:
## issues
- [discord: Android: GUI Bug When Opening Chats from History](https://discord.com/channels/1182847556032659546/1470231451604746320)

### Problem                                                                                                                                                                   
                                                                                                                                                                              
On mobile, Radix UI popovers and dropdown menus are portaled to `document.body`. When the user presses the back button, Obsidian hides the drawer by adding the `is-hidden`   
class, but the portaled popover remains visible and jumps to position `(0, 0)` because its anchor element disappears.                                                         

### Solution                                                                                                                                                                  
                                                                                                                                                                              
- Add a `MutationObserver` in `CopilotView` that watches the drawer's class list for the `visible → hidden` transition. On transition, it dispatches an `Escape` keydown event
 so Radix's dismissable-layer closes any open popovers/dropdowns scoped to this view.                                                                                         
- Set `hideWhenDetached={Platform.isMobile}` on `PopoverContent` and `DropdownMenuContent` so Radix natively hides the floating layer when its anchor is no longer visible.
- Clean up the observer on view close.

### DIFF
--- a/src/components/CopilotView.tsx
+++ b/src/components/CopilotView.tsx
@@ -63,6 +63,17 @@ export default class CopilotView extends ItemView {
     this.renderView(handleSaveAsNote, updateUserMessageHistory);
     this.setupMobileKeyboardObserver();
     this.setupDrawerHideObserver();
+
+    // Reason: The view can move between containers (e.g. editor tab → drawer)
+    // without onOpen firing again. Re-bind the drawer observer on layout changes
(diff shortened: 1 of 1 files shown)


Emt-lin:
@codex review

PR #2228, opened 2026-03-02
fix: skip redundant eligibility check when enabling Miyo

wenzhengjiang:
## Summary

- When a user enables Self-Host Mode, `validateSelfHostMode()` is called and eligibility is confirmed before `enableSelfHostMode` is set to `true`
- The Miyo toggle is only rendered when `settings.enableSelfHostMode` is `true`, so by the time a user can toggle Miyo on, they've already passed the eligibility check
- Removed the redundant `validateSelfHostMode()` call from `handleMiyoSearchToggle` — the only check still needed is `isBackendAvailable` to confirm the Miyo app process is running

## Test plan

- [ ] Enable Self-Host Mode → eligibility check runs once (expected)
- [ ] Enable Miyo → no second eligibility check, only checks if Miyo app is running
- [ ] Disabling and re-enabling Miyo does not trigger another API validation call
- [ ] Disabling Self-Host Mode still disables Miyo as expected

### DIFF
--- a/src/settings/v2/components/CopilotPlusSettings.tsx
+++ b/src/settings/v2/components/CopilotPlusSettings.tsx
@@ -59,11 +59,6 @@ export const CopilotPlusSettings: React.FC = () => {
         new Notice("Miyo app is not available. Please start the Miyo app and try again.");
         return;
       }
-
-      const isValid = await validateSelfHostMode();
-      if (!isValid) {
-        return;
-      }
     } finally {
       setIsValidatingSelfHost(false);
     }


wenzhengjiang:
> ### 💡 Codex Review
> https://github.com/logancyang/obsidian-copilot/blob/260554b694a705fd761902d965c9acabb280a8f1/src/settings/v2/components/CopilotPlusSettings.tsx#L56-L58
> 
> 
> **P2 Badge Revalidate self-host state before turning Miyo on**
> This toggle now only checks whether the local Miyo backend is reachable, so users with `enableSelfHostMode=true` but an expired self-host validation window can enable Miyo without refreshing `selfHostModeValidatedAt`; in that state Miyo still won’t actually be used because retrieval paths require `isSelfHostAccessValid()` (`src/search/findRelevantNotes.ts:24`, `src/search/RetrieverFactory.ts:326`). Re-running `validateSelfHostMode()` here was the path that renewed validation when grace had expired.
> 
> ℹ️ About Codex in GitHub

This is acceptable.

PR #2229, opened 2026-03-02
feat: add custom Miyo server URL setting for remote deployments

wenzhengjiang:
## Summary

- Adds a `miyoServerUrl` setting (stored in `CopilotSettings`) that lets advanced users point Copilot at a Miyo instance running on a remote machine
- Exposes a **"Custom Miyo Server URL (Optional)"** text field in the Self-Host Mode settings section, shown before the Enable Miyo toggle — with placeholder `http://127.0.0.1:8742` and a description making clear it is for advanced users only
- When left blank, all existing automatic local service discovery behaviour is unchanged
- Introduces `getMiyoCustomUrl(settings)` helper in `miyoUtils.ts` that safely reads the new field (using `|| ""` to guard against `undefined` on older saved settings), used consistently across all Miyo call sites (`MiyoSemanticRetriever`, `MiyoIndexBackend`, `findRelevantNotes`, `FileParserManager`, `CopilotPlusSettings`)
- Fixes a crash (`TypeError: Cannot read properties of undefined (reading 'trim')`) that occurred when settings loaded from disk lacked the new fields

## Test plan

- [x ] Leave the field blank → Miyo connects via local service discovery as before
- [ x] Enter a remote URL (e.g. `http://192.168.1.10:8742`) → Miyo connects to that address
- [ x] Upgrade from a previous version (existing saved settings without the field) → no crash on Miyo index stats or any Miyo operation

[
Screenshot 2026-03-02 at 11 14 31
](url)


src/settings/model.ts
```diff
@@ -139,6 +139,8 @@ export interface CopilotSettings {
   selfHostUrl: string;
```
logancyang, line 139: should this be removed?

src/settings/model.ts
```diff
@@ -139,6 +139,8 @@ export interface CopilotSettings {
...
   /** API key for the self-host mode backend (if required) */
   selfHostApiKey: string;
```
logancyang, line 141: should we have the option to let them set an API key for self-host as well?

logancyang:
Right now there's no guard for the refunded except on the client side, should we have proper check on the miyo side? @wenzhengjiang

src/settings/model.ts
wenzhengjiang: This is part of SelfHostRetriever which looks like dead code. We can remove it if we don't plan to use SelfHostRetriever anymore, it should be done in a separate PR.

src/settings/model.ts
wenzhengjiang: What is the API key used for? If it's for connecting to remote server safely, we should ask or suggest users to use service like Tailscale.

wenzhengjiang:
> Right now there's no guard for the refunded except on the client side, should we have proper check on the miyo side? @wenzhengjiang

Yeah, I think we can add license check for all external requests (e.g. requests from copilot).

PR #2238, opened 2026-03-03
fix: prevent silent agent loop termination with no output

### LINKED ISSUE #2233: Fix Gemini 3.1 Pro Preview agent loop early-stop (no output)
## Summary
With Google Gemini 3.1 Pro Preview, the agent loop sometimes stops after a few seconds with no output.

## Expected
- Agent loop continues until completion/cancel/error with explicit reason.
- If provider/model incompatibility occurs, show actionable error.

## Actual
- Loop exits early/silently without user-visible output.

## Scope
- Investigate stream/loop handling for Gemini 3.1 Pro Preview.
- Ensure lifecycle/error propagation is surfaced to user.

## Acceptance Criteria
- Repro is fixed, or converted into explicit surfaced error path.
- No silent early-stop.
- Regression test around loop termination/error propagation.

logancyang:
Fixes #2233

## Problem
Gemini 3.1 Pro Preview sometimes returns only thinking/reasoning tokens with no actual response content, causing silent empty responses.

### 1. Symptom Fix (76c7775)
- Detect empty response condition (no content + no tool calls)
- Provide explicit error message instead of silent failure
- Add regression tests

### 2. Root Cause Fix (e6fa8eb)
- Add single instruction to GeminiModelAdapter: "ALWAYS produce substantive response content after reasoning - never return empty"
- Guides Gemini thinking models to ensure output follows reasoning

## Files Changed
- AutonomousAgentChainRunner.ts (+17 lines)
- AutonomousAgentChainRunner.test.ts (+102 lines)  
- modelAdapter.ts (+1 line)

## Testing
- All 1,717 tests pass
- 6 new regression tests for empty response detection

Closes #2233

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,17492 @@
+lockfileVersion: "9.0"
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+  .:
+    dependencies:
+      "@dnd-kit/core":
+        specifier: ^6.3.1
+        version: 6.3.1(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+      "@dnd-kit/sortable":
+        specifier: ^10.0.0
+        version: 10.0.0(@dnd-kit/core@6.3.1(react-dom@18.3.1(react@18.3.1))(react@18.3.1))(react@18.3.1)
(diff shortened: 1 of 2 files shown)


logancyang:
@codex review

logancyang:
@codex review

PR #2247, opened 2026-03-03
fix: prevent silent agent loop termination with Gemini

logancyang:
## Summary

- **Root cause fix**: Gemini's `@langchain/google-genai` nests tool call names inside `functionCall.name` instead of at the top-level `name` property. Our streaming accumulator only checked `tc.name`, so all Gemini tool call names were silently dropped. `buildToolCallsFromChunks` then skipped nameless entries → treated as "no tool calls" → empty response since thinking tokens were filtered.
- **Defense in depth**: Added empty response detection with diagnostic logging (content length, raw chunk count, parsed tool call count) and user-facing error message
- **Tests rewritten**: Replaced local-helper-based tests with 17 tests covering real `accumulateToolCallChunk` and `buildToolCallsFromChunks` code paths for both Gemini-format and OpenAI-format chunks

## Test plan

- [x] `npm run test` — all 1711 tests pass
- [x] `npm run build` — clean build
- [x] Manual: Use Gemini 3.1 Pro Preview in agent mode with a query requiring tool calls (e.g. vault search)
- [x] Manual: Verify OpenAI/Anthropic models still work correctly in agent mode

### DIFF
--- a/src/LLMProviders/chainRunner/AutonomousAgentChainRunner.test.ts
+++ b/src/LLMProviders/chainRunner/AutonomousAgentChainRunner.test.ts
@@ -1,258 +1,262 @@
-import { createToolCallMarker } from "./utils/toolCallParser";
+import {
+  buildToolCallsFromChunks,
+  accumulateToolCallChunk,
+  ToolCallChunk,
+} from "./utils/nativeToolCalling";
 
 /**
- * Test suite for AutonomousAgentChainRunner tool call ID handling
+ * Test suite for Gemini tool call name extraction fix (Issue #2233)
  *
- * This test suite specifically addresses the bug where temporary tool call IDs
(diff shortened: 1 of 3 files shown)


logancyang:
## Testing Note

Gemini 3.1 Pro Preview consistently returns **503 (Service Unavailable)** when tool definitions are included in the request. Non-tool queries work fine, and Gemini 2.5 Pro works with tools. This appears to be a Google-side limitation of the preview model — it doesn't support tool calling via their API yet.

This means the `functionCall.name` extraction fix cannot be verified end-to-end with Gemini 3.1 Pro Preview at this time. However, the fix is still correct — it addresses the LangChain `@langchain/google-genai` chunk format difference (`functionCall.name` vs top-level `name`) that would affect any Gemini model returning tool calls via the native client.

logancyang:
@codex review

PR #2288, opened 2026-03-13
feat: drag relevant notes and sources into editor to insert wikilinks

logancyang:
## Summary

- Adds Obsidian-native drag-and-drop to the Relevant Notes panel (both the expanded card title link and the collapsed badge pills) and the Sources modal source links
- New `useNoteDrag` hook (`src/hooks/useNoteDrag.ts`) wraps `app.dragManager.dragLink` + `onDragStart` - identical to how Smart Connections implements it
- Dropping a dragged note onto the Obsidian editor automatically inserts `[[wikilink]]` with no custom drop handler needed on the editor side
- `app.metadataCache.fileToLinktext` is used to build the correct link text, respecting aliases and path disambiguation

## Test plan

- [x] Open the Relevant Notes panel in the chat sidebar with semantic search enabled
- [x] Drag a note title link (expanded view) onto an open editor - confirm `[[wikilink]]` is inserted at cursor
- [x] Drag a badge pill (collapsed view) onto an open editor - confirm `[[wikilink]]` is inserted
- [x] Open Sources modal on a message that has RAG sources; drag a source link onto editor - confirm insertion
- [x] Confirm hover tooltip shows "- drag to insert wikilink" hint
- [x] Confirm clicking still navigates to the note normally
- [x] Confirm all 1793 unit tests pass: `npm run test`
- [x] Confirm build passes: `npm run build`

### DIFF
--- a/src/hooks/useChatFileDrop.ts
+++ b/src/hooks/useChatFileDrop.ts
@@ -104,6 +104,11 @@ export function useChatFileDrop(props: UseChatFileDropProps): UseChatFileDropRet
       if (e.dataTransfer) {
         e.dataTransfer.dropEffect = "copy";
 
+        // Skip showing drop zone for drags originating from within the plugin (e.g. relevant notes)
+        if (e.dataTransfer.types.includes("copilot/internal-drag")) {
+          return;
+        }
+
         // Check if we have string items (Obsidian nav bar drag) or file items (external files)
         const hasStringItems = Array.from(e.dataTransfer.items).some(
           (item) => item.kind === "string"
--- a/src/hooks/useNoteDrag.ts
+++ b/src/hooks/useNoteDrag.ts
@@ -20,6 +20,9 @@ export function useNoteDrag() {
     const dragManager = (app as any).dragManager;
     if (!dragManager) return;
 
+    // Mark this drag as internal so the chat drop zone overlay doesn't appear
+    e.dataTransfer.setData("copilot/internal-drag", "true");
+
     const linkText = app.metadataCache.fileToLinktext(file, "");
     const dragData = dragManager.dragLink(e.nativeEvent, linkText);
     dragManager.onDragStart(e.nativeEvent, dragData);


src/components/modals/SourcesModal.tsx
```diff
@@ -55,6 +55,19 @@ export class SourcesModal extends Modal {
         href: `obsidian://open?vault=${encodeURIComponent(this.app.vault.getName())}&file=${encodeURIComponent(source.path || source.title)}`,
         text: displayText,
       });
+      link.title = `${displayText} - drag to insert wikilink`;
+      link.draggable = true;
+      link.addEventListener("dragstart", (e) => {
```
zeroliu, line 60: Can we have the agent check whether this can lead to memory leaks? I don't see how the event listeners are removed

logancyang: Investigated this thoroughly. **This is not a real memory leak concern.** Here's why:

**How cleanup works in this modal:**

1. `SourcesModal` extends Obsidian's `Modal` class.
2. `onClose()` (line 165) calls `contentEl.empty()`, which is Obsidian's DOM helper that removes all child elements from the container.
3. Obsidian's `Modal.close()` method also removes the entire `containerEl` from the DOM.

**Why the event listeners don't leak:**

When `contentEl.empty()` runs, it destroys every child DOM node including the `<a>` links that have the `dragstart` (and `click`) listeners attached. Once a DOM element is removed and has no remaining JavaScript references, the garbage collector reclaims both the element and its attached event listeners. This is standard browser behavior per the DOM spec.

There are no closures or external references keeping these link elements alive after `onClose()`. The `source` objects referenced in the listener callbacks are local to `createSourceList` and go out of scope once the modal is closed and the elements are collected.

**Worth noting:** the pre-existing `click` listener on line 58 (present before this PR) uses the exact same pattern with no explicit `removeEventListener`. The new `dragstart` listener is consistent with the established approach. Adding explicit removal would be defensive but unnecessary given the DOM lifecycle here.

**Verdict:** False positive. No memory leak.

PR #2311, opened 2026-03-19
feat(miyo): add customizable vault name setting

wenzhengjiang:
## Summary

- Adds a `miyoVaultName` setting that lets users set a custom name for their vault in Miyo, overriding the auto-detected vault filesystem path
- The Vault Name field appears **before** the Enable Miyo toggle so users can set it up first
- When changed while Miyo is already enabled, clears the old index and triggers a full re-index under the new name; cancelling the confirmation resets the input
- Description explains the purpose, instructs users to set it before enabling Miyo, and notes that remote/multi-device users must use the same name on every device to share the same index

## Test plan

- [ ] With Miyo disabled: set a vault name → clicks Apply → saves silently, no modal
- [ ] With Miyo enabled: change vault name → Apply → confirmation modal appears with old source ID → confirm → old index cleared, new name saved, re-index triggered
- [ ] Cancel the confirmation → input resets to the previously saved value
- [ ] Leave vault name blank → `getMiyoSourceId` falls back to vault path as before
- [ ] Set vault name, disable + re-enable Miyo → indexing uses the custom name as source ID
- [ ] Remote setup: use the same vault name on two devices → both index under the same source ID in Miyo

### DIFF
--- a/src/settings/v2/components/CopilotPlusSettings.tsx
+++ b/src/settings/v2/components/CopilotPlusSettings.tsx
@@ -106,7 +106,6 @@ export const CopilotPlusSettings: React.FC = () => {
    */
   const handleVaultNameApply = async () => {
     const newName = pendingVaultName.trim();
-    if (!newName) return;
     const oldName = (settings.miyoVaultName || "").trim();
     if (newName === oldName) return;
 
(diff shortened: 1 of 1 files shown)


wenzhengjiang:
@codex review

wenzhengjiang:
@codex review

PR #2320, opened 2026-03-21
fix: use /responses for GitHub Copilot codex models

Keryer:
## Summary                                                                                                                            
                                                                                                                                        
  Fix GitHub Copilot-authenticated Codex models (for example `gpt-5.3-codex`) failing with HTTP 400 because they were being sent to `/chat/completions`.                                                                                                                    
                                                                                                                                        
  GitHub Copilot Codex models should use the `/responses` endpoint instead. This PR adds provider-side routing so Copilot Codex models  
  use Responses API, while other Copilot chat-compatible models continue using Chat Completions.                                        
                                                                                                                                        
  ## Root Cause                                                                                                                         
                                                                                                                                        
  The GitHub Copilot chat path was hardwired to `ChatOpenAICompletions`, which always targets `/chat/completions`.                      
                                                                                                                                        
  That works for normal Copilot chat models, but fails for Codex-family models that only support `/responses`.                          
                                                                                                                                        
  ## Changes                                                                                                                            
                                                                                                                                        
  - add a dedicated `GitHubCopilotResponsesModel` for Copilot models that must call `/responses`                                        
  - extract shared GitHub Copilot authenticated fetch logic so both API paths reuse the same token/header handling                      
  - add a reusable routing helper to detect when a GitHub Copilot model should use Responses API                                        
  - wire that routing into both runtime model creation and model ping/verification                                                      
  - keep existing `GitHubCopilotChatModel` behavior for models that still support `/chat/completions`                                   
  - add unit coverage for the routing logic                                                                                             
                                                                                                                                        
  ## Behavior After This PR                                                                                                             
                                                                                                                                        
  - `gpt-5.3-codex` with GitHub Copilot auth no longer hits `/chat/completions`                                                         
  - Copilot Codex models use `/responses`                                                                                               
  - chat runtime and connection testing use consistent endpoint selection                                                               
                                                                                                                                        
  ## Testing                                                                                                                            
                                                                                                                                        
  - added unit tests for GitHub Copilot Responses API routing logic                                                                     
  - manual validation target:                                                                                                           
    - authenticate with GitHub Copilot                                                                                                  
    - select `gpt-5.3-codex`
    - send a simple prompt                                                                                                              
    - confirm the request no longer fails with 400 from `/chat/completions`                                                             
                                                                                                                                        
  ## Notes                                                                                                                              
                                                                                                                                        
  This change is intentionally provider-scoped and generalizable:                                                                       
  it avoids hardcoding a single model at the call site and instead routes GitHub Copilot Codex-family models through the correct API.

### DIFF
--- a/src/LLMProviders/chatModelManager.ts
+++ b/src/LLMProviders/chatModelManager.ts
@@ -814,6 +814,7 @@ export default class ChatModelManager {
 
     // For GPT-5 models, automatically use Responses API for proper verbosity support
     const constructorConfig: any = { ...modelConfig };
+    const useCopilotResponses = shouldUseGitHubCopilotResponsesApi(model);
     if (
       modelInfo.isGPT5 &&
       (selectedModel.vendor === ChatModelProviders.OPENAI ||
@@ -823,7 +824,7 @@ export default class ChatModelManager {
       logInfo(`Enabling Responses API for GPT-5 model: ${model.name} (${selectedModel.vendor})`);
     }
 
-    if (shouldUseGitHubCopilotResponsesApi(model)) {
+    if (useCopilotResponses) {
       constructorConfig.useResponsesApi = true;
       logInfo(`Enabling Responses API for GitHub Copilot model: ${model.name}`);
     }
@@ -836,7 +837,7 @@ export default class ChatModelManager {
       return lmStudioInstance;
     }
 
-    if (shouldUseGitHubCopilotResponsesApi(model)) {
+    if (useCopilotResponses) {
       return new GitHubCopilotResponsesModel(constructorConfig);
     }
 
@@ -900,6 +901,7 @@ export default class ChatModelManager {
         ...pingConfig,
         ...tokenConfig,
       };
+      const useCopilotResponses = shouldUseGitHubCopilotResponsesApi(model);
 
       if (
         modelInfo.isGPT5 &&
@@ -909,7 +911,7 @@ export default class ChatModelManager {
         constructorConfig.useResponsesApi = true;
       }
 
-      if (shouldUseGitHubCopilotResponsesApi(model)) {
+      if (useCopilotResponses) {
         constructorConfig.useResponsesApi = true;
       }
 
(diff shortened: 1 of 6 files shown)


Emt-lin:
@Keryer 

Nice work! CI passes and the routing approach is sound. Two suggestions before merging:

### 1. Deduplicate `shouldUseGitHubCopilotResponsesApi` calls in `chatModelManager.ts`

The helper is called 3 times per flow (`createModelInstance` and `pingModel`). This creates drift risk — if one call site is updated, the others can
 easily be missed.

```typescript
const useCopilotResponses = shouldUseGitHubCopilotResponsesApi(model);

if (useCopilotResponses) {
  constructorConfig.useResponsesApi = true;
  logInfo(`Enabling Responses API for GitHub Copilot model: ${model.name}`);
}

// later:
if (useCopilotResponses) {
  return new GitHubCopilotResponsesModel(constructorConfig);
}
```

Same pattern for pingModel().


### 2. Preserve key // Reason: comments in GitHubCopilotChatModel.ts

The PR removes several comments that explain non-obvious SDK/protocol constraints — these aren't restating the code, they document why workarounds
exist. Per our conventions ("Comment non-obvious code"), I'd suggest keeping at least:

- 401 vs 403 retry: only retry on 401 (expired token), not 403 (no subscription — permanent)
- typeof Request guard: some Obsidian mobile runtimes lack the Request global
- streamUsage: false: Copilot API may not support stream_options.include_usage
- In-place delta.content mutation: spread would lose non-enumerable properties like tool_calls

The new JSDoc @param/@returns tags are great — they complement rather than replace these rationale comments.

Keryer:
@Emt-lin 
Thanks for the thoughtful suggestions. I addressed both points in 39dba5e8933b575b6a0d8b7565050e3e6a7750cf.

I deduplicated the `shouldUseGitHubCopilotResponsesApi(model)` checks in `chatModelManager.ts`, and I also restored the key rationale comments in `GitHubCopilotChatModel.ts` for the non-obvious SDK/protocol constraints you called out.

Emt-lin:
@Keryer Thanks for your PR.

Ready to merge  cc @logancyang

logancyang:
Thanks @Keryer, and thanks @Emt-lin for the review!

PR #2326, opened 2026-03-26
feat(miyo): rename vault name to remote vault path and align with server API

wenzhengjiang:
## Summary

- Renames the `miyoVaultName` setting to `miyoRemoteVaultPath` with automatic migration from the old key on first load
- Updates the settings UI: **"Vault Name"** → **"Remote Vault Path (Optional)"** and **"Custom Miyo Server URL"** → **"Remote Miyo Server URL (Optional)"**
- Both fields now default to blank with no placeholder text — empty means local/auto-detected, non-empty overrides
- Remote Vault Path is always visible (no longer gated by server URL being set)
- Adds a bot name indicator under **Enable Miyo** showing the effective vault path and whether it resolves as local or remote
- Renames `sourceId` parameters in `MiyoClient` public methods to `vault` to match the actual `vault` field the Miyo server expects, eliminating the confusing `vault: sourceId` mapping
- Updates `searchRelated` options from `{ sourceId }` to `{ vault }` and fixes the call site in `findRelevantNotes.ts`

## Test plan

- [ ] Verify existing `miyoVaultName` values migrate correctly to `miyoRemoteVaultPath` on settings load
- [ ] Confirm both fields appear blank by default with no placeholder
- [ ] Confirm Remote Vault Path field is always visible under Self-Host Mode (not gated by server URL)
- [ ] Verify bot name indicator appears under Enable Miyo when toggled on, showing correct path and (local)/(remote) label
- [ ] Confirm local Miyo usage works with both fields empty
- [ ] Confirm remote Miyo usage works when Remote Miyo Server URL and Remote Vault Path are set

### DIFF
--- a/src/search/findRelevantNotes.test.ts
+++ b/src/search/findRelevantNotes.test.ts
@@ -221,7 +221,7 @@ describe("findRelevantNotes", () => {
     expect(mockGetDocumentsByPath).not.toHaveBeenCalled();
     expect(mockSearchRelated).toHaveBeenCalledTimes(1);
     expect(mockSearchRelated).toHaveBeenCalledWith("http://127.0.0.1:8742", "source.md", {
-      sourceId: "test-source",
+      vault: "test-source",
       limit: 20,
     });
   });


wenzhengjiang:
@codex review

PR #2328, opened 2026-03-26
feat(miyo): disable miyo on mobile without a remote server URL

wenzhengjiang:
## Summary

- Adds `shouldUseMiyo()` to `miyoUtils.ts` as the single source of truth for the Miyo activation check, replacing three duplicated copies in `RetrieverFactory`, `findRelevantNotes`, and `VectorStoreManager`
- On mobile, `shouldUseMiyo()` returns `false` unless an explicit remote server URL (`miyoServerUrl`) is configured — local service discovery is unavailable on mobile, so the `enableMiyo` setting is silently ignored in that case
- Drops the redundant `enableSemanticSearchV3` guard: the UI already enforces that enabling Miyo enables semantic search and disabling semantic search disables Miyo

## Test plan

- [ ] On desktop with Miyo enabled: verify Miyo search still activates normally
- [ ] On mobile with Miyo enabled and no `miyoServerUrl`: verify Miyo is not used (falls back to Orama/lexical)
- [ ] On mobile with Miyo enabled and a valid `miyoServerUrl`: verify Miyo activates correctly
- [ ] Confirm no TypeScript or lint errors (`npm run build`)

### DIFF
--- a/src/search/findRelevantNotes.test.ts
+++ b/src/search/findRelevantNotes.test.ts
@@ -2,9 +2,8 @@ import { TFile } from "obsidian";
 import { getBacklinkedNotes, getLinkedNotes } from "@/noteUtils";
 import { findRelevantNotes } from "@/search/findRelevantNotes";
 import { MiyoClient } from "@/miyo/MiyoClient";
-import { getMiyoVault } from "@/miyo/miyoUtils";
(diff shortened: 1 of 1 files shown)


wenzhengjiang:
@codex review

PR #2343, opened 2026-04-06
fix: prevent </div> from being consumed by indented code blocks in think sections

trulyshelton:
Bug: When thinking content ends with a 4-space-indented line (e.g. `google/gemma-4-31b-it`'s bullet-point reasoning), markdown's indented code block rule would consume the immediately-following `</div>` closing tag and render it as literal `</div>"` text. 

Fixed by always trimming section content and appending \n before the closing tag so it lands on its own unindented line.

Also adds tests covering non-streaming, streaming-complete, and streaming-unclosed think block paths.

Bug (before):
Screenshot 2026-04-05 at 11 23 52 PM

Fix (after):
Screenshot 2026-04-05 at 11 26 46 PM


logancyang:
Thanks for the fix and the tests! Merging.

PR #2358, opened 2026-04-15
docs(ai): unify AGENTS.md and CLAUDE.md

capyBearista:
This PR simply consolidates the `AGENTS.md` and `CLAUDE.md` files.

I noticed both were not being kept in synced, so I moved the contents of `CLAUDE.md` (seemingly the most up-to-date as it mentions Obsidian CLI and various other rules while the other doesn't) to `AGENTS.md`. Now, `CLAUDE.md` points to the `AGENTS.md` using an `@` reference.

Hopefully this will ensure harnesses across all contributors always have only one (main) source-of-truth to draw from. (I believe, for example, OpenCode will ignore `CLAUDE.md` if it finds `AGENTS.md`, meaning the more recent info will be ignored).


logancyang:
Thanks for catching the drift! Consolidating to AGENTS.md aligns with where things are going. Merging.

PR #2363, opened 2026-04-21
feat(provider): add MiniMax-M2.7 (Global + CN)

kapelame:
Adds MiniMax as a built-in OpenAI-compatible provider with `MiniMax-M2.7` selectable from the model picker. Mirrors the SiliconFlow pattern.

Two regional providers since URLs and API keys differ:
- **MiniMax** — `api.minimax.io`, key from [platform.minimax.io](https://platform.minimax.io)
- **MiniMax (CN)** — `api.minimaxi.com`, key from [platform.minimaxi.com](https://platform.minimaxi.com)

`format`, `lint`, `test` (1904/1904) all pass.


logancyang:
Thanks for the contribution! We're planning to move provider integrations to a unified gateway approach soon, which may make individual provider PRs unnecessary. Putting this on hold until that direction settles. Will revisit then.

PR #2367, opened 2026-04-26
fix: remove 100ms IME timeout that delayed Enter key for Korean/CJK input

### LINKED ISSUE #2354: Enter key requires longer press in chat window compared to other Obsidian windows
## Bug Description
The Enter key in the Copilot chat window requires a noticeably longer 
press to trigger compared to other Obsidian windows or panels.

## Steps to Reproduce
1. Open Obsidian Copilot chat window
2. Type any message
3. Press Enter key briefly (same as used in other windows)
4. Message does not send — requires a longer key press to trigger

## Expected Behavior
Enter key should respond immediately with a brief press,
consistent with other Obsidian windows.

## Actual Behavior
Enter key requires a longer/harder press to send the message.
Other Obsidian windows respond normally with the same key press.

## Environment
- OS: Windows11
- Obsidian version: 1.12.7
- Obsidian Copilot version: 3.2.7

## Additional Notes
- Keyboard filter key settings (OS level) are not the cause, as other windows work fine.
- Send button click works without any issue.
- Enter key send option is configured in plugin settings.
- Main issue occurs when typing in Korean (Hangul IME) Enter key requires a noticeably longer press to send
- English input also feels slightly slower compared to other chat dialogs, though less noticeable
- Suspected cause: IME composing event interfering with Enter key send event (isComposing flag may not be handled)
- This results in delayed Enter key response for both Korean and English, with Korean being significantly more affected

octo-patch:
Fixes #2354

## Problem

When typing in Korean (Hangul) IME, pressing Enter to send a message required a noticeably longer press than expected. The root cause was a 100ms timeout introduced in the `KeyboardPlugin` component's `compositionend` handler:

```ts
// Old code — 100ms delay before isComposingRef resets
compositionTimeoutRef.current = setTimeout(() => {
  isComposingRef.current = false;
}, 100);
```

Because `isComposingRef.current` stayed `true` for 100ms after composition ended, any Enter key pressed within that window was silently blocked. Users had to either hold Enter (waiting for key-repeat to fire after the timeout) or press Enter twice with a deliberate pause — both feel like "the key requires a longer press".

## Solution

Remove the ref-based composition tracking entirely and rely directly on `event.isComposing || event.keyCode === 229`.

In Chromium/Electron (Obsidian's runtime), the `keydown` event for the Enter that confirms the final IME character fires **with `isComposing: true`** — the same composition flag the browser exposes. This means the explicit `compositionstart`/`compositionend` ref is redundant: the flag on the event itself is already reliable.

This approach is consistent with `follow-up-input.tsx` in the same codebase, which has always used the simpler `nativeEvent.isComposing || nativeEvent.keyCode === 229` check without any timeout.

## Testing

- All 16 existing `KeyboardPlugin` unit tests pass (`npm test -- --testPathPattern="KeyboardPlugin"`).
- Manual verification: after this change, a single normal-speed Enter press sends the message immediately after Korean IME composition ends, with no perceptible delay.

### DIFF
--- a/src/components/chat-components/plugins/KeyboardPlugin.tsx
+++ b/src/components/chat-components/plugins/KeyboardPlugin.tsx
@@ -1,4 +1,4 @@
-import { useRef, useEffect } from "react";
+import { useEffect } from "react";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { COMMAND_PRIORITY_LOW, KEY_ENTER_COMMAND } from "lexical";
 import { SEND_SHORTCUT } from "@/constants";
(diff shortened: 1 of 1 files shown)


logancyang:
@codex review

logancyang:
Thanks for the fix! Verified, looks good. Merging.

PR #2368, opened 2026-04-26
New Agent Mode

zeroliu:


logancyang:
The `setMaxListeners` shim in `25b960a8` didn't fully fix the throw for me — Claude Agent SDK still fails on `newSession`:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "eventTargets" argument must be an instance of EventEmitter or EventTarget. Received an instance of AbortSignal
    at EventEmitter.setMaxListeners (node:events:331:17)
    at Q0 (plugin:copilot:252902:44)
    ...
    at probeClaudeSdkCatalog (plugin:copilot:267911:17)
    at _ClaudeSdkBackendProcess.ensureModelCatalog (plugin:copilot:268260:26)
    at _ClaudeSdkBackendProcess.newSession (plugin:copilot:268003:32)
```

The top frame is `EventEmitter.setMaxListeners` (the static-on-class form), but the shim only replaces `events.setMaxListeners` (the module-level export). In Node those two reference the same underlying function but live in separate property slots — patching one doesn't affect the other. The SDK's call site reaches the unpatched class-static binding.

Patching both fixes it. Verified locally:

```diff
diff --git a/src/utils/rendererEventsShim.ts b/src/utils/rendererEventsShim.ts
@@ -2,7 +2,10 @@ import * as nodeEvents from "node:events";

 type SetMaxListenersFn = (n?: number, ...targets: unknown[]) => void;
 type MarkedFn = SetMaxListenersFn & { [APPLIED]?: boolean };
-type EventsModuleShape = { setMaxListeners: MarkedFn };
+type EventsModuleShape = {
+  setMaxListeners: MarkedFn;
+  EventEmitter: { setMaxListeners: MarkedFn };
+};

@@ -49,6 +57,9 @@ function installShim(): void {
   wrapped[APPLIED] = true;

   eventsModule.setMaxListeners = wrapped;
+  if (eventsModule.EventEmitter) {
+    eventsModule.EventEmitter.setMaxListeners = wrapped;
+  }
 }
```

Worth updating the JSDoc too to note both bindings need to be patched.

logancyang:
Pulled this branch fresh, ran `npm install` + `npm run dev`, hit `Uncaught ReferenceError: React is not defined` in `SvgLogo` as soon as the chat view mounts.

Cause: `esbuild.config.mjs` passes `jsxRuntime: "automatic"` to the inline SVGR plugin, so svgr does not emit an `import * as React from 'react'` at the top of each generated component. But `tsconfig.json` has `"jsx": "react"` (classic) and esbuild has no `jsx: "automatic"` override, so esbuild compiles the resulting JSX to `React.createElement(...)` and blows up because `React` is not in scope.

One-line fix: set `jsxRuntime: "classic"` in the svgr options (or set esbuild `jsx: "automatic"` globally, but that is a bigger change).

PR #2371, opened 2026-04-29
docs: remove second Table of Contents from README

tredondo:
The README had two ToCs, the second one being under the Usage heading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,35 +101,6 @@ Read the [Changelog](https://github.com/logancyang/obsidian-copilot/releases/tag
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


logancyang (changes requested):
Thanks for the cleanup! One thought before we merge: I think the second TOC is actually the more useful one. It has all the Usage subsections like Chat Mode, Vault QA, and Agent Mode. Could we keep that one and drop the shorter first TOC instead?

logancyang:
Closing as it's addressed in #2409 . Thanks again for the PR!

PR #2385, opened 2026-05-06
Fix Azure embedding Base URL handling and validation

tgmof:
# Problem

  - When adding an Azure embedding model via “Embedding models → + Add Model”:
      - Setting a Base URL was ignored; code still built URLs using:
          - azureOpenAIApiInstanceName
          - azureOpenAIApiEmbeddingDeploymentName
          - azureOpenAIApiVersion
      - Runtime ended up calling:
          - POST https://[instance].openai.azure.com/openai/deployments/[deployment]/embeddings?api-version=[version]
            even when a custom Base URL was supplied.

 # Goals

  1. Make Azure embedding models honor the Base URL field, same as chat models.
  2. Allow users to configure Azure embeddings either via:
      - Full Base URL, or
      - Legacy instance / deployment / api-version fields, when Base URL is blank.
  3. Ensure “Test” works without requiring legacy fields when Base URL is provided.

 # Behavior After Changes

  1. Azure Embeddings with Base URL
      - User:
          - Selects provider: Azure.
          - Fills Base URL (full endpoint or base).
          - Sets API key.
      - UI:
          - Hides instance / deployment / version fields.
          - “Test” button no longer complains about azureOpenAIApiInstanceName.
      - Runtime:
          - Uses OpenAIEmbeddings with configuration.baseURL set to Azure endpoint.
          - Adds api-version via defaultQuery.
          - No dependency on Azure-specific constructor fields.
  1. Azure Embeddings with Legacy Fields
      - User:
          - Leaves Base URL empty.
          - Fills Instance Name, Embedding Deployment Name, API Version.
      - UI:
          - Requires these fields and shows inline errors if missing.
      - Runtime:
          - Builds https://[instance].openai.azure.com/openai/deployments/[deployment].
          - Uses OpenAIEmbeddings with the constructed URL + api-version.

### DIFF
--- a/src/LLMProviders/embeddingManager.ts
+++ b/src/LLMProviders/embeddingManager.ts
@@ -9,11 +9,12 @@ import { CohereEmbeddings } from "@langchain/cohere";
 import { Embeddings } from "@langchain/core/embeddings";
 import { GoogleGenerativeAIEmbeddings } from "@langchain/google-genai";
 import { OllamaEmbeddings } from "@langchain/ollama";
-import { AzureOpenAIEmbeddings, OpenAIEmbeddings } from "@langchain/openai";
+import { OpenAIEmbeddings } from "@langchain/openai";
 import { Notice } from "obsidian";
 import { BrevilabsClient } from "./brevilabsClient";
 import { CustomJinaEmbeddings } from "./CustomJinaEmbeddings";
 import { CustomOpenAIEmbeddings } from "./CustomOpenAIEmbeddings";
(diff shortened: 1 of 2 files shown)


logancyang:
Thanks for the fix! The Codex review flagged two real issues worth addressing before merge: the Azure path needs the `api-key` header set in `configuration.defaultHeaders` (matches what chat does), and we need a fallback `api-version` when the Base URL has no query params. Once those are in we can take another look.

PR #2394, opened 2026-05-12
feat(composer): replace Apply view diff with @pierre/diffs

caezium:
## Summary

Swap the bespoke blocks-based diff in the Apply view for [`@pierre/diffs`](https://github.com/pierre-co/diffs) (Shiki-backed), themed against Obsidian's CSS variables so it inherits the user's editor look.

The previous renderer rebuilt its own line/word diff UI from scratch (~600 lines of `SideBySideBlock` / `SplitBlock` / per-block accept-reject state). Pierre handles all of that — including markdown-aware syntax highlighting via Shiki — and the result reads as native Obsidian text rather than a separate pierre-dark pane.

## Before / After

| Before | After |
| --- | --- |
| before-1 | after-1 |
| before-2 | after-2 |
| before-3 | after-3 |
| before-4 | after-4 |
| no split view |  after-5 |

## What changed

**New renderer (`src/components/composer/PierreRenderer.tsx`):**
- Wraps Pierre's `PatchDiff` with `createPatch` + a custom Shiki theme (`"obsidian"`) registered once at module load.
- The theme maps every syntax token (`token-keyword`, `token-string`, `token-link`, …) to an Obsidian text variable (`--text-normal`, `--text-accent`, `--text-muted`), so code blocks, headings, links, and bold inside the diff render in the user's editor colors — not Pierre's stock red/orange/cyan.
- `disableWorkerPool` (Obsidian's renderer process is single-Electron-window, no need to spawn Shiki workers).
- `overflow: "wrap"` so long prose lines wrap instead of horizontally truncating in side-by-side.

**CSS theming (`src/styles/tailwind.css → .copilot-pierre-view`):**
- Overrides `--diffs-{addition,deletion,modified}-color-override` to `--text-normal` — changed-line *text* stays the normal Obsidian color; the diff signal comes from background wash + inline emphasis spans only.
- Comprehensive `--diffs-*` overrides for backgrounds (`--background-modifier-success-rgb` / `error-rgb` at 0.22 alpha for line wash, 0.55 for inline emphasis), gutters, line numbers, hover, selection, font family, line-height, and decoration bar.
- Container border + header strip themed to `--background-secondary`.

**`ApplyView.tsx` simplification:**
- Drops `~600 lines` of `SideBySideBlock` / `SplitBlock` / per-block accept-reject. Pierre doesn't model per-block decisions, so accept writes `newText` verbatim — same UX as the previous "accept all" path.
- The existing `viewMode` toggle (split / side-by-side) now drives Pierre's `diffStyle` prop and persists via the `diffViewMode` setting.
- Floating Accept/Reject pill is centered within the pane (full-width absolute row + `tw-flex tw-justify-center`, pointer-events split so empty space doesn't swallow clicks on the diff). The previous `tw-fixed + tw-left-1/2` anchored to the viewport and put the pill's *left edge* — not its center — at the 50% line.

**Misc:**
- Drop the dead `ApplyViewState.simple` prop (set by `editFile` but never read by any consumer).
- Add a `__mocks__/pierre-diffs-react.js` stub + two `moduleNameMapper` entries in `jest.config.js` — Pierre uses ESM subpath exports that `ts-jest`'s resolver doesn't follow, and tests touching the `ApplyView → PierreRenderer` import chain (e.g. `ComposerTools.test.ts`) would otherwise crash at module load. Tests don't assert on Pierre rendering, so the stub returns an empty `<div>`.

## Tradeoffs

- **Bundle size.** 
- **Per-block accept/reject is gone.** Pierre is a code-review-style renderer, not an interactive merge UI. Accept writes the full proposed text. In practice the prior per-block UI in the composer was rarely used — the model typically proposes one cohesive change and the user accepts or rejects the whole thing.


zeroliu:
This is so amazing. Huge upgrade from our bespoke UI. Using the design tokens is a huge win! The code is well organized. I will consider using it to help with the v4 diff view as well. @logancyang can we assign agent review and consider merging it when the review looks good? I manually went through the change and it LGTM.

Fantastic work @caezium! 🥳

logancyang:
@caezium Thanks for the PR! Great stuff! 

I've tried a bit and seems the pure additions are not highlighted green, only the edits are red/green. Is this something we can customize? For example, these line additions in the first image are hard to see since only the line number is green but not the text.
SCR-20260512-krma

SCR-20260512-krpt

wenzhengjiang:
@caezium Thank you! LGTM.

@logancyang I can see the highlights for pure additions, but I agree they should be more visible. @caezium can you make them stand out more?

Screenshot 2026-05-13 at 9 53 24

caezium:
https://diffs.com/ has more info on this

I pinned the Accept/Reject bar so it stays put when you scroll, tidy up the header, and changed pure additions and deletions colors.
Also forgot to push this but just did: collapsible per-line accept/reject if you want to override individual lines before applying. One click to accept the whole thing, one more click to pick lines.

To tweak the diff look later: all the colors live in src/styles/tailwind.css under .copilot-pierre-view as CSS variable overrides (--diffs-bg-addition-override, --diffs-bg-deletion-override, text colors, gutter colors, etc.).

Renders across light/dark mode and a few different themes:

<p>    </p> <p>    </p> <p>    </p>

idk why in light mode copilot looks like this, maybe custom css on my setup

wenzhengjiang (approved):
LGTM!  @caezium Can you resolve the merge conflicts?

logancyang:
@caezium looks great! BTW I've updated the merge requirements to enforce signed commits for security reasons. Could you please follow the CONTRIBUTING.md instructions and set it up if you haven't already? Also please resolve the merge conflicts and I will merge. Brilliant work!

EL-File4138:
Would like to chime in to point out that due to transitively pulling in Shiki’s full bundled language registry plus inlined Oniguruma WASM as a dependency, this change causes the plugin to be massively bloated by ~10 MB - a 3x increase over the whole functionality of Copilot for only a better edit preview experience. Would definitely recommend considering swapping out the rendering backend to avoid such unnecessary bloat.

caezium:
@logancyang Rebased on master and re-signed all 4 commits with SSH.

caezium:
@EL-File4138  @pierre/diffs/dist/highlighter/languages/resolveLanguage.js doing a static import { bundledLanguages } from "shiki", which makes esbuild pull in the entire @shikijs/langs tree (~9.8MB on its own). Fix is pre-registering markdown via Pierre's registerCustomLanguage. Should drop main.js from ~15MB to ~3–5MB

caezium:
and also: to anyone wanting to push the diff UI further: https://diffs.com/ has live demos of features this PR is not using yet, the package exposes a lot more than what's wired up here. Per-line accept/reject (hover-revealed), hunk-level controls, custom gutter buttons, scroll-sync between split panes - are all available, you can choose how you want it to look like

caezium:
@logancyang

logancyang:
@caezium thanks for following up on this. We are heads down on V4 preview right now and the Copilot plugin is going to be revamped. We will revisit this PR once v4 is ready. Thanks again for the contribution!

PR #2538, opened 2026-05-31
Steer agents toward Copilot Plus builtin skills (with fallback)

logancyang:
## Summary

Companion to #2537 (builtin Copilot Plus skills): a system-prompt nudge so agents actually reach for those skills instead of their own built-in web/fetch tools. This is "Option 2" — steering, not forcing (the agents' native tools stay available as a fallback).

Adds a Plus-only block to the shared agent system prompt (`agentSystemPrompt.ts`):

```
## Copilot Plus tools
For these requests, prefer the bundled Copilot skill over any built-in tool of your own:
- Searching the web → copilot-web-search
- Reading a PDF file → copilot-read-pdf
- Getting a YouTube video's transcript → copilot-youtube-transcript
- Fetching an X (Twitter) post → copilot-fetch-x

If the matching skill is missing, disabled, or reports that it needs an active
Copilot Plus license, fall back to whatever equivalent capability you have (or
tell the user it's unavailable) — don't refuse the request.
```

## Behavior

- **Plus users** get the steering → agents prefer the Copilot skills.
- **Non-Plus users** get nothing (gated on `isPlusUser`) → agents use their own tools directly. The skills can't run without a license, so steering a non-Plus agent would only cause a failed attempt + an upgrade nag on every web search.
- **Suppressed** with the rest of the builtin framing when "Disable builtin system prompt" is on.
- Backends already restart on Plus sign-in/out (from #2537's wiring), so this refreshes on the next session without a reload.

## Why a separate PR (and why merge order doesn't matter)

It touches only `agentSystemPrompt.ts`, which #2537 doesn't, so there's no conflict with #2537's review cycle. The **fallback clause makes it safe to merge in either order**: if this lands before the skills exist, a Plus agent told to "prefer copilot-web-search" simply finds no such skill and falls back to its own tools.

## Caveat (expectation-setting)

Steering raises the hit-rate but is **not deterministic** — skill selection is the agent model's call. Truly guaranteeing the Copilot path would require denying the agents' native web/fetch tools, which is a separate, heavier change (and a product tradeoff). This PR is the cheap, high-leverage lever.

## Testing

- `npm run build` (tsc) clean; `agentSystemPrompt.ts`/test lint clean.
- New unit tests in `agentSystemPrompt.test.ts`: steering present for a Plus user (all four skill names + the fallback clause), absent for a non-Plus user, and suppressed when the builtin prompt is disabled. Full file: 15/15.

Refs https://github.com/logancyang/obsidian-copilot-preview/issues/104

### DIFF
--- a/src/agentMode/backends/shared/agentSystemPrompt.test.ts
+++ b/src/agentMode/backends/shared/agentSystemPrompt.test.ts
@@ -1,4 +1,4 @@
-import { resetSettings } from "@/settings/model";
+import { resetSettings, updateSetting } from "@/settings/model";
 import {
   setDefaultSystemPromptTitle,
   setDisableBuiltinSystemPrompt,
@@ -6,7 +6,11 @@ import {
   updateCachedSystemPrompts,
 } from "@/system-prompts/state";
 import type { UserSystemPrompt } from "@/system-prompts/type";
-import { buildAgentSystemPrompt, COPILOT_PROMPT_BASE } from "./agentSystemPrompt";
+import {
+  buildAgentSystemPrompt,
+  COPILOT_PLUS_TOOLS_STEERING,
+  COPILOT_PROMPT_BASE,
+} from "./agentSystemPrompt";
 
 jest.mock("@/logger", () => ({
   logInfo: jest.fn(),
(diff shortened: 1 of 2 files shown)


logancyang:
Folded into #2537 (commit `1c304ac9`): the steering now ships with the builtin skills as one feature, and is **un-gated** — sent to every user (gating on `isPlusUser` missed valid self-host mode, which is Plus-enabled but reports `isPlusUser:false`). The fallback clause routes non-Plus / unlicensed agents back to their own tools. Closed in favor of #2537.

PR #2580, opened 2026-06-07
Filter Miyo search results by inclusion/exclusion rules and cap output

wenzhengjiang:
Ports preview repo #131.

Miyo semantic search results now honor the same QA **inclusion/exclusion** scope as locally-indexed search:

- New reusable `createCopilotPatternFilter(app)` predicate in `searchUtils.ts` resolves the active inclusion/exclusion patterns once and keeps only chunks whose source path passes. Paths that don't resolve to a vault `TFile` are kept (the rules can't be evaluated for them).
- `MiyoSemanticRetriever` over-fetches candidates (`RETURN_ALL_LIMIT`) so filtering still leaves enough chunks to fill the final cap, then slices the result set to `finalK` (the requested `maxK`, or `DEFAULT_FINAL_K = 20` when unspecified).
- The result-sizing math moves out of `RetrieverFactory.createMiyoRetriever` into the retriever itself.

Adds tests covering the over-fetch-then-cap behavior and inclusion/exclusion filtering. All `MiyoSemanticRetriever` tests pass.

### DIFF
--- a/src/search/RetrieverFactory.ts
+++ b/src/search/RetrieverFactory.ts
@@ -6,7 +6,6 @@ import { App } from "obsidian";
 import { SelfHostRetriever, VectorSearchBackend } from "./selfHostRetriever";
 import { MiyoSemanticRetriever } from "./miyo/MiyoSemanticRetriever";
 import { MergedSemanticRetriever } from "./v3/MergedSemanticRetriever";
-import { RETURN_ALL_LIMIT } from "./v3/SearchCore";
 import { TieredLexicalRetriever } from "./v3/TieredLexicalRetriever";
 
 /**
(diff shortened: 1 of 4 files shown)


wenzhengjiang:
Addressed the over-fetch concern in 30dda6bf.

Added `hasActiveCopilotPatterns()` and changed the limit to only request `RETURN_ALL_LIMIT` when an inclusion/exclusion filter is active or `returnAll` is set; otherwise it bounds the request to `finalK` (the requested/expanded `maxK`). This restores the old bounded behavior for default Miyo searches while keeping the over-fetch needed to refill the cap when filtering can actually drop results.

Tests updated accordingly: added a case asserting unfiltered searches bound to the requested limit, and one asserting over-fetch still happens when a pattern filter is active.

PR #2585, opened 2026-06-08
Remove unnecessary chat bottom padding calculation

zeroliu:
There is no need to introduce custom logic. Obisidian default CSS already takes care of the status bar style

Screenshot 2026-06-08 at 8 50 55 PM


logancyang:
Heads up: this conflicts with #2584, which fixes Brevilabs/obsidian-copilot-private#54 (status bar covering the chat input/send button).

Quick case for keeping the measurement rather than removing it:

- **It's self-disabling.** `ChatViewLayout` sets `--copilot-status-bar-clearance` to `0` whenever the bar doesn't actually overlap (theme positions content above it, or the bar is hidden). In the default-theme, no-banner case in your screenshot it already measures zero overlap and adds **zero** padding. It only reserves space on a real measured overlap, so it's inert exactly when Obsidian's default suffices.
- **Obsidian's built-in handling is editor-only.** The [documented status-bar vars](https://docs.obsidian.md/Reference/CSS+variables/Window/Status+bar) are all cosmetic (no height/clearance var); the one functional knob, `--status-bar-scroll-padding`, applies scroll-padding to the editor's scroll container. Our chat input is pinned in a custom `ItemView`, not scrolled, so that never applies to it.
- **The bug is the banner case.** Brevilabs/obsidian-copilot-private#54 is about a tall/visible bar (update-available notice, sync status, themes that enlarge it) covering the pinned send button. A clean default-theme screenshot doesn't exercise that.

#2584 keeps the (self-disabling) measurement and just makes it re-measure on a `ResizeObserver` so it doesn't go stale when a status-bar item appears at runtime. Full rationale + forum references in that PR. Worth reconciling the two before either merges.

zeroliu:
The bug is the banner case. Prevent Obsidian's bottom-right banner from covering the chat input obsidian-copilot-preview#99 is about a tall/visible bar (update-available notice, sync status, themes that enlarge it) covering the pinned send button. A clean default-theme screenshot doesn't exercise that.

---

I don't believe this is accurate. My test vault has default status bar with default theme and the bug can be reproduced with the existing implementation. The effect can cause race conditions. The existing self disabling mechanism doesn't always work

zeroliu:
Obsidian's built-in handling is editor-only. The documented status-bar vars are all cosmetic (no height/clearance var); the one functional knob, --status-bar-scroll-padding, applies scroll-padding to the editor's scroll container. Our chat input is pinned in a custom ItemView, not scrolled, so that never applies to it.

---

The screenshot shows obsidian view-content is using hardcoded spacing tokens not --status-bar-scroll-padding. Can you please help me understand why it's related to how we style copilot view content? 

I'm hearing that there are combinations of themes and plugins that I'm missing. Do you mind sharing a test case where relying on default obsidian view content style will lead to UX issues?

logancyang:
Tested this in a real vault across many themes. **This removal is correct** -- I've closed #2584. Here's the confirmed root cause for the record, and a retraction of my earlier comment.

### What actually happens

Obsidian core (`app.css`) sets:

```css
.view-content { padding: var(--size-4-3) var(--size-4-3) var(--size-4-8); }  /* 12px top, 12px sides, 32px bottom */
```

That 32px (`--size-4-8`) bottom padding is core's reserved clearance for the **floating** bottom-right status bar. On the default theme and ~every theme I tested, the status bar floats *over* that 32px region, so the padding is exactly the clearance and the input sits cleanly above the bar -- no gap, no overlap.

### Why the old code caused obsidian-copilot-preview#99

`ChatViewLayout` overrode that bottom padding down to `--size-4-1` (4px) and tried to add the room back via a measured `--copilot-status-bar-clearance`. When the measurement was stale/didn't fire, the input was left with 4px under a floating bar that needs ~32px -> the bar covered the input. So our own override was the bug. Removing it restores core's 32px and fixes it, no JS needed. Your "the self-disabling mechanism doesn't always work" was exactly right.

### Retraction

My earlier comment was wrong on two counts: (1) Obsidian's clearance is **not** editor-only -- it's on `.view-content`, which is our container; (2) the override was not harmlessly self-disabling -- it actively under-padded. Apologies for the noise.

### The one remaining wrinkle: Velocity

Velocity is the only theme I found with a large empty gap after this change. That's because Velocity moves the status bar **out of overlay into normal flow** (a flex sibling below the leaf), so core's 32px no longer overlaps any bar and becomes dead space. That's a theme-side omission -- when a theme relocates the status bar it should also trim `.view-content`'s `--size-4-8` bottom padding. We can't fix it generically without re-covering the input on every floating-bar theme (which is the majority), so I'd leave it to the theme rather than reintroduce measurement. Worth a heads-up to the Velocity author.

LGTM to merge.

PR #2589, opened 2026-06-10
fix: normalize provider base URL per consumer (legacy chat vs opencode agent chat)

logancyang:
Fixes logancyang/obsidian-copilot-preview#152

## The bug

A user set up Gemini and Groq as BYOK providers. Legacy chat and opencode agent chat could not both work at the same time:

- With base URL `https://generativelanguage.googleapis.com/v1beta`, agent chat worked but legacy chat failed ("Failed to fetch").
- After removing `/v1beta`, legacy chat worked but agent chat failed ("opencode finished the turn without returning any assistant text").
- Same story with Groq and `/openai/v1`.

## Why it happens

The provider's saved base URL goes to two different SDKs, and they disagree about what a base URL is.

**Legacy chat adds the version path itself.** Verified in the installed packages:

- Google: we pass the base URL to `ChatGoogleGenerativeAI`, which hands it to Google's SDK. The SDK builds the request as `${baseUrl}/v1beta/models/...` (`@google/generative-ai/dist/index.js:335`, default version `v1beta` at line 308). So if the saved URL already ends in `/v1beta`, the request goes to `/v1beta/v1beta/...` and 404s. That doubled path is visible in the user's log.
- Groq: `ChatGroq` hands the base URL to groq-sdk (`@langchain/groq/dist/chat_models.js:595`), which posts to `${baseURL}/openai/v1/chat/completions` (`groq-sdk/resources/chat/completions.js:8`). Same doubling if the saved URL already ends in `/openai/v1`.

**opencode does the opposite.** It treats the base URL as the complete prefix and only adds the route. So it needs the URL *with* the version path. Give it a bare host and the version segment is missing from every request.

So legacy chat wants `https://api.groq.com` and opencode wants `https://api.groq.com/openai/v1`. One saved value cannot satisfy both — until now the user just picked which mode to break.

We also seed the bad values ourselves: the provider dialog fills in Groq's URL in the versioned form (breaks legacy chat) and Google's in the host-only form (breaks agent chat). And the setup "Test" button accepts both forms, so everything looks fine until an actual chat message fails.

## The fix

Don't change what's saved. Instead, each consumer converts the saved URL into the form its own SDK needs (new helpers in `src/utils/providerBaseUrl.ts`):

- **Legacy chat** (`chatModelManager.ts`): before handing the URL to the SDK, remove a trailing `/v1beta` or `/v1` (Google) / `/openai/v1`, `/openai`, or `/v1` (Groq), since the SDK is about to add that path back. This is the same trick `listGoogleModels.ts` already used for the model list. Only a *trailing* suffix is removed, so gateway URLs with `/v1/` in the middle are untouched.
- **opencode** (`buildOpencodeConfig`): if the saved URL is just the provider's own standard endpoint in any spelling (with or without the version path, trailing slash, etc.), don't send a baseURL at all — opencode's built-in default for that provider is already the correct versioned URL. A URL on a different host is a real proxy override and is still sent through unchanged, exactly as before.

Result: any spelling of the Gemini or Groq endpoint now works in both chat modes.

## Testing

- New `src/utils/providerBaseUrl.test.ts`: 19 cases covering suffix stripping, trailing slashes, proxy hosts, mid-path `/v1` left alone, and endpoint recognition (rejects `/v1alpha` and unknown providers).
- `chatModelManager.bridge.test.ts`: mocks capture the constructor config and assert the stripped URL actually reaches `ChatGoogleGenerativeAI` / `ChatGroq`.
- `OpencodeBackend.test.ts`: Google host-only seed and Groq versioned seed both produce a config without baseURL; the existing proxy passthrough test still passes unchanged.
- Full suite: 3638 passed, 0 failed. `tsc -noEmit -skipLibCheck` clean, eslint clean.

### DIFF
--- a/src/LLMProviders/chatModelManager.bridge.test.ts
+++ b/src/LLMProviders/chatModelManager.bridge.test.ts
@@ -6,7 +6,26 @@ import { getSettings, setSettings } from "@/settings/model";
 import ChatModelManager from "./chatModelManager";
 
 jest.mock("@langchain/anthropic", () => ({ ChatAnthropic: class {} }));
-jest.mock("@langchain/groq", () => ({ ChatGroq: class {} }));
+// Capture constructor configs so tests can assert what the manager actually
+// hands the LangChain clients (e.g. base-URL normalization).
+jest.mock("@langchain/groq", () => {
+  class ChatGroq {
+    static configs: unknown[] = [];
+    constructor(config: unknown) {
+      ChatGroq.configs.push(config);
+    }
+  }
+  return { ChatGroq };
+});
+jest.mock("@langchain/google-genai", () => {
+  class ChatGoogleGenerativeAI {
+    static configs: unknown[] = [];
+    constructor(config: unknown) {
+      ChatGoogleGenerativeAI.configs.push(config);
+    }
+  }
+  return { ChatGoogleGenerativeAI };
+});
 
 function bridgedModel(overrides: Partial<CustomModel> = {}): CustomModel {
   return {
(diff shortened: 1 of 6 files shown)


logancyang:
@codex review

PR #2627, opened 2026-06-20
feat(models): flag non-vision models, guard image sends

zeroliu:
## Summary

Two gaps, one root cause: models with missing/wrong capability metadata get no `vision`/`reasoning`, and nothing stopped a user sending images to a non-vision model — it failed silently (and frustrated users). This PR **flags the exception instead of the norm**: the model UI now surfaces a single indicator *only* on models that **don't** support vision, hard-blocks image sends to those models on **both** the chat and Agent Mode paths

Screenshot 2026-06-21 at 8 38 51 PM
Screenshot 2026-06-21 at 8 39 08 PM

## Modality indicator — only flag models without vision

Vision is ubiquitous on modern models, so badging it is noise. We invert it:

- **Known to lack vision** → a single muted **eye-off** icon (with a "This model does not support image inputs." tooltip). This is the only vision signal we surface.
- **Vision-capable** → no icon. Vision is the norm; we don't badge it (mirrors the already-hidden reasoning icon).
- **Unknown** (no modality snapshot — e.g. an agent-provided model not in the catalog) → no icon. We never assert a missing capability we don't actually know about.
- **Web search** keeps its blue globe; **reasoning** stays hidden.

The distinction is encoded in `capabilitiesFromConfiguredInfo()`: a snapshot **with** modalities but no image input returns a defined array without `VISION` ("known text-only" → eye-off); a snapshot with **no** modalities returns `undefined` ("unknown" → render nothing). `hasCapabilityIcons()` drives the surrounding wrapper so a vision-capable/unknown row never renders an empty icon slot. This applies consistently across the BYOK Configure dialog, the model checklist, and the agent model pickers.

## Guard image sends to non-vision models

The guard fires **only when a model's capabilities are known and exclude vision**; unclassifiable models keep `capabilities: undefined` and are never falsely blocked.

- `Chat.handleSendMessage` (chat path) and `AgentChatInput.handleSendMessage` (Agent Mode) both block the send when images are attached and the active model is known to lack vision. The `Notice` names the model; the typed text + images are preserved (no silent failure, nothing cleared).
- Agent picker entries are enriched with capabilities derived from the models.dev catalog by `(provider, baseModelId)`, threaded as a structural type to respect the agent-mode `ui` import boundary (no `@/modelManagement` import); catalog misses stay `undefined`.

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -22,3 +22,15 @@ if (typeof Node !== "undefined" && !Object.prototype.hasOwnProperty.call(Node.pr
     configurable: true,
   });
 }
+
+// Obsidian exposes `activeDocument` / `activeWindow` globals pointing at the
+// focused popout's document/window. Under jsdom there's only one document, so
+// alias them onto `window` (the jsdom global object) — plugin code that portals
+// into `activeDocument.body` (e.g. the Radix tooltip) would otherwise throw
+// `activeDocument is not defined`.
+if (typeof window.activeDocument === "undefined") {
+  window.activeDocument = window.document;
+}
+if (typeof window.activeWindow === "undefined") {
+  window.activeWindow = window;
+}
--- a/src/agentMode/backends/opencode/opencodeModelResolve.ts
+++ b/src/agentMode/backends/opencode/opencodeModelResolve.ts
@@ -1,6 +1,6 @@
 import type { CopilotSettings } from "@/settings/model";
 import type { ConfiguredModel, Provider } from "@/modelManagement";
-import { providerRequiresApiKey } from "@/modelManagement";
+import { capabilitiesFromConfiguredInfo, providerRequiresApiKey } from "@/modelManagement";
 import type { EnabledModelCredentialState, EnabledModelEntry } from "@/agentMode/session/types";
 
 export interface OpencodeProviderMapping {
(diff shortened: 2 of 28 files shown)


logancyang (commented):
Review against `v4-preview`:

**Finding**
- **[P2] Quick Chat treats unknown model capabilities as known text-only.** In `src/components/Chat.tsx`, the new image guard blocks whenever the selected model has a defined `capabilities` array without `VISION`. But `src/components/chat-components/useChatModelPicker.ts` always creates `capabilities: []` when `ConfiguredModel.info.modalities` is absent. That turns legacy/manual/unknown snapshots into “known text-only,” showing the eye-off icon and blocking images even though the PR’s intended semantics say unknown capability should remain unblocked. The chat picker should use `capabilitiesFromConfiguredInfo()` and leave `capabilities` undefined when modalities are absent.

**Open question**
- The PR title/body mentions BYOK capability overrides, but I didn’t find persisted override UI/state in the current diff. If that feature was dropped, the PR description should be updated; if it is intended, it appears missing.

**Verification**
- Reviewed current head `dc39e1aa` against `v4-preview` base `f10430a`.
- Ran focused tests: 8 suites / 118 tests passed. `ConfigureProviderDialog.test.tsx` still emits React `act(...)` warnings from the new catalog-loading effect.

PR #2648, opened 2026-06-30
fix(agent-mode): coalesce streamed prose split by an invisible trail part

zeroliu:
## Problem

In agent mode, a streamed assistant message occasionally shows a **line break in the middle of a sentence** in the live UI, e.g.:

> …That's exactly the Pillar-2 "let the **⏎**
> kid feel the change, he proposes the fix himself" working in the wild.

The model never produced the break: the raw Claude Code session log has the text as one contiguous block, and the **saved** conversation `.md` (built from the flat `displayText`, deltas concatenated with no separator) has no break either. It is a **render-only artifact** that self-heals on reload.

## Root cause

The live trail renders each streamed `text` part as its own block-level markdown `<div>`. When a part that is **invisible at the top level** lands between two prose deltas, one sentence splits into two stacked blocks. The three invisible triggers:

- a dropped `ToolSearch` (deferred-tool loader, filtered from the trail),
- an empty `plan` (`PlanPill` renders `null`),
- a **concurrent background sub-agent's event** (rendered nested under its parent card, never as a top-level peer) — the reason this only showed up in a session that backgrounded sub-agents *while still streaming prose*.

## Fix

One shared `isTopLevelInvisible` predicate, applied to all three consumers of the parts stream (`src/agentMode/ui/agentTrail.ts`):

1. **Live render** (`foldNodes`) — skip empty-plan nodes and **coalesce adjacent text nodes**, so the streamed trail is byte-identical to the saved `displayText`.
2. **Copy/Insert** (`agentResponseText`) — derived from the folded trail, so a split sentence no longer gets a `\n\n` blank line; genuinely separate prose (around a *visible* tool card / reasoning / non-empty plan) still gets `\n\n`.
3. **Collapsed "Worked for X"** (`splitTrailingText` + `AgentTrailView`) — invisible parts are transparent to the trailing-run boundary, and the final run renders as one coalesced block, so the collapse can't strand half a sentence in the hidden research.

## Testing

- New unit tests cover all three invisible triggers (hidden ToolSearch, empty plan, sub-agent child) across `buildAgentTrail`, `agentResponseText`, and `splitTrailingText`; visible-separator behavior is preserved.
- `npx tsc --noEmit`, `npm run lint`, `npm run format` clean.
- **1350/1350 agentMode tests pass.**

### DIFF
--- a/src/agentMode/ui/AgentTrailView.tsx
+++ b/src/agentMode/ui/AgentTrailView.tsx
@@ -78,10 +78,11 @@ export const AgentTrail: React.FC<AgentTrailProps> = ({
       return (
         <div className="tw-group tw-flex tw-flex-col tw-gap-1">
           <WorkedForBlock research={research} durationMs={turnDurationMs} app={app} />
-          {final.map((p, i) => (
-            // eslint-disable-next-line @eslint-react/no-array-index-key -- text parts are append-only and may contain duplicate text
-            <AgentMarkdownText key={`final-${i}`} text={p.text} app={app} />
-          ))}
+          {/* `final` may be several text parts only because a top-level-invisible
+              part (a background sub-agent event, a dropped ToolSearch, an empty
+              plan) split one prose run; join with "" to render the original
+              sentence as one block, not a broken-apart paragraph. */}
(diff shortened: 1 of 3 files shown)


zeroliu:
No need for now as we disabled background agent

PR #2656, opened 2026-07-03
feat(agent-mode): show agent file-edit diffs — chip, read-only pane, permission preview

zeroliu:
## Why

When the agent edits a file, it was hard to see what actually changed — the chat showed a raw JSON input blob (for the default Claude SDK agent) or a naive full-file dump, and there was no way to open a real diff. That erodes trust in the agent. This makes every agent edit legible.

**Root cause found during planning:** the in-process Claude Agent SDK already ships each edit's result on `tool_use_result` — with `originalFile` (full before) and a real minimal `structuredPatch` — for both Edit and Write, but the SDK translator **discarded it** (extracted text only). The old `diffStats` also counted *every* line (ACP-only), so a 1-line change reported the whole file size.

## What

- **Chat trail:** a completed edit now shows an always-visible `+N / −M` chip, and the **whole row opens a read-only diff pane** in the editor area.
- **Permission card:** the pending edit shows the same `+N / −M` chip + a capped preview, and clicking opens the same pane (a Write reads the current file for its true "before").
- **Real diffs, one shape:** everything normalizes to a canonical `EditDiff {path, oldText, newText}`, sourced from the SDK's `tool_use_result` (guard-parsed against the documented, byte-stable `FileEditOutput`/`FileWriteOutput`) for completed edits, from ACP `{type:"diff"}` content otherwise, and synthesized from tool input for permission previews.

## Design decisions (reviewed plan)

- Support the default **Claude SDK** agent, not just ACP.
- A **dedicated read-only** `AgentDiffView` (not the composer `ApplyView` accept/reject flow); its split/side-by-side primitives are shared with `ApplyView`.
- The **whole edit row** opens the pane (filename included).
- Source completed SDK diffs from `tool_use_result` (**hybrid**: guard-parse + input fallback, **no SDK bump** — verified 0.3.197 doesn't improve the `unknown` typing). A shape-pin test trips loudly if the SDK reshapes the result.
- The permission card reads the current file for a Write's "before".

## Phases (one commit each)

1. `9d2354b6` — canonical `deriveEditDiff` + real jsdiff line stats.
2. `66bb1db6` — surface real SDK edit diffs from `tool_use_result`.
3. `068f1c11` — read-only `AgentDiffView` + shared diff primitives (behavior-preserving `ApplyView` refactor).
4. `8bb1491e` — edit-row chip + whole-row opens the pane.
5. `e4b4e084` — permission-card preview + open.
6. `daab29af` — docs + formatting.

## Notable review catches (fixed before commit)

- **`$`-pattern corruption:** the Edit after-text reconstruction used `String.replace`, which mangles a `newString` containing `$&`/`$1` — switched to a literal first-occurrence replace (+ proof test).
- **Dead/bubbling chevron:** SDK edits emit `[text, diff]` output, so edit rows were both pane-opening and expandable — made edit rows pane-only.
- **Orphaned React root** on pane reuse; **absolute-vs-vault-relative path** mismatch that broke pane reuse from the permission card. Both fixed.

## Testing

Full suite **4384 tests / 0 failures**, `eslint` (incl. agent-mode boundaries) clean, and the production build all green. New unit tests cover the derivation, the guard-parse (shape-pinned to the SDK types), and the permission synthesis.

> Base branch: `v4-preview` (the agent-mode code doesn't exist on `master`).


zeroliu:
@codex review

PR #2660, opened 2026-07-04
Fix project tab strip collapsing into overflow despite available width

Emt-lin:
## Summary

`AgentTabStrip` collapsed project chat tabs into the `...` overflow menu even when the panel had room for them. `computeVisibleCount()` is driven by `stripWidth` state that started at `0` and was updated **only** by an async `ResizeObserver` callback — with no synchronous fallback measurement. On a real remount (detach/reopen the Agent Mode view), a missed or delayed first `ResizeObserver` delivery left `stripWidth` stuck at `0`, so the strip showed only 1 tab despite unused width.

Confirmed via an A/B repro: stub `window.ResizeObserver` to a no-op + force a genuine remount → 3 tabs collapse to 1 at an identical panel width; unstubbed, all 3 render.

## Fix

- Switch the measurement effect from `useEffect` to `useLayoutEffect` and **measure synchronously once** before handing the callback to `ResizeObserver` (sync primer), so the first paint never uses `stripWidth = 0`.
- Depend on `sessionCount` so the effect re-runs on the `0 → N` mount path — the strip returns `null` at `0`, so `stripRef` is absent until the first session exists.
- Guard `typeof ResizeObserver === "undefined"` for the jsdom test env.

This mirrors the sync-primer pattern already used by `PatternListEditor`, `ProjectContextBadgeList`, and `ProjectContextSourceEditor`; `AgentTabStrip` was the lone outlier.

A `DESIGN NOTE` records why the `ResizeObserver` is left on the bare global (not `strip.win.ResizeObserver`) for now: it's a codebase-wide consistency question (all 6 call sites use the bare global), not this bug's root cause — tracked for a separate sweep.

## Verification

- `npm test -- src/agentMode/ui/AgentTabStrip.test.ts` → 12/12 pass (pure-function coverage for `computeVisibleCount` / `partitionSessions`).
- `eslint` + `prettier` clean on the changed file.
- Manual: with `ResizeObserver` stubbed + the tab strip force-remounted on a project with 3+ sessions, all tabs that fit render instead of collapsing to 1.

## Notes

- Base: `v4-preview`. Single-file change (`src/agentMode/ui/AgentTabStrip.tsx`).
- Addresses logancyang/obsidian-copilot-preview#206 (cross-repo; won't auto-close).
- Out of scope, flagged in the issue's open questions: the "`ResizeObserver` should use `element.win`" codebase-wide sweep, and a separate `requestUrl`-no-timeout hardening.

### DIFF
--- a/src/agentMode/ui/AgentTabStrip.tsx
+++ b/src/agentMode/ui/AgentTabStrip.tsx
@@ -130,6 +130,7 @@ export const AgentTabStrip: React.FC<Props> = ({ manager }) => {
   // sessions, the global workspace shows the global ones. `getSessions()` stays
   // reserved for whole-pool consumers (draft prune / auto-spawn), per §2.4.
   const sessions = manager.getSessionsForScope(manager.getActiveProjectId());
+  const sessionCount = sessions.length;
   const activeId = manager.getActiveSession()?.internalId ?? null;
   const isCreating = manager.getIsStarting();
   const [renamingId, setRenamingId] = React.useState<string | null>(null);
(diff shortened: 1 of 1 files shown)


Emt-lin:
@codex review

PR #2677, opened 2026-07-13
feat(report-issue): bundle sanitized data.json in the Agent Mode issue report

zeroliu:
## What

Adds a **sanitized copy of the plugin settings (`data.json`)** to the Agent Mode "Report an Issue" bundle, so maintainers can diagnose configuration-dependent bugs. Previously the bundle contained only `report.md`, the screenshot, the frame log, and the optional opencode log.

### Behavior

- **New file in the bundle:** the raw on-disk `data.json` (via `Plugin.loadData()`, so the report reflects what is actually persisted) is sanitized and written as `data.json.txt` — same `.txt` suffix workaround as the frame log, since GitHub's issue-attachment allowlist rejects `.json`.
- **Proactive redaction** via a new pure `sanitizeSettingsDataForReport()` (`src/utils/sanitizeSettingsData.ts`). Sensitive values are **masked as `[REDACTED]`** rather than removed, so which providers/keys are configured stays visible. Covered:
  - everything matching the canonical `isSensitiveKey()` (all `*ApiKey` fields, `plusLicenseKey`, tokens, secrets, passwords), at any depth — including per-model `apiKey` in `activeModels` / `activeEmbeddingModels`
  - **every value in any `envOverrides` map** (backend env vars like `AWS_ACCESS_KEY_ID` don't match key-name patterns, so they're masked wholesale, key names kept)
  - infra identifiers (Azure instance/deployment names, org IDs, API versions), `userId`, and legacy `enc_*` encrypted strings wherever they appear
  - empty values pass through unchanged so "not configured" stays visible
- **Surfaced in the dialog:** the "When you click Prepare report" list now states that a copy of your Copilot settings is included with secrets replaced by `[REDACTED]`, and the privacy callout notes the copy still describes your configuration. The settings-page button description and `report.md`'s attachment list ("data.json.txt (plugin settings; secrets redacted)") say the same.
- **Graceful degradation:** if settings can't be loaded or written, the report is still produced without the snapshot (same best-effort pattern as the other bundle files).

## Tests

- New `src/utils/sanitizeSettingsData.test.ts` (10 tests) against a realistic disk-mode `data.json` fixture, including a catch-all assertion that **no secret string appears anywhere in the serialized output**.
- Extended `src/utils/issueReport.test.ts`: snapshot written and listed when provided, skipped when null or on write failure, annotated in `report.md`.
- Full suite: 4415 passing. `npm run format`, `npm run lint`, and `tsc --noEmit` clean.

## Docs

- `docs/agent-mode-and-tools.md` "Reporting an Issue" section updated, including the privacy note.

https://claude.ai/code/session_01W4ZjSKUSbWyWjx4CLeJS5v


zeroliu:
@codex review

PR #2688, opened 2026-07-15
fix(agent-mode): keep background task turns alive

zeroliu:
## Stack position

1 of 3, based on `v4-preview`.

Stack:
1. #2688 — keep background task turns alive
2. #2686 — project task lifecycle onto launch cards
3. #2661 — fold TaskOutput into the launch card

PR #2687 has merged into `v4-preview`; this branch was restacked onto that merged head.

## What

- Updates the Claude Agent SDK and the permission callback typing it requires.
- Enables TaskOutput and installs a Stop hook that keeps a turn alive while Claude background tasks or workflows are still active.
- Covers Stop-hook behavior alongside the owning Claude backend process.

## User impact

Claude no longer ends the turn before background work completes. At this stack point, users get the complete result, but launch and TaskOutput activity still appear as separate cards. Later PRs improve that presentation.

## Verification

- ClaudeSdkBackendProcess.test.ts and permissionBridge.test.ts: 64 tests passed
- npm run lint

The complete top-of-stack tree also passes the full test suite and production build.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "AGPL-3.0",
       "dependencies": {
         "@agentclientprotocol/sdk": "^0.20.0",
-        "@anthropic-ai/claude-agent-sdk": "^0.2.126",
+        "@anthropic-ai/claude-agent-sdk": "0.3.206",
         "@dnd-kit/core": "^6.3.1",
         "@dnd-kit/sortable": "^10.0.0",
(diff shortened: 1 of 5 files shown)


logancyang (approved):
Reviewed the SDK upgrade and Stop-hook behavior. The production build and all 64 focused tests pass; I found no blocking issues.

PR #2689, opened 2026-07-15
fix(agent-mode): surface Claude usage errors

zeroliu:
## Summary
Screenshot 2026-07-15 at 5 19 20 PM

| SDK evidence | Before | After |
| --- | --- | --- |
| `result.is_error=true`, HTTP 429, exact reset text | mapped to `end_turn` | rejected with Claude provider text |
| no streamed text or tool activity | generic empty-turn fallback | visible provider error |

**Why:** Claude can report exhausted subscription usage as a success-shaped terminal SDK result. Copilot discarded the synthetic assistant error and treated that result as a normal empty turn, hiding the actionable reset message.

**What:** The Claude SDK adapter now rejects success-shaped terminal errors with the exact non-empty provider result, so the existing session error path shows the reset message in chat.

## Decisions

- Preserve the exact human-readable Claude usage message, including the reset time when provided.
- Detect failure from the terminal `result.is_error` contract; `rate_limit_event` remains diagnostic because it also carries allowed warnings.
- Keep the change provider-local and reuse the existing prompt-error rendering rather than expanding the shared session protocol.

## Changes

- `709dfb8e` — recognize success-shaped Claude terminal errors, replay the captured rate-limit SDK sequence in the `prompt()` suite, and document the visible reset-message behavior. Verify by confirming exhausted usage rejects with the provider message while ordinary success still returns `end_turn`.

## Notes / follow-ups

- Pre-limit warning UX is intentionally out of scope; this change handles terminal provider failures.

### DIFF
--- a/docs/agent-mode-and-tools.md
+++ b/docs/agent-mode-and-tools.md
@@ -166,6 +166,8 @@ Claude can delegate independent work to background subagents. Agent Mode keeps t
 
 This requires Claude Code 2.1.206 or newer. Copilot checks the installed version when it starts and whenever you apply or auto-detect a Claude binary. An older binary is marked **Incompatible version** in Settings → Copilot → Agents → Claude and cannot start a session. If you select Claude in chat, Copilot shows the required version and a **Configure Claude** button that opens the same setup dialog; installation guidance stays in that dialog.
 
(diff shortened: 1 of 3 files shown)


zeroliu:
@codex review

PR #2690, opened 2026-07-15
Redact diagnostic logs before bundling them for a bug report

logancyang:
## Problem

Agent Mode's "Report an Issue" flow (Settings → Advanced, and the agent pane button) bundles the frame log and OpenCode log for the user to drag onto a **public** GitHub issue. `assembleReportBundle` copied them **verbatim** (`runtime.copyFile`). Those logs contain absolute home paths — which leak the OS username (the same exposure as unredacted vault paths, cf. #213) — and can contain emails or API tokens. Nothing redacts them, so a reporter publishes all of it.

## Fix

Read → redact → write the two logs instead of copying:

- **`src/utils/redactLog.ts`** (new, pure): `redactLogText(text)` — a deterministic pattern pass (home-dir usernames in Unix/Windows paths, emails, provider key shapes `sk-`/`AIza`/`AKIA`/`ghp_`/`xox`, bearer tokens, and `key/secret/password` field values → visible markers). Synchronous and fast, because a frame log can be tens of megabytes — an LLM pass is not an option. Best-effort by design; the flow still shows the bundle for the user to review before attaching (the manual drag is the final gate).
- **`issueReport.ts`**: `ReportRuntime` gains `readFile`, drops the now-unused `copyFile`; the two near-duplicate copy blocks collapse into one redact-and-write loop. Screenshot stays a binary write.

## Verification

- `redactLog.test.ts`: each pattern, a clean-text no-op, and idempotency (re-redacting is a no-op).
- `issueReport.test.ts`: the written frame/opencode files are redacted (home path → `/Users/<user>`, `sk-…` → `<secret>`, originals absent), and a log that fails to read is skipped rather than failing the report.
- 19 tests pass; `prettier`, `eslint`, and `tsc --noEmit` clean on the change; the sole caller (`ReportIssueModal`) uses the default runtime, so dropping `copyFile` strands nothing.

### DIFF
--- a/src/utils/issueReport.test.ts
+++ b/src/utils/issueReport.test.ts
@@ -8,8 +8,9 @@ import {
 
 function makeRuntime(overrides: Partial<ReportRuntime> = {}) {
   const writes: Array<{ path: string; data: string | Uint8Array }> = [];
-  const copies: Array<{ src: string; dest: string }> = [];
   const mkdirs: string[] = [];
+  // Every log read carries a home path and a secret so tests can assert the
+  // written bundle copy is redacted rather than verbatim.
   const runtime: ReportRuntime = {
     join: (...parts) => parts.join("/"),
     mkdir: async (p) => {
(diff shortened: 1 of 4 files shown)


logancyang:
@codex review

PR #2770, opened 2026-08-07
Drop the create-agent-issue skill, superseded by create-issue

Emt-lin:
The user-level `create-issue` skill is a strict superset of this repo-local copy:

- It resolves the target repo through the shared routing map, which already encodes the Copilot split (issues in `logancyang/obsidian-copilot-preview`, PRs here against `v4-preview`) — so hardcoding the repo bought nothing.
- It **assigns** the new issue and verifies the assignee stuck.
- It files a **board card** and sets the board `Priority` field.

This copy did neither of the last two, and explicitly listed "assign users" under what it does not do. Two issues filed through it this week landed unassigned and off-board, which is how the gap surfaced; both were backfilled by hand.

Its only remaining advantage — hardcoding the repo and the `agent-mode` label — is expressible as arguments to the skill that replaces it.

No code or docs referenced the skill directory, so this is a pure deletion.


Emt-lin:
Closing: keeping the repo-local skill. The two follow-up issues it filed unassigned/off-board were backfilled by hand, so nothing is outstanding.

PR #2803, opened 2026-08-11
[Provider keychain names] - Step 1: Use readable provider identities

zeroliu:

## Why

Obsidian exposes Copilot's SecretStorage IDs in its Keychain manager. Provider credentials currently appear under raw UUID-based entries, so a user cannot tell whether an entry belongs to OpenRouter, OpenAI, or another provider before removing it.

Renaming a provider also needs one authoritative credential location: the visible provider name and the local Keychain pointer must not drift apart or leave the working key behind.

## What

| Scenario | Before | After |
| --- | --- | --- |
| Add a provider | Credential entry is led by an opaque provider UUID | Entry is led by a safely encoded form of the unique provider name |
| Add another provider with the same name | Duplicate display names are persisted | Copilot allocates the next readable numbered name |
| Rename a provider | The old Keychain pointer can remain authoritative | Copilot writes the credential under the new name first, updates the pointer, then removes the old entry |
| Destination write fails | A partial rename can strand the credential | The old name, pointer, and credential remain authoritative |

## Non goal

- Removing the vault namespace or preserving downgrade access to builds that only understand the UUID entry.
- Permanently dual-writing old and new IDs, or touching Keychain entries owned by other plugins.
- Removing every opaque collision/recovery token or changing model and non-provider credential names; this slice makes provider entries name-led.
- Keeping the Keychain ID unchanged after a provider rename; this design safely moves the local credential to follow the unique display name.

## Screenshot

Not available — the changed surface is Obsidian's external Keychain manager, and this branch was not loaded into the configured test vault for a GitHub-hosted capture.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Provider identity and registry tests cover naming, collisions, Unicode encoding, rename, and rollback |
| A defect would fail CI or be obvious on first use | ✅ | Deterministic tests exercise every changed registry write path |
| A revert fully restores prior state, including persisted data | ❌ | A reverted build may not recognize a pointer already moved to the new Keychain ID |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | Provider secret naming and movement change |
| No public API, plugin API, message, or on-disk contract changes | ❌ | The persisted Keychain pointer contract changes from UUID-led to name-led IDs |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | Provider rename and API-key writes use a new write-before-pointer sequence |
| No hot-path behavior lacks deterministic coverage | ✅ | Add, update, rename, set-key, collision, and failure directions are covered |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | No untested human-only behavior is required for correctness |

Review: inspect `src/modelManagement/providers/providerIdentity.ts` — provider-name normalization and Keychain-ID construction — and `src/modelManagement/providers/ProviderRegistry.ts` — `add()`, `update()`, and `setApiKey()` — line by line, then run Verification steps 2–3, including the destination-write failure cases.

## Verification

1. Load `codex/provider-keychain-identity` in a test vault and add a provider named `OpenRouter Work` with a working key.
2. Open Obsidian's Keychain manager and confirm the Copilot entry is name-led rather than UUID-led; use the provider once to confirm the credential remains usable.
3. Rename it to `OpenRouter Personal`; confirm the new entry becomes authoritative and the old entry is removed. Then run `npm test -- src/modelManagement/providers/providerIdentity.test.ts src/modelManagement/providers/ProviderRegistry.test.ts --runInBand` to exercise Unicode, collision, and forced write-failure variants.

### DIFF
--- a/src/modelManagement/providers/ProviderRegistry.test.ts
+++ b/src/modelManagement/providers/ProviderRegistry.test.ts
@@ -6,11 +6,12 @@
  * `resetSettings` / `setSettings`).
  */
 
-import { resetSettings, getSettings } from "@/settings/model";
+import { resetSettings, getSettings, setSettings } from "@/settings/model";
 import { KeychainService } from "@/services/keychainService";
 
 import type { ProviderAdapter } from "./adapters/ProviderAdapter";
 import { ProviderAdapterRegistry } from "./adapters/ProviderAdapterRegistry";
+import { buildProviderKeychainId } from "./providerIdentity";
 import { ProviderRegistry } from "./ProviderRegistry";
 
 import type { App } from "obsidian";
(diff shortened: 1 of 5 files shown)


zeroliu:
@codex review

PR #2805, opened 2026-08-11
[Provider keychain names] - Step 2: Migrate credentials safely

zeroliu:

## Why

Existing users can already have duplicate provider names and credentials stored under UUID-based or earlier readable IDs. A normal write-path fix does not repair those installations, and settings sync can deliver a provider rename to a device whose local Keychain has not moved yet.

Migration must preserve the user's working account choice. An explicit clear must not resurrect an old key, and two different leftover credentials must not be resolved by arbitrary ordering.

## What

| Existing state | Startup result |
| --- | --- |
| Duplicate provider names | Settings v9 keeps the oldest deterministic winner and adds readable numbered suffixes to the others |
| One unambiguous UUID or prior-name credential | The value moves to the current name-led entry and the provider pointer follows it |
| Explicitly cleared pointer or empty Keychain tombstone | Older local aliases remain inactive and cannot resurrect the credential |
| Multiple candidates with different values | Copilot records a conflict and preserves every candidate without choosing an account |
| Keychain enumeration, read, write, or cleanup failure | Copilot avoids destructive inference, keeps the authoritative source, and retries reconciliation on a later startup |

Migration and device-local reconciliation finish before initial Plus sync and agent discovery consume provider settings.

## Non goal

- Syncing Keychain secrets between devices or automatically choosing between genuinely different conflicting credentials.
- Removing the vault namespace, supporting downgrade access, permanently dual-writing IDs, or touching another plugin's entries.
- Migrating model-specific and other non-provider credential families, or removing every remaining collision/recovery token.

## Screenshot

Not applicable — this PR changes persisted-settings migration and device-local Keychain reconciliation, with no plugin visual state.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Migration, startup, Keychain, and reconciliation tests cover every defined state |
| A defect would fail CI or be obvious on first use | ✅ | Deterministic regression tests cover duplicate repair and credential authority |
| A revert fully restores prior state, including persisted data | ❌ | Settings v9 name repairs and removed legacy Keychain entries are one-way |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | Existing provider secrets are copied, repointed, and cleaned up |
| No public API, plugin API, message, or on-disk contract changes | ❌ | The settings version and persisted provider names and pointers change |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | Startup now awaits migration and reconciliation before provider consumers start |
| No hot-path behavior lacks deterministic coverage | ✅ | Null, undefined, tombstone, ambiguity, enumeration, and write/delete failure paths are covered |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Correctness does not depend on an untested visual or manual-only path |

Review: inspect `src/settings/migrations/providerNameMigration.ts` — `planProviderNameMigration()` — `src/modelManagement/providers/ProviderRegistry.ts` — `reconcileCredentials()` — and `src/services/providerStartup.ts` — `prepareProviderStartup()` — line by line, then run Verification steps 1–3, including tombstone, conflicting-value, and forced-failure variants.

## Verification

1. Run `npm test -- src/settings/migrations/providerNameMigration.test.ts src/settings/migrations/runSettingsMigrations.test.ts src/modelManagement/providers/ProviderRegistry.test.ts src/services/keychainService.test.ts src/services/providerStartup.test.ts --runInBand` and confirm the adversarial migration and reconciliation cases pass.
2. On the preceding stack branch, create two providers with the same display name and working local keys, then load `codex/provider-keychain-migration` in the same test vault.
3. Confirm settings show deterministic numbered names, each provider still uses its original credential, the current name-led Keychain entries are authoritative, and redundant UUID entries are gone. Clear one credential and restart once more to confirm no older alias is reactivated.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -65,7 +65,7 @@ import {
 } from "@/services/webViewerService/webViewerServiceSingleton";
 import { WebSelectionTracker } from "@/services/webViewerService/webViewerServiceSelection";
 import VectorStoreManager from "@/search/vectorStoreManager";
-import { runSettingsMigrations } from "@/settings/migrations";
+import { prepareProviderStartup } from "@/services/providerStartup";
 import { CopilotSettingTab } from "@/settings/SettingsPage";
 import {
   type CopilotSettings,
@@ -229,9 +229,6 @@ export default class CopilotPlugin extends Plugin {
         syncCopilotPlusProvider(this.modelManagement, !!isPaidUser, licenseKey)
       );
     };
-    // Initial reconcile: an already-signed-in user's `isPaidUser` is restored
-    // from disk without firing the subscription, so register on load.
-    syncPlus(getSettings().isPaidUser, getSettings().plusLicenseKey);
     this.settingsUnsubscriber = subscribeToSettingsChange((prev, next) => {
       void (async () => {
         try {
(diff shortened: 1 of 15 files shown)


zeroliu:
@codex review

PR #2806, opened 2026-08-11
[Provider keychain names] - Step 3: Require unique provider names

zeroliu:

## Why

Readable provider Keychain entries depend on provider display names being unique, but the provider dialog previously allowed blank or duplicate names. A user could add two visually identical OpenRouter providers and only discover the ambiguity later in settings or the Keychain manager.

The dialog should prevent ordinary collisions before save while still giving a useful default when someone intentionally adds a second instance.

## What

> **Before:** Adding another provider starts with the same display name, and Save does not explain that the name is blank or already used.
>
> **After:** Copilot suggests the next available numbered name, validates against BYOK, Plus, and agent providers, and shows an associated inline error while Save is disabled for blank or equivalent names.

Names that differ only by surrounding whitespace, capitalization, or Unicode compatibility formatting count as the same name. Editing a provider keeps its current name valid while reserving every other provider's name.

## Non goal

- Adding rename controls for Plus or agent providers; their names participate in uniqueness but remain managed by their existing flows.
- Removing the registry's final suffixing safeguard for a rare save-time race.
- Removing the vault namespace, supporting downgrade access, permanently dual-writing IDs, touching another plugin's entries, or changing non-provider credential naming.

## Screenshot

Not available — the isolated valid and duplicate-name stories bundle successfully, but the aggregate gallery is currently blocked by pre-existing SVG-loader and Claude SDK Node-built-in errors, so no GitHub-hosted rendered capture could be produced.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Identity, dialog, field, and shared form tests cover suggestions, validation, save guards, and accessibility |
| A defect would fail CI or be obvious on first use | ✅ | Invalid-name saves and accessible error relationships are deterministic test assertions |
| A revert fully restores prior state, including persisted data | ✅ | This slice adds validation and documentation without a new migration |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | Provider-name input handling and save eligibility change |
| No public API, plugin API, message, or on-disk contract changes | ✅ | The shared form props are additive and internal; persisted schema is unchanged |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Existing save lifecycle remains; validation only guards entry to it |
| No hot-path behavior lacks deterministic coverage | ✅ | New/edit, blank, duplicate, Unicode, self-exclusion, and race fallback behavior are covered |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Any visual defect is confined to the provider dialog and appears while typing |

Review: inspect `src/modelManagement/providers/providerIdentity.ts` — `providerDisplayNameValidationError()` — `src/modelManagement/ui/dialogs/ConfigureProviderDialog.tsx` — `ConfigureProviderForm` and `ConfigureProviderBody` — `src/modelManagement/ui/components/ProviderDisplayNameField.tsx` — `ProviderDisplayNameField` — and `src/components/ui/form-field.tsx` — `FormField` — then run Verification steps 1–3 with blank, full-width Unicode, and existing-name inputs.

## Verification

1. Load `codex/provider-name-validation` in a test vault, open **Settings → Copilot → Models (BYOK)**, and add a second OpenRouter provider. Confirm the suggested name is `OpenRouter 2` or the next available suffix.
2. Replace the name with blanks, then with an existing provider name using different case or full-width Unicode characters. Confirm the inline explanation is associated with the field and Save remains disabled.
3. Edit an existing provider: confirm its unchanged name remains valid, another provider's name is rejected, and a trimmed unique name saves. Run `npm test -- src/modelManagement/providers/providerIdentity.test.ts src/modelManagement/ui/dialogs/ConfigureProviderDialog.test.tsx src/modelManagement/ui/components/ProviderDisplayNameField.test.tsx src/components/ui/form-field.test.tsx --runInBand` for the deterministic variants.

### DIFF
--- a/docs/llm-providers.md
+++ b/docs/llm-providers.md
@@ -8,13 +8,13 @@ Copilot includes 16 built-in AI providers, and you can add an unlimited number o
 
 1. Go to **Settings → Copilot → Models (BYOK)**
 2. Add a provider
-3. Enter the provider API key and select the models you want to configure
+3. Enter a unique provider display name, the provider API key, and the models you want to configure
 4. Click Save
 5. Enable models for chat under **Basic → Agents → Quick Chat models**
 
-You can configure multiple providers simultaneously and switch between them by changing the default model.
+You can configure multiple providers simultaneously and switch between them by changing the default model. Provider display names must be unique across Copilot (names that differ only by capitalization or Unicode formatting count as the same name). When you add another instance of a provider, Copilot suggests the next available numbered name, such as `OpenRouter 2`.
 
-Copilot stores provider credentials in this device's Obsidian Keychain. Synced vault settings do not transfer credentials to another device, so enter the required keys separately on each device.
+Copilot stores provider credentials in this device's Obsidian Keychain. The keychain entry uses a readable, safely encoded form of the provider display name, and renaming a provider moves its credential to an entry that matches the new name. Synced vault settings do not transfer credentials to another device, so enter the required keys separately on each device.
 
 ---
 
--- a/src/components/ui/form-field.test.tsx
+++ b/src/components/ui/form-field.test.tsx
@@ -1,0 +1,29 @@
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import { FormField } from "./form-field";
+
+describe("form-field", () => {
+  describe("FormField()", () => {
+    it("associates its label, required indication, and identified alert with the caller's input", () => {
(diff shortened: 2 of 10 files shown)


zeroliu:
@codex review

PR #2818, opened 2026-08-12
Resolve npm audit advisories with semver-compatible lockfile bumps

zeroliu:
Relates to logancyang/obsidian-copilot-preview#303

## Why

The Obsidian community plugin scorecard (https://community.obsidian.md/plugins/copilot) shows "Dependency has a potential vulnerability advisory". A user evaluating Copilot on that page sees a wall of security flags before they ever install the plugin.

Locally on `master`, `npm audit` reports **18 vulnerabilities** (1 critical, 11 high, 3 moderate, 3 low). Nine of them sit in the production tree that actually ships inside `main.js`: `langsmith`, `hono`, `@hono/node-server`, `ws`, `fast-uri`, `ip-address`, `qs`, `uuid`, and `body-parser`.

The most serious of those is [GHSA-fw9q-39r9-c252](https://github.com/advisories/GHSA-fw9q-39r9-c252) against `langsmith`: its vendored lodash `set()` blocks the `__proto__` key but does not block traversal via `constructor.prototype`, so keys an attacker controls reaching `createAnonymizer()` can pollute `Object.prototype` for the whole process.

## What

> **Before:** `npm audit` reports 18 advisories (1 critical, 11 high, 3 moderate, 3 low); `npm audit --omit=dev` reports 9 (5 high, 3 moderate, 1 low).
>
> **After:** both report 0.

No user-visible behavior changes. What changes is which upstream versions the dependency tree pins.

This is the plain `npm audit fix` result — no `--force` and no `package.json` edits. Every bump lands inside the semver ranges `package.json` already declares, so a fresh `npm install` today would already be entitled to resolve these versions; this PR pins them in `package-lock.json`.

| Production package | Before | After |
| --- | --- | --- |
| `langsmith` | 0.5.4 | 0.8.11 |
| `hono` | 4.12.18 | 4.13.2 |
| `@hono/node-server` | 1.19.14 | 1.19.17 |
| `ws` | 8.18.0 | 8.21.3 |
| `ip-address` | 10.2.0 | 10.5.0 |
| `fast-uri` | 3.1.2 | 3.1.5 |
| `body-parser` | 2.2.2 | 2.3.0 |
| `qs` | 6.15.1 | 6.15.3 |
| `uuid` | 11.1.0 | 11.1.1 |

The remaining 36 version changes are dev-only, including the lone critical (`shell-quote` 1.8.2 → 1.10.0) and the Babel, `brace-expansion`, `minimatch`, `picomatch`, `js-yaml`, `form-data`, and `svgo` bumps. The transitive set shrinks by 11 packages.

## Non goal

- Any change to declared dependency ranges in `package.json`. This PR is lockfile-only.
- The scorecard's "Dynamic Code Execution" disclosure — it comes from vendored zod and lodash inside `@anthropic-ai/claude-agent-sdk` and from ajv codegen under `@modelcontextprotocol/sdk`. No semver-compatible bump removes it; it stays tracked in logancyang/obsidian-copilot-preview#303.
- Upgrading `@anthropic-ai/claude-agent-sdk` or `@modelcontextprotocol/sdk` to pull in refreshed vendored code.

## Screenshot

Not applicable — lockfile-only change with no visual surface.

## Risk

**Medium**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Every bump stays inside `package.json`'s declared ranges; the full unit suite (441 suites, 6078 tests) passes on the updated tree |
| A defect would fail CI or be obvious on first use | ✅ | `tsc -noEmit` plus the esbuild production build and the unit suite run in CI; a runtime-dep regression surfaces on the first chat message or agent session |
| A revert fully restores prior state, including persisted data | ✅ | Reverting `package-lock.json` restores the exact prior tree; no data migration and no persisted-format change |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | No first-party code changes; the bumps apply upstream security patches only. See Note on parser semantics |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Lockfile only; `manifest.json`, settings schema, and chat persistence format untouched |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | No source changes |
| No hot-path behavior lacks deterministic coverage | ✅ | Chat, composer, and agent paths are covered by the existing suite, which passes on the new tree |
| No new dependency | ✅ | No direct dependency added; three transitive packages added by upstream bumps, fourteen removed |
| Human-only behavior stays in one feature area and surfaces quickly | ❌ | `langsmith` crosses two `0.x` minors (0.5 → 0.8) and the `hono`/MCP-SDK chain backs Agent Mode, so a regression could appear in either chat or Agent Mode rather than one area |

Review: confirm the diff touches only `package-lock.json` and that `git diff origin/master...HEAD -- package.json` is empty, then run Verification steps 2–4. No first-party code to inspect.

## Verification

1. Check out the branch, run `npm ci`, then `npm audit` and `npm audit --omit=dev`. Both should report 0 vulnerabilities.
2. Run `npm run build` and load the built plugin in a test vault.
3. Send a chat message with any configured provider and confirm a streamed reply renders. This exercises the `langsmith` bump, which rides the LangChain callback path.
4. Start an Agent Mode session and let it complete one tool call. This exercises the `@modelcontextprotocol/sdk` → `hono` / `@hono/node-server` chain.
5. Confirm the composer still applies an edit, exercising the `uuid` bump.

## Note

Several of the dev-tree bumps are parser hardening (`qs`, `js-yaml`, `fast-uri`, `form-data`, `ip-address`). These deliberately tighten what their parsers accept, which is the point of the advisories. Nothing in `src/` calls them directly — they are reached only through transitive callers — but that is the one class of change where a lockfile-only bump can alter behavior rather than just patch a hole, which is why Verification asks for a real chat turn and a real agent tool call rather than trusting CI alone.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -126,19 +126,6 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "node_modules/@ampproject/remapping": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.1.tgz",
-      "integrity": "sha512-lFMjJTrFL3j7L9yBxwYfCq2k6qqwHyzuUl/XBnif78PWTJYyL/dfowQHWE3sp6U6ZzqWiiIZnpTMO96zhkjwtg==",
-      "dev": true,
-      "dependencies": {
-        "@jridgewell/gen-mapping": "^0.3.0",
-        "@jridgewell/trace-mapping": "^0.3.9"
-      },
-      "engines": {
-        "node": ">=6.0.0"
-      }
-    },
     "node_modules/@anthropic-ai/claude-agent-sdk": {
       "version": "0.3.206",
       "resolved": "https://registry.npmjs.org/@anthropic-ai/claude-agent-sdk/-/claude-agent-sdk-0.3.206.tgz",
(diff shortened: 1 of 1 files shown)


zeroliu:
@codex review

PR #2827, opened 2026-08-12
Remove :has() selectors and the mobile CSS overrides they anchored

zeroliu:
Fixes logancyang/obsidian-copilot-preview#295

## Why

The public Obsidian scorecard flags every `:has(` occurrence in the shipped bundle, warning that it "can cause significant performance issues due to broad selector invalidation." The worst offenders anchored on `body`: `body.is-mobile:has(.mobile-navbar)` asks the browser to re-check the whole document on any DOM mutation — on mobile that includes every token appended to a streaming chat message.

Rather than port each selector to an equivalent mechanism, each one was checked against what it actually does today. Three of the four turned out to need no replacement, and chasing the mobile one surfaced a cluster of mobile CSS in the same state.

The built `styles.css` now contains **zero** `:has(` occurrences — and zero `is-mobile` / `is-phone` / `is-tablet` selectors.

## What

| Surface | Before | After |
| --- | --- | --- |
| Agent Configure dialog | `.modal:has(.copilot-config-dialog)` zeroed host padding and hid the native title | `ReactModal` takes a `modalClass` and applies it to `modalEl` in its constructor; the three install modals pass `CONFIG_MODAL_CLASS`, gallery stories declare it via a `modalClass` story parameter |
| Mobile chat / apply view navbar spacing | Three `body:has(.mobile-navbar)` rules inside `@supports` guards | **Deleted** — see below |
| Agent home project rows | Time/actions swap also keyed on a descendant `[data-state=open]` | Swap keys only on hover / focus-within; the variants matched nothing |
| Commands settings table checkbox columns | Cell right padding zeroed via `:has([role=checkbox])` | **Deleted with no replacement** |
| Remaining mobile-only overrides | Four `is-mobile` / `is-phone` blocks | **Deleted** |

### The Configure dialog: the frame is styled by whoever builds it

Obsidian's `Modal` constructor builds `contentEl` inside `modalEl` eagerly, and `open()` attaches the frame to the document *before* `onOpen()` runs. Setting the class in `ReactModal`'s constructor means the frame is already correct the first time it can be painted, with no lifecycle that can desync it from the dialog it describes.

The three install modals also stop setting a native title. They were passing `"Configure X"` only for a stylesheet rule to hide it again; the shell draws its own heading beside the status badge, and core already ships `.modal-title:empty { display: none }`.

Measured against Obsidian 1.13.7's `app.css` plus this branch's built stylesheet:

| | plain modal | `.copilot-config-modal` |
| --- | --- | --- |
| `.modal` padding | 16px | 0 |
| `.modal-header` height | 52px | 0 |
| `.modal-title` display | block | none |
| Divider inset, both edges | 17px | 1px (the frame border — full bleed) |

### The mobile navbar rules: they guarded a placement we don't support

The rule reserved bottom padding so the chat input would clear the floating mobile navbar. It only ever applied inside `.workspace-tab-container` — the main editor area — but `defaultOpenArea` defaults to `VIEW`, which opens the chat in the right sidebar. On mobile that sidebar is a drawer that slides over the whole screen, navbar included, so **in the default placement the rule matched nothing**. It covered only the case where a user switches Default open area to Editor.

The overlap it addressed is real where it applies: verified live against Obsidian 1.13.7 that `floatingNavigation` defaults to `true`, so `is-floating-nav` sits on the body and `.mobile-navbar` computes to `position: fixed` at 52px. But that is a non-default mobile placement, and mobile editor-mode chat is not a surface this plugin supports. The apply-view rules went for the same reason — AI editing is not a supported mobile route.

Removing them retires the last `:has()` without a replacement, so `attachMobileNavbarBodyClass` and its tests go with them.

### The rest of the mobile CSS

With the navbar rules gone, the neighbouring mobile overrides were re-examined and none has a demonstrated job:

- `.is-phone .modal .modal-close-button` z-index, whose own comment ("Maybe ob's error removed the hierarchy … for now") describes a workaround for an Obsidian bug nobody re-confirmed.
- The legacy `body.is-mobile` 4px view-content padding — same unsupported main-editor placement.
- The drawer keyboard rule hiding drawer header and tab options while the soft keyboard is open.
- A `min-width: fit-content` override on Radix popper wrappers.

The keyboard rule was the only one with runtime behind it, so the observer toggling `copilot-keyboard-open` goes too — it existed solely to drive that selector, and was duplicated between `CopilotView` and `attachChatViewLayoutObservers`. The drawer-hide observer in both places is unrelated (it closes Radix popovers when Obsidian hides the drawer, otherwise they re-anchor to a detached node) and stays.

### The table rule was miscentering, not centering

`[&:has([role=checkbox])]:tw-pr-0` is stock shadcn boilerplate, and `Table` has one consumer in this repo. On `TableHead` it matched nothing — every header cell holds text. On `TableCell` it matched the two checkbox cells, which are `tw-text-center` with `tw-mx-auto` on the `Checkbox`; zeroing the right padding widened the content box on one side only, landing the checkbox 4px right of the column axis. Upstream wrote that rule for a left-aligned row-select column. Measured: 4px off-axis before, 0px after.

### Test infrastructure

`__mocks__/obsidian.js` gave `Modal` no DOM at all, so no `ReactModal` subclass was constructible under Jest. It now builds the same element tree Obsidian's constructor does, and `jest.setup.js` polyfills the `addClass` / `setText` element augmentations those paths use.

## Non goal

- The review-gate policy is unchanged: `designdocs/OBSIDIAN_COMMUNITY_REVIEW.md` keeps risky CSS warnings visible and nonblocking; this PR retires the underlying selectors, not the policy.
- No restyling of the Agent Configure dialog or agent home rows — those swaps are pixel-equivalent.
- `:has()` usages originating in Obsidian core or community themes are out of scope.
- No pinned-open trailing state for row popovers; if a dropdown ever returns to the trailing slot, it should drive a data attribute on its row rather than restore descendant probing.
- `.modal.copilot-config-modal .modal-content { padding: 0 }` is kept even though core's `.modal-content` carries no padding — one line of insurance for themes that move it there.

## Screenshot

One deliberate visual change on desktop: the **Commands settings table** checkbox columns shift 4px left, onto the column axis. Everything else on desktop is pixel-equivalent.

The mobile deletions are behavioral by design — see Verification 4–6 for what to look for.

## Risk

**Medium**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Desktop swaps are pixel-equivalent and covered by `ReactModal.test.tsx`; the mobile removals were checked on a physical iPhone (see Verification 6) rather than left to inference |
| A defect would fail CI or be obvious on first use | ✅ | Wrong padding or an unmarked modal shows immediately on the affected surface |
| A revert fully restores prior state, including persisted data | ✅ | CSS and constructor-time classes only; nothing persisted. The two mobile commits are separate and revert cleanly on their own |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | Styling state only |
| No public API, plugin API, message, or on-disk contract changes | ✅ | `ReactModal`'s constructor gains a third optional parameter, so existing subclasses are unaffected |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | One constructor class assignment; one MutationObserver removed |
| No hot-path behavior lacks deterministic coverage | ✅ | `ReactModal.test.tsx` covers title, frame class, and that the class lands before the frame is attached |
| Human-only behavior stays in one feature area and surfaces quickly | ❌ | Spans the Configure dialog, the settings table, and several mobile surfaces |

Review order: `ReactModal.tsx` and `ConfigDialogShell.tsx` first (the only new mechanism), then the three deletions in `tailwind.css`, then the observer removal in `CopilotView.tsx` / `attachChatViewLayoutObservers.ts`.

## Verification

1. Open each agent's Configure dialog (Agent mode → agent selector → Configure) for Claude, Codex, and opencode: dividers run edge to edge, and exactly one heading appears — no duplicate native title.
2. Repeat under a non-default theme, the only case the retained `.modal-content` rule covers.
3. Component gallery → "Agent Mode/Config Dialog Shell": the same full-bleed bands, now driven by the story's `modalClass` parameter.
4. Settings → Copilot commands table: the "In Menu" / "Slash Cmd" checkboxes sit centred (4px left of before).
5. Agent Home landing: hover or keyboard-focus a project row — the relative time still swaps for the Reveal/Edit/Delete cluster.
6. **Done on a physical iPhone**, against an iCloud-hosted vault carrying this exact build: Copilot chat in the right sidebar (the default placement) with the soft keyboard open and closed, the drawer opening and closing with a popover open, modal close-button stacking, and popover widths. No regression found on any of them — including the drawer keyboard case, which is the only removal that touched the default mobile placement.

   `app.emulateMobile(true)` cannot substitute here: desktop has no soft keyboard and Obsidian never sets `--keyboard-height`, so that path is unreachable under emulation.
7. Run `npm run build`, then `grep -c ':has(' styles.css` and `grep -c 'is-mobile\|is-phone\|is-tablet' styles.css`: both are 0.

## Note

The removed `group-has-[[data-state=open]]` variants guarded a trailing dropdown previously replaced by inline action buttons (`AgentProjectRowActions`); no element inside a row carries `data-state` anymore, so the variants matched nothing.


zeroliu:
@codex review

PR #2829, opened 2026-08-12
[Node runtime guards] - Step 1: Guard eager Node imports

zeroliu:
Relates to logancyang/obsidian-copilot-preview#299

## Why

The plugin's community scorecard warns prospective installers about mobile compatibility: `manifest.json` ships `isDesktopOnly: false`, yet dozens of first-party modules statically import Node built-ins that do not exist on phones and tablets. Static imports run at module evaluation, so today the plugin survives mobile load only through an implicit contract of build-time stubs tolerating `undefined` modules. Two of the riskiest cases load eagerly: `rendererEventsShim` (`node:events`) evaluates from `main.ts` on every platform, and `detectBinary` builds a promisified `execFile` at module scope — on mobile that is a property read off an `undefined` module, which would kill the whole plugin at load the moment the eager graph reaches it.

## What

Phase 1 of the issue's one-subsystem-per-PR plan: the renderer events shim and the binary-detection/path utilities reachable from the settings UI now load Node built-ins through a shared, desktop-gated call-time accessor.

| Condition | Before | After |
| --- | --- | --- |
| Module evaluation on mobile | Node built-ins required at load; safety depends on build-time stubs tolerating `undefined` modules | No Node access at evaluation time, by construction |
| Desktop call paths | Static imports resolve at load | The same modules resolve through the same `require` at first call — behavior unchanged |
| Node-dependent call on a non-desktop runtime (e.g. binary detection) | `TypeError` from an `undefined` module | Clear error naming the unavailable built-in |

Converted imports: `rendererEventsShim` (`node:events`), `binaryPath` (`node:fs`, `node:os`), `detectBinary` (`node:child_process`, `node:util`, `node:fs`, `node:path`), `nodeToolBinDirs` (`node:path`), `appPaths` (`node:path`). These converted files now produce no `obsidianmd/no-nodejs-modules` warnings; the remaining warnings stay visible for the follow-up migration.

## Non goal

- The Agent Mode ACP/coding-agent backends bulk (~40 imports), the context conversion cache, and the Symposium agent handoff — later phases of #299.
- Making Agent Mode or any other desktop-only feature work on mobile — guards make them fail safe, not function.
- Converting the three `buffer` polyfill imports (`src/utils.ts`, `src/utils/base64.ts`, `src/LLMProviders/BedrockChatModel.ts`) — mobile-safe by design and deliberately retained.
- Removing the esbuild `process` banner shim or the externals list — bundled dependencies still rely on them.
- Flipping `isDesktopOnly` in `manifest.json` or changing which features are desktop-gated.

## Screenshot

Not applicable — no visual change; the conversions preserve desktop behavior and only move module resolution from load time to call time.

## Risk

**Low**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Each converted suite gains an eval-time isolation test (module import must not touch Node built-ins) plus non-desktop guard tests; existing desktop-path tests still pass unchanged |
| A defect would fail CI or be obvious on first use | ✅ | The new tests run in `npm run test`; a desktop resolution defect also surfaces on the first binary detection or Agent Mode start |
| A revert fully restores prior state, including persisted data | ✅ | No persisted data, settings schema, or migration is touched |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | The binary-name validation in `detectBinary` is byte-identical |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Every exported signature is unchanged; only import mechanics move |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Requires stay synchronous; only their timing moves from module load to first call |
| No hot-path behavior lacks deterministic coverage | ✅ | Desktop resolution, the live-`EventEmitter` patch identity, and every guard branch are unit-tested |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | A wrong guard shows up immediately as a plugin load error or a binary-detection error in Agent settings |

Review: skim `Why` and `What`, spot-check `requireNodeModule` in `src/utils/desktopRuntime.ts`, run Verification step 3, and confirm CI is green.

## Verification

1. Build and install the branch in a desktop test vault, then reload Obsidian — the plugin loads with no console errors.
2. In Settings → Copilot → Agent settings, trigger binary detection for a coding agent (e.g. Claude Code) — the detected path appears exactly as on `master`, with the home directory collapsed to `~`.
3. In the developer console run `app.emulateMobile(true)` and reload — the plugin still loads cleanly (this is the stricter case where Node modules are stubbed to null while `Platform.isDesktopApp` stays true). Run `app.emulateMobile(false)` to restore.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -396,21 +396,16 @@ export default [
   // Agent Mode: backends spawn subprocesses (ACP) and the in-process Claude
   // SDK uses node:async_hooks. The plugin runs in Electron renderer where
   // these modules are available; the desktop-only Agent Mode is also gated by
-  // `Platform.isMobile` at runtime in main.ts.
-  // detectBinary / binaryPath / nodeToolBinDirs / rendererEventsShim are
-  // sibling utilities pulled in by agent-mode wiring and share the same
-  // Electron-renderer assumptions. The Symposium handoff consumer is likewise
-  // desktop-only, but remains in the host publishing layer. openVaultPath
-  // lazy-requires node:fs behind a desktop vault-base guard, mirroring
-  // opencodeLog.
+  // `Platform.isMobile` at runtime in main.ts. The Symposium handoff consumer
+  // is likewise desktop-only, but remains in the host publishing layer.
+  // openVaultPath lazy-requires node:fs behind a desktop vault-base guard,
+  // mirroring opencodeLog. Modules converted to call-time
+  // `requireNodeModule()` guards (see src/utils/desktopRuntime.ts) need no
+  // exemption; the goal is to shrink this list to empty as remaining
+  // subsystems convert (logancyang/obsidian-copilot-preview#299).
   {
     files: [
       "src/agentMode/**",
-      "src/utils/appPaths.ts",
-      "src/utils/detectBinary.ts",
-      "src/utils/binaryPath.ts",
-      "src/utils/nodeToolBinDirs.ts",
-      "src/utils/rendererEventsShim.ts",
       "src/utils/issueReport.ts",
       "src/utils/opencodeLog.ts",
       "src/utils/openVaultPath.ts",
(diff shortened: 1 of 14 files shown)


zeroliu:
@codex review

PR #2839, opened 2026-08-13
Rebuild an ACP chat's transcript when it is reopened from history

Emt-lin:
Fixes logancyang/obsidian-copilot-preview#308

## Why

Closing an opencode or codex chat and reopening it from Chat History shows the empty landing page — the greeting and the model picker, as if the chat had never happened — instead of the conversation. Nothing in the UI brings the messages back. Two users have reported it, one describing a chat vanishing after OpenCode restarted mid-session, with no way to reach it again through history.

The conversation itself survives: the agent still has it, and the next prompt continues from where it left off. Only the plugin's view of it is gone, which makes the loss look total to the user while the chat is in fact intact.

A chat saved as a markdown note is unaffected, because it restores from the note rather than from the agent. The bug therefore only appears with **Autosave chat** off, which is why it can go unnoticed.

## What

> **Before:** close an opencode or codex chat, reopen it from Chat History, and the landing page appears with no messages.
>
> **After:** the same chat reopens showing every message it had when it was closed, in order, with the text it had.

| Reopening a chat that… | Before | After |
| --- | --- | --- |
| had several turns | Landing page, no messages | The conversation as it was left |
| is then given a new prompt | The reply lands in an empty view | The reply appends after the restored messages |
| is longer than 32 replayed frames | Landing page | The whole conversation |
| had a note attached to a prompt | Nothing restored | The user's bubble shows what they typed, not the context block wrapped around it |
| used a tool mid-answer | Landing page | The assistant's text from before and after the tool call, as one message |
| had a prompt whose turn was cancelled before the agent replied | Landing page | That prompt and the one after it restore as two separate messages, not one run-on bubble |
| is reopened at all | Every restored message is dated to the moment of reopening | Restored messages keep the time they were sent, or none when the source has none |

Restored messages carry sender and text only, the same fidelity the Claude and markdown restore paths already offer.

One live-path behavior changes as well: an opencode chat no longer loses its title when the agent echoes the user's prompt. That echo previously reached the translator's unknown-discriminant fallback, which reports a titleless session update, and opencode is the backend whose pushed titles are trusted.

## Non goal

- Restoring tool cards, thinking, or plan entries. Every restore path in the plugin is sender + text only, and the message trail takes over an assistant message's whole body as soon as it carries any parts, so a partial restoration would hide the answer itself. Unifying trail fidelity across the Claude, markdown, and ACP paths is separate work.
- Changing how Claude-backed or markdown-backed chats restore.
- The rest of the title-clearing defect tracked in logancyang/obsidian-copilot-preview#307. This PR closes the one path it had to touch — the user-prompt echo described above — and leaves the root fallback, and every other update that lands on it, to that issue.

## Screenshot

Not attached. Before-and-after captures were taken during end-to-end verification of each scenario below, but they live on the verification machine and were not uploaded through the GitHub editor. The difference is reproducible in under a minute through Verification steps 2 and 3: an empty landing page before, the restored conversation after.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | `replayTranscript.test.ts` covers every grouping rule and `AcpBackendProcess.test.ts` covers the accumulator's install, return, and retirement |
| A defect would fail CI or be obvious on first use | ✅ | A wrong transcript shows as a blank, glued, or shattered conversation on the first reopen, and the grouping tests fail on any change to the rules |
| A revert fully restores prior state, including persisted data | ✅ | Nothing is written to disk; the transcript is transient and a revert returns to the previous empty-restore behavior |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | No credential, permission, or request-authentication path is touched |
| No public API, plugin API, message, or on-disk contract changes | ❌ | `LoadSessionOutput` gains an optional `transcript`; additive and plugin-internal, but every backend implements this interface |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | An accumulator is installed before the `session/load` RPC and retired in its `finally`, and the RPC response becomes the barrier after which a replayed conversation frame is refused |
| No hot-path behavior lacks deterministic coverage | ✅ | The notification-routing short-circuit runs for every `session/update` and is covered; the one branch without observed data is named in Note |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Confined to Agent Mode session restore, and visible on the next reopen |

Review: inspect `src/agentMode/acp/AcpBackendProcess.ts` — `loadSession` and `routeSessionUpdate` — line by line for the accumulator lifecycle and the replay barrier, then `src/agentMode/acp/replayTranscript.ts` — `consumeReplayUpdate` and `flushCurrent` — for the grouping rules, and `src/agentMode/session/types.ts` — `LoadSessionOutput` — for the interface addition. Then run Verification steps 4 through 8, which exercise continuing a restored chat, a turn that only thinks, a turn that calls a tool, a prompt with a note attached, and a cancelled turn followed immediately by another prompt.

## Verification

Turn off **Settings → Copilot → Autosave chat** before starting. With it on, chats restore from their markdown note and the bug cannot appear.

1. Check out the branch, build it, and reload Obsidian so the new bundle is running.
2. Open Agent Mode on codex, send `Hello`, wait for the reply, then send `Goodbye` and wait again. Close the chat with the ✕ in the tab.
3. Open Chat History and click that chat. Expect all four messages in their original order and wording. Expect no `dropping … — no target` or `pending buffer full` in the developer console.
4. In the reopened chat, send `1+1=?`. Expect the reply to append after the four restored messages, with none of them duplicated or reordered.
5. Repeat steps 2 and 3 on opencode.
6. On codex, ask a question that produces visible thinking, such as comparing B-trees and LSM trees across write, read, and space amplification. Close and reopen. Expect one assistant message holding the final answer, with no thinking text in it.
7. On codex, ask it to list the vault root with its directory tool, and confirm a tool card appears. Close and reopen. Expect the assistant's text from before and after the tool call in a single message; the tool card is not restored. Then attach a note with `@`, send `hi`, close, and reopen. Expect the user's bubble to read exactly what was typed, with no `<copilot-context>` or `<user-message>` tags.
8. On opencode, send a long prompt such as `Write a detailed 800-word history of the printing press.` and press Stop before any reply text appears. Send `What is 2+2?` and let it answer. Close and reopen. Expect the two prompts as two separate bubbles, not one bubble reading `Write a detailed 800-word history of the printing press.What is 2+2?`.

## Note

One rule has no observed data behind it: a `tool_call` or `plan` frame marks the agent as the speaker, which only changes the restored output when a turn that runs a tool and says nothing sits between two prompts. No captured replay contains that shape — opencode always surrounds tool frames with thinking, and codex does not replay them at all — so that branch rests on the live model, where a tool part alone already counts as assistant activity, plus unit tests, rather than on a reproduced case.

A second branch is covered by unit tests only. A user chunk that carries no `messageId` starts a new prompt rather than continuing the previous one, which is what keeps two prompts apart on a backend that omits the field. Captured replays contain 7 such chunks, all from codex, so the input is real; but the codex build available for end-to-end verification has since started sending ids, so the id-less path itself was exercised in tests rather than through the UI.


Emt-lin:
@codex review

src/agentMode/acp/replayTranscript.ts
```diff
@@ -0,0 +1,143 @@
...
+  state.messages.push({
+    // Positional like the Claude loader's `claude-loaded-N`: this is a display
+    // identity, unrelated to any id the backend put on the wire.
+    id: `acp-loaded-${state.messages.length}`,
+    sender: current.sender,
+    message,
+    isVisible: true,
+    // The replay carries no original send time, and stamping "now" would show
+    // every restored message as if it had just been sent.
+    timestamp: null,
```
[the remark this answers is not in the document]
Emt-lin: Confirmed, and it reached further than the display. `AgentMessageStore.loadMessages` was replacing the null with the load time because the store's own `StoredAgentMessage.timestamp` demanded a non-null value — so the intent in this file never survived the handoff.

The persistence consequence is not limited to a manual save either: a markdown chat attaches its autosave subscription before its messages load, so the notify that follows `loadDisplayMessages` queues a rewrite of the original note, turning an "Unknown time" entry into whenever the user reopened it. The ACP replay this PR adds carries no times at all, so every restored message would have taken that stamp.

Fixed at the store in 059c6c6e: `StoredAgentMessage.timestamp` is widened to `FormattedDateTime | null`, matching the public `AgentChatMessage` which has always allowed null, and `loadMessages` passes a loaded time through unchanged. Live messages are unaffected — `addMessage` keeps its own fallback. Consumers already handled the absence: the UI reads `timestamp?.display`, markdown serializes "Unknown time", and filename and index derivation use `?.epoch ?? now`. Added round-trip tests in `AgentMessageStore.test.ts` for both the null and the dated case.

Thanks — this one was a real gap between what the comment claimed and what shipped.

Emt-lin:
@codex review

src/agentMode/session/promptEnvelope.ts
```diff
@@ -0,0 +1,37 @@
...
+ * note in place of an image it cannot send, and the transcript hands both
+ * blocks over as one string.
+ *
+ * @param content - Stored prompt text, wrapped or not.
+ * @returns What the user typed, or `content` unchanged when there is no wrapper
+ *   (prompts sent without attached context are not wrapped).
+ */
+export function stripUserMessageWrapper(content: string): string {
+  const start = content.indexOf(OPEN_TAG);
+  const end = content.lastIndexOf(CLOSE_TAG);
```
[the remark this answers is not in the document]
Emt-lin: Confirmed and fixed in 724361ea. `buildContextEnvelope` inlines selected excerpts verbatim, line by line, so the context in front of the wrapper really can be any note the user attached — including one that mentions the tag.

I compared the shapes before changing anything:

| Stored prompt contains | before this PR | on this branch | now |
| --- | --- | --- | --- |
| an excerpt mentioning `<user-message>` | leaks the context tail | leaks the context tail | the typed text |
| an excerpt with both tags | returns the excerpt's inner text | leaks the real wrapper too | the typed text |
| `</user-message>` typed by the user | truncates at it | keeps it | keeps it |
| `<user-message>` typed by the user | keeps it | keeps it | loses what precedes it |

So the second row was a regression this branch introduced over the previous regex, which you caught.

The fix searches back from the closing tag for the opening one, since the envelope always ends the prompt. That trades the last row for the first two. Excerpt content is any note the user attaches; typing the opening tag into the composer is deliberate, so the larger surface wins. The tags are unescaped and this cannot be made exact without changing the wire format, which is out of scope here — the residual is written into the function's doc comment rather than left implicit. Added tests for both excerpt shapes.

Emt-lin:
@codex review

src/agentMode/acp/replayTranscript.ts
```diff
@@ -0,0 +1,143 @@
...
+    // that opened it already marked the boundary; letting the update mark one
+    // too would split a user message that a late arrival landed in the middle
+    // of.
+    case "tool_call_update":
+      return true;
+    default:
+      return false;
+  }
+
+  if (state.current?.sender !== sender) flushCurrent(state);
```
logancyang: **Blocking: preserve separate user turns.**

This boundary changes only when the sender changes. If a turn is cancelled, refused, or fails before the agent emits any frame, the replay has two consecutive `user_message_chunk`s with different `messageId`s and this concatenates them into one bubble (`"cancelled promptnext prompt"`). That violates the PR contract to restore every message in order and with its original text. ACP 0.20.0 defines a changed `messageId` as a new message.

Please track the current user `messageId` and flush when a different non-null user ID arrives. Keep ignoring assistant ID changes so one live assistant turn still restores as one bubble, and add a regression test with two consecutive user chunks carrying distinct IDs.

src/agentMode/session/types.ts
```diff
@@ -512,6 +512,7 @@ export interface PlanSummary {
  * model/mode/configOption deltas, so consumers don't see raw catalog shapes.
  */
 export type SessionUpdate =
+  | { sessionUpdate: "user_message_chunk"; content: PromptContent; messageId?: string }
```
logancyang: **Blocking: keep the ACP echo out of the session-domain contract.**

`user_message_chunk` is an ACP prompt echo, not a backend-neutral session event. `AgentSession` has no behavior for it and reaches the generic ignore path. The load collector already consumes the raw frame before translation, so widening `SessionUpdate` and adding a translator case only creates a new domain concept to discard it later.

Please have `acpNotificationToEvents` return no events for a live `user_message_chunk`, then remove this union member and the matching translation case. That preserves the OpenCode title fix with a smaller contract and no unused event variant.

logancyang (changes requested):
@Emt-lin, requesting changes for two blocking issues in the replay design. The collector is the right boundary, but these need fixing before merge so restored history preserves message identity and the backend-neutral contract stays minimal.

src/agentMode/acp/replayTranscript.ts
Emt-lin: Done in 4dbba41c — with one deliberate deviation from the literal ask, which I want to flag rather than bury.

You asked to flush when a different **non-null** user id arrives. I implemented exactly that first, then found it still reproduces your failure shape on one of the two ACP backends: codex replays its prompts with **no `messageId` at all**. With "flush only on a different non-null id", two id-less prompts both fail the check and concatenate — the same `"cancelled promptnext prompt"` bubble, just on a different backend.

Evidence from real `session/load` replays captured locally: 85 user chunks across 72 bursts, 78 carrying an id and 7 carrying none. All 7 id-less ones are codex; opencode always sends `msg_*`. The plugin has no minimum-version gate for codex (`CodexBackend.ts:44` keeps a legacy path), so that input is reachable today.

So the rule is inverted rather than additive: a user chunk **continues** the current prompt only when it repeats that prompt's id, and starts a new one otherwise. That covers both a different id and an absent id. The joining branch stays because, as ACP 0.20.0 says, one message may arrive as several chunks sharing an id.

Assistant ids are still ignored, as you asked — `resolveContentTarget` puts a whole turn in one placeholder live, so honouring them would invent bubbles that never existed.

Tests: `separates two prompts that carry different message ids` is the regression you asked for. `separates two prompts that the agent replayed without any message id` covers the id-less variant. I also removed two older cases that encoded the previous merge-by-default assumption. I mutation-checked the condition: dropping either guard turns a test red.

Verification: the exact shape you described — two consecutive user chunks, no agent frame between them, distinct ids — now restores as two bubbles end to end in the GUI on both backends. One gap stated plainly: the **id-less** branch is covered by unit tests only, because this machine's codex-acp has since upgraded to a build that does send ids, so I could not reach that branch through the GUI.

src/agentMode/session/types.ts
Emt-lin: Done in 84ee0f48, exactly as described: `acpNotificationToEvents` returns no events for a `user_message_chunk`, and both the `SessionUpdate` union member and the translator case are gone. You were right that it was a domain concept invented only to be discarded — nothing read it.

One implementation detail worth stating, since it is the reason the OpenCode title fix survives: the frame is dropped at the **entry** to `acpNotificationToEvents`, before `acpUpdateToSessionUpdate` runs. Translating first and dropping after would send it through the unknown-discriminant fallback, which reports a titleless session update and clears the label on a backend whose titles are trusted.

Net effect on this file is a deletion. The test now asserts the empty result instead of an event.

Emt-lin:
@codex review

logancyang (approved):
Both blocking findings are resolved. The updated replay boundaries preserve separate user turns, and ACP prompt echoes no longer widen the backend-neutral session contract.

logancyang (approved):
Re-approved after updating from master. The new head is only the merge-from-master commit; the reviewed PR changes are unchanged.

PR #2859, opened 2026-08-15
[Node runtime guards] - Step 2: Guard remaining Node imports

zeroliu:
Relates to logancyang/obsidian-copilot-preview#299

## Why

PR #2829 removes the riskiest eager Node imports, but the active `obsidianmd/no-nodejs-modules` rule still found Node built-ins throughout Agent Mode, the context cache, the Symposium handoff, and desktop report/path adapters. Those imports resolved during module evaluation even though their call paths were desktop-only, so mobile loading depended on the bundle tolerating modules that do not exist there.

The rule also treated three imports of the browser-compatible `buffer` package as Node built-ins because its bare package name matches Node's built-in module name. That left false-positive warnings in the review output after the unsafe imports were migrated.

## What

Desktop-only Node modules now resolve at call time behind the shared desktop-runtime guard. Mobile bundle evaluation no longer loads those built-ins, while desktop subprocess, filesystem, binary-management, report, and Symposium behavior remains unchanged.

The browser `buffer` package now uses its explicit package entry point. Review tooling can distinguish it from Node's built-in `buffer`, so `obsidianmd/no-nodejs-modules` reports no warnings without disabling or suppressing the rule.

An error-level `copilot/no-direct-node-imports` rule prevents direct runtime access to Node built-ins from returning. It covers imports, re-exports, dynamic imports, and literal `require()` calls, and directs contributors to `requireNodeModule()` for runtime access or a Node import type query for types. Tests and mocks remain exempt because they execute under Jest rather than the Obsidian renderer.

## Non goal

- Making subprocesses, native filesystem access, or Agent Mode function on mobile; unsupported calls still fail at the shared desktop guard.
- Changing binary discovery, installation, authentication, session persistence, Symposium validation, report-bundle behavior, or binary encoding behavior.
- Changing dependencies, build externals, the `process` banner shim, or plugin manifests.
- Addressing unrelated pre-existing Obsidian review warnings.

## Screenshot

Not applicable — this changes internal runtime module resolution and review output, with no visual change.

## Risk

**Medium**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Existing unit coverage exercises the migrated adapters and browser-compatible byte conversions |
| A defect would fail CI or be obvious on first use | ✅ | Type checking, lint, mobile bundle evaluation, and unit tests run in CI |
| A revert fully restores prior state, including persisted data | ✅ | No settings, files, schemas, or migrations change |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | Auth and path-validation behavior are unchanged; only module-resolution timing changes |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Exported signatures and serialized formats are unchanged |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Existing subprocess, installer, and persistence flows retain their ordering and control flow |
| No hot-path behavior lacks deterministic coverage | ❌ | Live OS-specific spawn, download, and archive behavior cannot be reproduced across every supported desktop platform in CI |
| No new dependency | ✅ | The browser `buffer` package was already a direct dependency |
| Human-only behavior stays in one feature area and surfaces quickly | ❌ | Desktop module access spans several internal Agent Mode and utility workflows |

Review: inspect the call-time module access in `src/agentMode/backends/opencode/OpencodeBinaryManager.ts`, the explicit browser package imports in `src/utils/base64.ts`, `src/utils.ts`, and `src/LLMProviders/BedrockChatModel.ts`, and the enforcement in `eslint.config.mjs`, then follow Verification.

## Verification

1. Run `npm run lint` and confirm it reports no `obsidianmd/no-nodejs-modules` warnings and rejects the direct-Node negative fixtures with `copilot/no-direct-node-imports`.
2. Run `npm run review:obsidian` and confirm the review regression fixtures pass.
3. Run `npm run test:mobile-load` and confirm the production bundle evaluates while Node built-ins are unavailable.
4. On desktop, start a configured Agent Mode backend and confirm it launches and responds normally.

## Note

Node's module detector classifies the bare name `buffer` as built-in even when esbuild resolves the installed browser polyfill. Naming the dependency's package entry point removes that ambiguity while keeping the review rule enabled globally.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -4,6 +4,65 @@ import reactHooks from "eslint-plugin-react-hooks";
 import tailwind from "eslint-plugin-tailwindcss";
 import boundaries from "eslint-plugin-boundaries";
 import globals from "globals";
+import { isBuiltin } from "node:module";
+
+const NODE_IMPORT_GUIDANCE =
+  "Use requireNodeModule() from '@/utils/desktopRuntime' for runtime access; use an import(\"node:...\") type query when only a type is needed.";
+
+const noDirectNodeImportsRule = {
+  meta: {
+    type: "problem",
+    schema: [],
+    messages: {
+      directNodeImport: `Do not access Node.js built-in module "{{moduleName}}" directly. ${NODE_IMPORT_GUIDANCE}`,
+    },
+  },
+  create(context) {
+    const reportIfBuiltin = (node, moduleName) => {
+      if (typeof moduleName === "string" && isBuiltin(moduleName)) {
+        context.report({
+          node,
+          messageId: "directNodeImport",
+          data: { moduleName },
+        });
+      }
+    };
+
+    return {
+      ImportDeclaration(node) {
+        reportIfBuiltin(node, node.source.value);
+      },
+      ExportNamedDeclaration(node) {
+        if (node.source) {
+          reportIfBuiltin(node, node.source.value);
+        }
+      },
+      ExportAllDeclaration(node) {
+        reportIfBuiltin(node, node.source.value);
+      },
+      ImportExpression(node) {
+        if (node.source.type === "Literal") {
+          reportIfBuiltin(node, node.source.value);
+        }
+      },
+      CallExpression(node) {
+        if (
+          node.callee.type === "Identifier" &&
+          node.callee.name === "require" &&
+          node.arguments[0]?.type === "Literal"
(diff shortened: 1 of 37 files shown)


zeroliu:
@codex review

Issue #293, opened 2024-02-15
New Chat does not stop streaming


logancyang:
**Describe the bug**
Right now the new chat button doesn't disrupt ongoing streaming

**To Reproduce**
Click new chat when it's streaming

**Expected behavior**
New chat should disrupt streaming like Stop Generation, and then create a clean slate

logancyang:
cc @petery789

PR #2862, opened 2026-08-15
Build DOM through Obsidian's createEl helpers

zeroliu:
Fixes logancyang/obsidian-copilot-preview#317

## Why

The community listing at [community.obsidian.md/plugins/copilot](https://community.obsidian.md/plugins/copilot) carries the line "Uses `document.createElement` instead of Obsidian's `createEl` helpers". Running the official rule behind that line — `npm run review:obsidian:source`, which is `eslint-plugin-obsidianmd` — reports 30 findings in first-party source today.

logancyang/obsidian-copilot-preview#293 records the plugin-realm migration as finished, but its success criterion is `grep -rn "document.createElement" src/`. That search only sees the literal global receiver. Every remaining site reaches its `Document` through a variable, so it passes the grep and still trips the rule: `doc.createDocumentFragment()` and `doc.createElement("span")` in `linkInlineCitations`, `getEditorDocument(editor).createElement("span")` in each Lexical pill's `createDOM`/`exportDOM`, `root.doc.createElement("span")` in the Symposium task-checkbox and internal-link normalizers, `host.doc.createElement("div")` in the Quick Ask overlay, and so on.

Those variable receivers are not incidental — each one is a deliberate popout fix, resolving the document that will host the element rather than trusting whichever window is focused. The obvious migration loses that: `obsidian.d.ts` declares the helpers as ambient globals and on `Node`, never on `Window`, so the rule's own suggestion `someDoc.win.createDiv()` does not typecheck, and the ambient global always builds in the main window's document. Obsidian does install a per-window copy — it re-evaluates its `globalEnhance()` bootstrap inside every popout — so only the declaration was missing.

## What

Every first-party construction site now goes through Obsidian's helpers, and the produced DOM is unchanged.

| Site shape | Before | After |
| --- | --- | --- |
| Element appended to a known parent | `doc.createElement("div")`, set `className`, `parent.appendChild(…)` | `parent.createDiv("…")` — one call, still in the parent's document |
| Element that must stay detached | `someDoc.createElement("span")` | `someDoc.win.createSpan("…")` — the copy Obsidian binds to that window |
| Fragment | `doc.createDocumentFragment()` | `doc.win.createFragment()` |

`typings/global.d.ts` declares the four helpers this codebase uses on `Window`, which is what makes the detached cases both type-safe and popout-correct. Five Lexical pills carried a `createDOM` override byte-identical to the base class's; they are deleted rather than migrated.

`obsidianmd/prefer-create-el` becomes an error for shipped source, so a native DOM call fails `npm run lint` instead of drifting back onto the listing. Tests keep the native API — jsdom has no Obsidian helpers — and `jest.setup.js` gains a polyfill mirroring `globalEnhance`, alongside the `Node.doc` / `Node.win` polyfills already there.

## Non goal

- The 40 `document.createElement` occurrences in the shipped `main.js` that come from bundled dependencies (React DOM, Radix, style-loader). They are upstream code.
- The guest-realm snippet in `getSelectedMarkdown()`, which runs inside the Web Viewer's page where Obsidian's helpers do not exist. logancyang/obsidian-copilot-preview#293 owns that one and it stays native.
- Widening #293's grep criterion to the lint rule. Recorded in #317 as follow-up.
- Migrating `createElement` calls inside test files.

## Screenshot

Not applicable — this PR changes how elements are constructed, not what is rendered. Every affected surface produces the same tag, classes, attributes, and text as before; the Verification path below is how a reviewer confirms that visually.

## Risk

**Medium**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Construction-only refactor; tests cover the chat citation link, the `markdown-rendered` text segment, the pill DOM contract, the image picker's `accept`/`multiple`, the Symposium document pipeline, and the gallery colour probe |
| A defect would fail CI or be obvious on first use | ✅ | A missing helper throws on the first call, so the surface fails to render rather than degrading quietly |
| A revert fully restores prior state, including persisted data | ✅ | No migration and no persisted write; the diff only changes how in-memory elements are built |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | No such surface is touched |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Pill `exportDOM` output feeds Lexical's own clipboard round-trip and keeps identical `data-lexical-*` attributes and text |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Synchronous element construction only |
| No hot-path behavior lacks deterministic coverage | ✅ | Chat message rendering is the hot path and its changed sites are covered in `ChatSingleMessage.test.tsx` |
| Human-only behavior stays in one feature area and surfaces quickly | ❌ | The Quick Ask overlay, the custom-command modal container, and the command-UI markdown preview have no automated coverage, and the change spans several feature areas rather than one |

Review: inspect `typings/global.d.ts` — the `Window` interface block — to confirm the declared signatures match `obsidian.d.ts`'s ambient `createEl` / `createDiv` / `createSpan` / `createFragment`, then inspect `src/components/quick-ask/QuickAskOverlay.tsx` (`getOverlayRoot`, `mountOverlay`), `src/commands/CustomCommandChatModal.tsx` (`open`), and `src/components/command-ui/markdown-preview.tsx` (`MarkdownPreview`) line by line, since those three carry no test. Then run Verification steps 4–6, which exercise exactly those surfaces, plus step 7 in a popout window.

## Verification

1. Build and load this branch in a test vault (`npm run test:vault`), then reload Obsidian.
2. Open Copilot chat and send a message that returns sources. Confirm the inline `[1]` marker renders as a clickable citation link and opens the cited note.
3. In the composer, add a note pill with `[[`, a folder pill with `{`, and a URL pill by pasting a link. Confirm each renders as a pill, then select the composer contents and copy — pasting back into the composer restores the same pills.
4. Click the composer's image attach button. Confirm the file dialog opens, then cancel it and confirm the composer still accepts clicks and typing.
5. Select text in a note and run a custom command from the Copilot menu. Confirm the modal opens over the note and its markdown preview renders.
6. Trigger Quick Ask in the editor. Confirm the overlay appears anchored to the cursor and closes cleanly.
7. Move the chat to a popout window (right-click the tab → Open in new window) and repeat steps 2 and 3 there, with the main window focused. Confirm citations and pills still render in the popout.
8. Publish a note to Symposium. Confirm the published page keeps task checkboxes, internal links, and embedded images.

### DIFF
--- a/dev/gallery/audit.ts
+++ b/dev/gallery/audit.ts
@@ -226,9 +226,8 @@ function hasVisibleUnsupportedColor(value: string): boolean {
  */
 export function resolveObsidianColorTokens(doc: Document): ReadonlySet<string> {
   const colors = new Set<string>();
-  const probe = doc.createElement("span");
+  const probe = doc.body.createSpan();
   probe.hidden = true;
-  doc.body.append(probe);
 
   for (const source of [doc.documentElement, doc.body]) {
     const styles = source.win.getComputedStyle(source);
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -64,6 +64,38 @@ const copilotLintPlugin = {
   },
 };
 
+// obsidianmd ships its rules as warnings, and a warning stream nobody gates on
+// is how 30 `prefer-create-el` violations reached the plugin's community listing
+// unnoticed. Promote every rule the codebase already satisfies to an error so
+// the next one fails `npm run lint` in CI on the PR that introduces it. Rules
+// listed here keep the recommended severity because they have a known backlog or
+// a deliberate override; each needs its own reason, not a blanket exemption.
+const OBSIDIANMD_UNRATCHETED = new Set([
+  // Declarative settings migration, tracked by logancyang/obsidian-copilot-preview#297.
+  "settings-tab/prefer-setting-definitions",
+  // Turned off below; promoting here would resurrect them.
+  "ui/sentence-case",
+  "platform",
+  // Configured below with a project-specific message payload.
+  "rule-custom-message",
+]);
+
+// Only raise the severity of rules the recommended config already turns on.
+// Rules it leaves off are its own judgement call; adopting one is a separate
+// decision with its own cleanup, not something this ratchet should smuggle in.
+const OBSIDIANMD_RATCHET = Object.fromEntries(
+  Object.entries(
+    (Array.isArray(obsidianmd.configs.recommended)
+      ? obsidianmd.configs.recommended
+      : [obsidianmd.configs.recommended]
+    ).reduce((rules, block) => Object.assign(rules, block.rules), {})
+  )
+    .filter(([id]) => id.startsWith("obsidianmd/"))
+    .filter(([, severity]) => severity !== "off" && severity !== 0)
+    .filter(([id]) => !OBSIDIANMD_UNRATCHETED.has(id.slice("obsidianmd/".length)))
+    .map(([id]) => [id, "error"])
+);
+
 const restrictedSourceImports = [
   {
     selector:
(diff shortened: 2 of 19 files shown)


zeroliu:
@codex review

PR #2863, opened 2026-08-15
[LangChain deprecation cleanup] - Step 2: Drop OpenRouter kwargs tool_calls

zeroliu:
Relates to logancyang/obsidian-copilot-preview#300

## Why

Stacks on #2823 and must merge after it.

`@typescript-eslint/no-deprecated`, added to the Obsidian review gate by #2823, still reports two findings in `src/LLMProviders/ChatOpenRouter.ts`. Both are `additional_kwargs.tool_calls`, which LangChain marks deprecated with "Use \"tool_calls\" field on AIMessages instead".

The adapter writes that container on the way out and reads it on the way in, and both halves are now duplicates of a first-class surface that already carries the same data. While streaming, `buildMessageChunk` emits every tool-call fragment as `tool_call_chunks` — the shape LangChain aggregates into `AIMessageChunk.tool_calls` — and then copies the same raw fragments into `additional_kwargs`. On the request side, #2823 added a branch that serializes `AIMessage.tool_calls` into the OpenAI wire format; because `AIMessage.isInstance` is true for an `AIMessageChunk` as well, an aggregated streamed turn already matches that branch, and the `additional_kwargs` branch below it never runs for anything the plugin produces.

## What

Quick Chat's tool-calling behavior on OpenRouter, LM Studio, and Copilot Plus is unchanged: the assistant turn is still sent with its `tool_calls`, immediately followed by the matching `role: "tool"` result, and reasoning and plain streaming still render. Nothing consumes `additional_kwargs.tool_calls` any more, so the plugin stops producing it.

Whole-`src/` `no-deprecated` findings go from 5 to 3 (master is merged through the base branch, and master has since cleared the `keyCode` and lucide `Youtube` findings).

> **Before:** A streamed OpenRouter assistant turn carried its tool calls twice — as `tool_call_chunks`, and again as raw OpenAI fragments under the deprecated `additional_kwargs.tool_calls`.
>
> **After:** The turn carries them once, as `tool_call_chunks`. The bytes sent to the provider are the same.

## Non goal

- No wider `ChatOpenRouter` restructuring — the reasoning members of `additional_kwargs` (`delta.reasoning`, `reasoning_details`) are not deprecated, `ThinkBlockStreamer` reads them, and they stay.
- No `invalid_tool_calls` fallback. When a model streams tool-call arguments that never parse as JSON, LangChain leaves `tool_calls` empty and fills `invalid_tool_calls`; the deleted branch is what echoed those raw fragments back. Nothing in the plugin produces such a message today, and the live run never hit one, so the case ships uncovered rather than growing a second serialization path on speculation.
- No LangChain 2.x or `openai` 7.x jump, no provider or embedding-model changes, no prompt content changes.

## Screenshot

Not uploaded — the images exist but could not be attached from this session: GitHub image hosting requires the authenticated web editor, and this session could not reach it.

What was captured, and what it shows: Quick Chat in `copilot plus` mode, `google/gemini-2.5-flash` over the real OpenRouter API, the same prompt run on a build of the base branch (`1fbbfd54-clean`) and on a build of this PR (`c7373255-clean`). Both renders are the same: the model calls `localSearch`, the answer carries numbered citations, and the Sources block appears — no visual difference, which is the point. A third capture on the PR build, same model with the reasoning capability enabled and a plain non-tool question, shows the collapsible "Thought for a while" block above the streamed answer.

The load-bearing evidence is textual. A `window.fetch` interceptor in the Obsidian renderer recorded every outgoing OpenRouter request body on both builds. Normalizing only ids and timestamps, the four request bodies differ in exactly three lines, all traceable to one cause — the model phrased its own search query as `"wikilinks versus RAG"` on the base run and `"wikilinks vs RAG"` on the PR run, which then changes the downstream salient-term prompt and the search results embedded in the tool result. Message roles, message keys, tool definitions, and the `tool_calls` structure are identical:

```
#3  base   roles: system,user,assistant,tool
    assistant.tool_calls = [{"id":"tool_localSearch_JNYBwNUrbTPSBqlw871t","type":"function",
                             "function":{"name":"localSearch","arguments":"{…}"}}]
    tool.tool_call_id    = tool_localSearch_JNYBwNUrbTPSBqlw871t

#3  PR     roles: system,user,assistant,tool
    assistant.tool_calls = [{"id":"tool_localSearch_jJggpMMb3qrvjdJLPpJv","type":"function",
                             "function":{"name":"localSearch","arguments":"{…}"}}]
    tool.tool_call_id    = tool_localSearch_jJggpMMb3qrvjdJLPpJv
```

## Risk

**Medium**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | `ChatOpenRouter.test.ts` covers both serialization paths: a first-class `AIMessage.tool_calls`, and an `AIMessageChunk` aggregated from streamed `tool_call_chunks` the way `stream()` aggregates it |
| A defect would fail CI or be obvious on first use | ✅ | OpenRouter rejects an assistant/`tool` pair whose assistant message lost its `tool_calls` with a 400, on the first tool-using Quick Chat turn |
| A revert fully restores prior state, including persisted data | ✅ | Delete-only; no migration, no persisted write |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | Message serialization only |
| No public API, plugin API, message, or on-disk contract changes | ✅ | The outgoing chat-completions request shape is unchanged; request bodies recorded from base and PR builds differ only in the model's own wording |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Synchronous mapping inside `toOpenRouterMessages` and `buildMessageChunk` |
| No hot-path behavior lacks deterministic coverage | ✅ | Both remaining serialization branches are asserted |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Confined to Quick Chat's tool-calling turn on the three `ChatOpenRouter`-backed providers |

Review: read `toOpenRouterMessages` and `buildMessageChunk` in `src/LLMProviders/ChatOpenRouter.ts` and confirm nothing else in the diff reaches `additional_kwargs`, then run Verification steps 2–4 and confirm the second turn returns an answer rather than an OpenRouter 400.

## Verification

1. Configure an OpenRouter provider with a tool-capable model (`google/gemini-2.5-flash` works) and enable it for chat.
2. Open Quick Chat, switch the mode selector to `copilot plus`, and pick that model.
3. Send a prompt that forces a vault lookup, e.g. `Search my vault and tell me in one sentence what I wrote about <a topic in your vault>`. Expect a normal answer with numbered citations and a Sources block — the model called `localSearch`, the result was fed back, and OpenRouter accepted the follow-up request. An assistant message missing its `tool_calls` would surface here as a 400 instead.
4. Send a follow-up in the same chat that needs no tool, and confirm it streams normally.
5. Switch the mode selector back to `chat (free)` and ask a plain question with a model that has the reasoning capability enabled. Confirm the collapsible thinking block still appears above the streamed answer.

## Note

The accepted behavior change is narrow: a message that carries tool calls *only* under `additional_kwargs.tool_calls` now serializes as a plain assistant message. After this PR nothing in the plugin creates such a message — `buildMessageChunk` no longer writes the container, and both chain runners build assistant turns from first-class `tool_calls`.


zeroliu:
@codex review

PR #2871, opened 2026-08-16
Decode entitlement tokens through the buffer polyfill instead of atob

zeroliu:
Fixes logancyang/obsidian-copilot-preview#304

## Why

The Obsidian community plugin review scorecard for the published release reports
"`atob` is deprecated. Use `Buffer.from(data, 'base64')` instead — 1 occurrence"
against `src/entitlement/verify.ts:28`. That call decodes the three base64url
segments of the offline ES256 entitlement token — header, claims, signature —
which the plugin reads on every check of whether a user is on Plus or Lite. The
finding has to clear for the plugin's review to pass.

The deprecation is a typings artifact rather than a platform problem. `atob` is
standard WHATWG API in both the Electron renderer and the mobile WebView;
`@types/node` marks only its own Node global deprecated, and TypeScript resolves
to that declaration.

A previous attempt, #2819, read the scorecard's suggested `Buffer.from` as a
desktop-only Node built-in unavailable on mobile, and hand-rolled a WHATWG
forgiving-base64 decoder to work around it. That premise does not hold in this
repo: `buffer` is a runtime dependency bundled by esbuild, and `src/utils.ts` and
`src/LLMProviders/BedrockChatModel.ts` already decode base64 to bytes through it
on mobile today.

## What

Entitlement tokens decode through the bundled `buffer` polyfill. Every
well-formed segment produces the same bytes as before on desktop and mobile, so
license recognition is unchanged. Malformed segments are still rejected, but by
a different mechanism.

| Segment | Before | After |
| --- | --- | --- |
| Well-formed | decodes to bytes | decodes to the same bytes |
| Malformed | decoder throws, caller catches, token rejected | decoder returns partial bytes, token rejected by JSON parsing or the ES256 check |

No production code calls `atob`, and the decoder's base64url alphabet handling
and padding step are gone, since the polyfill accepts both natively.

## Non goal

- The `atob` occurrences inside `src/agentMode/skills/builtin/builtinSkills.ts`
  shell-script string templates. Different execution context, not flagged, and
  skill content must not change.
- `atob` calls inside bundled third-party dependencies.
- Other scorecard warnings, tracked separately.
- Any change to token semantics: signature verification, expiry, or user
  binding.
- Preserving the exact throw behavior of `atob` on malformed segments. Rejection
  is preserved; the mechanism that produces it is not.

## Screenshot

Not applicable. This changes headless license-verification internals with no
visual surface.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Three tests in `verify.test.ts` cover base64url-alphabet segments, unpadded segments at every remainder length, and out-of-alphabet rejection |
| A defect would fail CI or be obvious on first use | ✅ | Those tests run in the unit suite in CI, and a decode defect rejects every license on the next plan check |
| A revert fully restores prior state, including persisted data | ✅ | No persisted data or format change; the token wire format is untouched |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | This changes how untrusted license-token input is decoded inside the entitlement verifier |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Module-internal helper; `verifyEntitlement`'s signature and the token format are unchanged |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | The decoder is pure and synchronous |
| No hot-path behavior lacks deterministic coverage | ✅ | Valid decode, malformed rejection, alphabet, and padding cases are all deterministic tests |
| No new dependency | ✅ | `buffer` is already a runtime dependency; dependency manifests are unchanged |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | A wrong decode surfaces immediately as a license no longer being recognized |

Review: inspect `src/entitlement/verify.ts` — `base64UrlToBytes` — against the
malformed-segment row of the What table, confirming both callers still reject
what it returns; then `src/entitlement/verify.test.ts` — the three tests named
in the first Risk row — to confirm they pin decode output rather than the
decoder's identity. Then run Verification steps 2 and 3, including the
tampered-token variant.

## Verification

1. Build the branch and load it in a test vault signed in to a licensed Copilot
   account. Reload the plugin and confirm the plan still shows as Plus or Lite
   in settings.
2. Change one character of the cached license token in the plugin's `data.json`,
   reload, and confirm the license is no longer recognized and the plan falls
   back.
3. Restore the original token, reload, and confirm the plan is recognized again.
4. Repeat step 1 on a real mobile device, where the `buffer` polyfill rather than
   any Node built-in supplies the decoder, and confirm the plan is recognized
   there too.

## Note

The issue's success criteria ask for decode output that is byte-identical to the
`atob` path *and* identical throw behavior. This PR delivers the first and
deliberately not the second, as recorded under Non goal. Malformed segments are
still rejected, by JSON parsing or the ES256 check rather than by a throw, so the
security property holds; the throw-parity criterion should be reconciled on the
issue rather than read as unmet work.

### DIFF
--- a/src/entitlement/verify.test.ts
+++ b/src/entitlement/verify.test.ts
@@ -133,4 +133,40 @@ describe("verifyEntitlement", () => {
     expect(await verify("not-a-jwt", { publicKeys })).toBeNull();
     expect(await verify("only.two", { publicKeys })).toBeNull();
   });
+
+  it("rejects segments containing characters outside the base64url alphabet", async () => {
+    const token = await signToken(keyPair.privateKey, plusClaims());
+    const [header, payload, signature] = token.split(".");
+    // The decoder skips these characters instead of throwing, so each segment
+    // must still be rejected downstream by JSON parsing or ES256 verification.
+    expect(await verify(`!!!!.${payload}.${signature}`, { publicKeys })).toBeNull();
+    expect(await verify(`${header}.!!!!.${signature}`, { publicKeys })).toBeNull();
+    expect(await verify(`${header}.${payload}.!!!!`, { publicKeys })).toBeNull();
+  });
+
+  it("verifies a token whose segments use the base64url-only '-' and '_' characters", async () => {
+    // U+07FF encodes to base64 indices 62 and 63, which base64url spells "-" and
+    // "_". A decoder handling only the standard alphabet drops both, corrupting
+    // the bytes and silently rejecting a valid license.
+    const userId = `${USER_ID}\u07ff\u07ff`;
+    const token = await signToken(keyPair.privateKey, plusClaims({ user_id: userId }));
+    const payloadSegment = token.split(".")[1];
+    expect(payloadSegment).toContain("-");
+    expect(payloadSegment).toContain("_");
+    expect((await verify(token, { publicKeys }))?.user_id).toBe(userId);
+  });
+
+  it("verifies tokens whose unpadded segments cover every base64 remainder length", async () => {
+    // base64url segments carry no padding, so each byte length mod 3 produces a
+    // different implied-padding case; a decoder mishandling one would reject
+    // otherwise-valid licenses.
+    const remainders = new Set<number>();
+    for (const filler of ["", "x", "xx"]) {
+      const token = await signToken(keyPair.privateKey, plusClaims({ user_id: USER_ID + filler }));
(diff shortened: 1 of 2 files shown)


zeroliu:
@codex review

Issue #2836, opened 2026-08-13
Semantic search is unreachable on a fresh install: the embedding model cannot be selected from the UI


LesYeuxNoirz:
Copilot version: 4.0
Model used: Doesn't matter

Build index fails permanently for any user who does not happen to hold an OpenRouter API key, and there is no supported way to fix it. The default embedding model is openai/text-embedding-3-small|openrouterai, and v4 appears to ship no UI for changing it.

Steps to reproduce:

1. Enable the debug mode for better error logging
2. Fresh v4 install, no OpenRouter key configured
3. Open Relevant Notes → Build index
4. Check the console

Fatal error during indexing: API key is not provided for the embedding model:

```js
openai/text-embedding-3-small|openrouterai
    at t.getEmbeddingsAPI
    at QQ.indexVaultToVectorStore
```

danbarua:
Worth noting that basic Chat is broken, not just Vault QA. It seems that text embeddings are needed for both.

JianNicky:
I ran into the same problem: I couldn't find the settings page for the Embedding model.  
1. Version: V4.0.2  
2. In BYOK, I set up the company's internal Embedding  
3. When running "Copilot: Force reindex vault," it shows “1 error: API key is not provided for the embedding model: openai/text-embedding-3-smalllopenrouterai”

GeorgeTsau:
In the Miyo settings, I tried to switch to a local embedding model, the URL is supposed to be like http://localhost:11434/v1, but the configuration required a "HuggingFace URL", which is not a local address. How can I solve that? 

<img width="831" height="757" alt="Image" src="https://github.com/user-attachments/assets/b7fd31de-3042-4428-a05a-a3f4ee8ee3a8" />

GeorgeTsau:
by the way, the previous versions are much convenient to configure chat and embedding models than the current version using MIYO and BYOK, the upgrade is an inconvenience

zeroliu:
cc @wenzhengjiang for the local embedding model config issue in Miyo

zeroliu:
Regarding the local embedding issue, we have improved our Miyo UI and made setting up local embedding model more intuitive. Please feel free to let us know if you still have questions.

Issue #333, opened 2024-03-05
plugin running on system windows reported error


szx-zhao:
I'm running the Ollama service on Windows, and I'm wondering how to enable the Obsidian plugin to retrieve data correctly. I'm getting an error in the console that says `Access to fetch at 'http://127.0.0.1:11434/api/chat' from origin 'app://obsidian.md' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.`, Meanwhile, requests received at ollama Server `[GIN] 2024/03/05 - 23:46:03 | 403 |            0s |       127.0.0.1 | OPTIONS  "/api/chat"` are .I'm not sure how to resolve it.

logancyang:
@szx-zhao have you checked the setup guide https://github.com/logancyang/obsidian-copilot/blob/master/local_copilot.md

szx-zhao:
> @szx-zhao have you checked the setup guide https://github.com/logancyang/obsidian-copilot/blob/master/local_copilot.md

Thanks, I ran the `setx OLLAMA_ORIGINS "app://obsidian.md*"` command in windows shell and the issue seems to be resolved now

PR #2908, opened 2026-08-19
Authenticate keyless openai-format embedding endpoints

### LINKED ISSUE #2837: 3rd party (openai-format) embeddings cannot authenticate to keyless local endpoints Summary
Copilot version: v4.0
Model used: any Ollama-managed embedding model

## Summary

The embedding config for 3rd party (openai-format) falls back to an empty string where every comparable provider falls back to "default-key". The OpenAI SDK refuses to construct with an empty credential, so keyless OpenAI-compatible endpoints — Ollama, LM Studio behind a proxy, LiteLLM, vLLM — cannot be used for embeddings at all.

## Findings

In `getEmbeddingConfig`:

- "3rd party (openai-format)": { modelName: r, openAIApiKey: e.apiKey || "",           … }
- "lm-studio":                 { modelName: r, openAIApiKey: e.apiKey || "default-key", … }
- ollama:                      { model: r, headers: { Authorization: `Bearer ${e.apiKey || "default-key"}` } }

The gate and the constructor disagree. `buildModelMap()` computes `hasApiKey: !!(r.apiKey || this.providerApiKeyMap[r.provider]())`, and `providerApiKeyMap["3rd party (openai-format)"]` returns the literal `"default-key"` — so the check passes. `getEmbeddingConfig()` then hands the constructor `e.apiKey`, which is `""`. The model is reported as usable and then fails to instantiate.

```js
Error getting vector length: Missing credentials. Please pass an `apiKey`, `workloadIdentity`,
`adminAPIKey`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
    at new $n
    at k0.embeddingWithRetry
    at k0.embedQuery
    at Zqe            ← getVectorLength
    at t.createNewDb
    at t.initializeDB
```

Surfaces to the user only as `Failed to initialize Copilot database. Some features may be limited`. The real cause is logged through `B()`, which is gated on debug, so it is invisible by default.

zeroliu:
Fixes #2837

## Why

A vault upgrading from 3.x can carry a saved embedding model whose provider is
`3rd party (openai-format)` and whose API key field is empty, because the
endpoint behind it — Ollama or LM Studio behind a proxy, LiteLLM, vLLM — needs
no key. Copilot lists that model as usable, then fails the moment it tries to
embed anything. The user sees only `Failed to initialize Copilot database. Some
features may be limited`; the real cause, `Missing credentials. Please pass an
apiKey ...` from the OpenAI SDK, is logged behind the debug flag and is
invisible by default.

The two halves of the code disagreed about what a keyless endpoint looks like.
The usability check treats a missing key for this provider as the placeholder
`default-key`, so the model passes the gate. The config that actually builds the
embedding client handed it an empty string instead, and the OpenAI SDK refuses
to construct a client without a credential.

## What

> **Before:** a saved `3rd party (openai-format)` embedding model with an empty
> API key is offered as usable, and every embedding attempt against it fails
> with a generic database-initialization notice.
>
> **After:** the same model reaches its endpoint. The request carries the same
> non-secret `default-key` placeholder that keyless LM Studio and Ollama
> embedding models already send, so a keyless endpoint answers normally and an
> endpoint that really does want a key answers with its own auth error instead
> of a silent local failure.

Nothing changes for a model that has an API key set: that key is still used.

## Non goal

- Does not generalize the fallback across the other provider blocks in the same
  config. Each block names its credential field differently — `apiKey`,
  `openAIApiKey`, `azureOpenAIApiKey`, an `Authorization` header — so deriving
  every one of the eleven from the shared provider-key map costs roughly ten
  lines and eleven touch points in a file that is scheduled for deletion.
- Does not add embedding-model UI. v4 has none, so this only helps vaults whose
  saved settings already contain such a model.
- Does not change how the underlying failure is surfaced or logged.

## Screenshot

Not applicable — this changes the credential handed to an embedding client
constructor. It has no rendered surface, and v4 ships no embedding-model UI to
photograph.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | `src/LLMProviders/embeddingManager.test.ts` asserts the credential handed to the embedding client for keyless openai-format, LM Studio, and Ollama models |
| A defect would fail CI or be obvious on first use | ✅ | Those tests run in CI and fail against the previous empty-string fallback |
| A revert fully restores prior state, including persisted data | ✅ | Nothing is written or migrated; the change is one literal in a config built per call |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | The credential sent to an openai-format embedding endpoint changes from an empty string to the `default-key` placeholder when the user set no key |
| No public API, plugin API, message, or on-disk contract changes | ✅ | No exported signature, settings schema, or serialized format changes |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | A synchronous value in a config object |
| No hot-path behavior lacks deterministic coverage | ✅ | The one changed branch is covered by the new tests |
| No new dependency | ✅ | `package.json` and `package-lock.json` are unchanged |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | A wrong value fails on the first embedding call against the configured endpoint |

Review: inspect the `EmbeddingModelProviders.OPENAI_FORMAT` entry of
`getEmbeddingConfig` in `src/LLMProviders/embeddingManager.ts` and confirm its
`openAIApiKey` fallback matches the `OPENAI_FORMAT` entry of
`providerApiKeyMap` in the same file. Then run Verification steps 2 and 3,
including the step-4 variant where the endpoint requires a key.

## Verification

1. In a vault whose settings already list an embedding model with provider
   `3rd party (openai-format)`, an empty API key, and a base URL pointing at a
   keyless OpenAI-compatible server (for example `llama-server`, vLLM, or
   LiteLLM with auth disabled), select that model as the embedding model.
2. Reload the plugin and trigger indexing. On `master`, `Failed to initialize
   Copilot database. Some features may be limited` appears and, with debug
   logging on, the console shows `Missing credentials`.
3. On this branch, repeat step 2. Indexing proceeds and the notice does not
   appear.
4. Point the same model at an endpoint that does require a key and leave the key
   blank. Confirm the failure now comes back from the server as an
   authentication error rather than as the generic database notice.

## Note

`src/LLMProviders/embeddingManager.ts` is scheduled for deletion by
`logancyang/obsidian-copilot-preview#333` under epic
`logancyang/obsidian-copilot-preview#336`, which retires the in-plugin Orama
indexing path. With no embedding-model UI in v4
(`logancyang/obsidian-copilot#2836`), this fix reaches only vaults upgrading
from 3.x that carry a persisted `activeEmbeddingModels` row. The reviewer may
want to judge whether the one-line fix is worth carrying until `#333` lands.

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -127,3 +127,16 @@ if (typeof window.activeDocument === "undefined") {
 if (typeof window.activeWindow === "undefined") {
   window.activeWindow = window;
 }
+
+// Polyfill Obsidian's `Array.prototype.contains` augmentation (an alias of
+// `includes`). It is defined non-enumerably so `for...in` over arrays is
+// unaffected, matching how the runtime installs it.
+if (typeof Array.prototype.contains !== "function") {
+  Object.defineProperty(Array.prototype, "contains", {
+    value: function (target) {
+      return this.includes(target);
+    },
+    writable: true,
+    configurable: true,
+  });
+}
--- a/src/LLMProviders/embeddingManager.test.ts
+++ b/src/LLMProviders/embeddingManager.test.ts
@@ -1,0 +1,62 @@
+import type { CustomModel } from "@/aiParams";
+import { EmbeddingModelProviders } from "@/constants";
+import { getModelKeyFromModel, setSettings } from "@/settings/model";
+import { OllamaEmbeddings } from "@langchain/ollama";
+import { OpenAIEmbeddings } from "@langchain/openai";
+import { CustomOpenAIEmbeddings } from "./CustomOpenAIEmbeddings";
+import EmbeddingManager from "./embeddingManager";
+
+jest.mock("@langchain/openai", () => ({
+  ...jest.requireActual<typeof import("@langchain/openai")>("@langchain/openai"),
+  OpenAIEmbeddings: jest.fn(),
+}));
+jest.mock("@langchain/ollama", () => ({
+  ...jest.requireActual<typeof import("@langchain/ollama")>("@langchain/ollama"),
+  OllamaEmbeddings: jest.fn(),
+}));
+jest.mock("./CustomOpenAIEmbeddings", () => ({
+  CustomOpenAIEmbeddings: jest.fn(),
+}));
+
+describe("embeddingManager", () => {
+  describe("EmbeddingManager", () => {
+    describe("getEmbeddingsAPI()", () => {
+      beforeEach(() => {
+        jest.clearAllMocks();
+      });
+
+      it.each([
+        [EmbeddingModelProviders.OPENAI_FORMAT, OpenAIEmbeddings, { openAIApiKey: "default-key" }],
+        [
+          EmbeddingModelProviders.LM_STUDIO,
+          CustomOpenAIEmbeddings,
+          { openAIApiKey: "default-key" },
+        ],
+        [
+          EmbeddingModelProviders.OLLAMA,
+          OllamaEmbeddings,
+          { headers: { Authorization: "Bearer default-key" } },
+        ],
+      ])(
+        "credentials a keyless %s model with the keyless sentinel instead of an empty key",
+        async (provider, EmbeddingConstructor, expectedCredential) => {
+          const model: CustomModel = {
+            name: "local-embedding-model",
+            provider,
+            enabled: true,
+          };
+          setSettings({
+            activeEmbeddingModels: [model],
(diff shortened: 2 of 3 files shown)


zeroliu:
We are getting rid of embedding endpoints. Close

PR #2914, opened 2026-08-20
Cover the full CSI parameter-byte range when stripping agent escape sequences

### LINKED ISSUE #2876: sanitize ANSI / OSC terminal escape sequences before parsing ACP JSON-RPC messages
#### Description
When launching Agent Mode with an external CLI agent backend (e.g. `opencode`, custom agent wrappers, or tools with terminal plugins/hooks), the Copilot UI permanently hangs on:
```text
Loading agent models…
```
In the Obsidian developer console (`⌥ ⌘ I`), the following unhandled exception is thrown:
```text
Failed to parse JSON message: ]0;opencode: ready]0;second-brain: ready{"jsonrpc":"2.0","id":1,"result":{...}}
SyntaxError: Unexpected token '', "]0;openco"... is not valid JSON
    at JSON.parse (<anonymous>)
    at Object.start (plugin:copilot:...)
```

---

### Root Cause

1. **Subprocess stdout Stream Decoding**:
   Copilot spawns agent processes via Node's `child_process.spawn` with piped `stdout`.
   In the ACP stream reader (`a5t` / `createAcpStream`), incoming chunks are split by newline (`\n`), trimmed, and directly passed to `JSON.parse(line)`.

2. **Terminal Title & ANSI Injection**:
   Many CLI tools, plugins, or terminal integrations emit OSC (Operating System Command) window title updates (e.g., `\x1b]0;project: status\x07`) or color/style ANSI escape codes to `stdout`.
   Because these sequences can be prepended to the JSON-RPC response before a newline:
   ```text
   \x1b]0;opencode: ready\x07{"jsonrpc":"2.0","id":1,"result":{...}}
   ```
   `JSON.parse(line)` fails with `SyntaxError`, dropping the ACP initialization/session payload. As a result, the preload probe promise never settles, leaving `isPreloadReady` in a pending state forever.

---

### Proposed Fix

Sanitize the line string before calling `JSON.parse`:
1. Strip ANSI / OSC escape sequences (`\x1b\[[0-9;]*[a-zA-Z]|\x1b\][^\x07\x1b]*(?:\x07|\x1b\\)`).
2. If non-JSON text precedes the JSON envelope, slice from the first `{` to the matching last `}`.

#### Code Diff

```diff
function createAcpStream(stdinStream, stdoutStream) {
  const encoder = new TextEncoder();
  const decoder = new TextDecoder();

+ function cleanJsonPayload(raw) {
+   // Strip ANSI escape codes and OSC terminal title sequences
+   let cleaned = raw.replace(/\x1b\[[0-9;]*[a-zA-Z]|\x1b\][^\x07\x1b]*(?:\x07|\x1b\\)/g, "").trim();
+   // Extract JSON object envelope if leading characters exist
+   const jsonStart = cleaned.indexOf("{");
+   if (jsonStart !== -1) {
+     const jsonEnd = cleaned.lastIndexOf("}");
+     if (jsonEnd !== -1 && jsonEnd >= jsonStart) {
+       cleaned = cleaned.slice(jsonStart, jsonEnd + 1);
+     }
+   }
+   return cleaned || raw;
+ }

  const readable = new ReadableStream({
    async start(controller) {
      let buffer = "";
      const reader = stdoutStream.getReader();
      try {
        while (true) {
          const { value, done } = await reader.read();
          if (done) {
            buffer += decoder.decode();
            break;
          }
          if (!value) continue;
          buffer += decoder.decode(value, { stream: true });
          const lines = buffer.split("\n");
          buffer = lines.pop() || "";
          for (const line of lines) {
            const trimmed = line.trim();
            if (trimmed) {
              try {
-               const parsed = JSON.parse(trimmed);
+               const parsed = JSON.parse(cleanJsonPayload(trimmed));
                controller.enqueue(parsed);
              } catch (err) {
                console.error("Failed to parse JSON message:", trimmed, err);
              }
            }
          }
        }
        const remaining = buffer.trim();
        if (remaining) {
          try {
-           const parsed = JSON.parse(remaining);
+           const parsed = JSON.parse(cleanJsonPayload(remaining));
            controller.enqueue(parsed);
          } catch (err) {
            console.error("Failed to parse JSON message:", remaining, err);
          }
        }
      } catch (err) {
        controller.error(err);
      } finally {
        reader.releaseLock();
      }
      controller.close();
    }
  });

  // ... writable stream remains unchanged
}
```

---

### Steps to Reproduce

1. Install an OpenCode or Claude hook/plugin that writes terminal titles to stdout (e.g. `process.stdout.write('\x1b]0;ready\x07')`).
2. Configure Copilot to use that agent binary in **Settings → Copilot → Agent Mode**.
3. Open the Copilot Agent tab in Obsidian.
4. **Observed**: View hangs indefinitely on `"Loading agent models…"`. Console displays `SyntaxError: Unexpected token '', "]0;..." is not valid JSON`.
5. **Expected**: Non-JSON ANSI/OSC prefix is ignored, JSON-RPC payload is parsed, and Agent Mode models load normally.

zeroliu:
Fixes #2876

## Why

Agent Mode strips terminal escape sequences from a backend's stdout before the ACP SDK parses each NDJSON frame (#2909, for #2876). Its CSI matcher accepts only digits, `;` and `?` as parameter bytes.

ECMA-48 defines that parameter range as `0x30`-`0x3f`, which also covers `:`, `<`, `=` and `>`. So a backend that prefixes a frame with a truecolor sequence such as `\x1b[38:2:255:0:0m`, or with a private `<` / `=` / `>` form, is left untouched today: the escape bytes reach `JSON.parse`, it throws, and `ndJsonStream` drops the frame with nothing surfaced to the user. When the dropped frame is the initialization response, the agent never finishes connecting and the chat sits on its connecting state forever, which is the hang #2876 reported.

`AcpProcessManager.start()` also had no test of its own. The suite calls `sanitizeAcpStdout()` directly and `AcpBackendProcess.test.ts` mocks `AcpProcessManager` wholesale, so unwiring the sanitizer from what `start()` returns left every test green while production went back to hanging.

## What

> **Before:** a backend writes `\x1b[38:2:255:0:0m{"jsonrpc":"2.0",...}` on stdout. The prefix survives sanitizing, the frame is dropped, and Agent Mode hangs on connect.
>
> **After:** the prefix is stripped, the frame parses, and the session connects.

| Frame prefix | Before | After |
| --- | --- | --- |
| `\x1b[32m` (digit parameters) | stripped | stripped |
| `\x1b]0;title\x07` (OSC) | stripped | stripped |
| `\x1b[38:2:255:0:0m` (truecolor, `:` separators) | reaches `JSON.parse`, frame dropped | stripped |
| `\x1b[<0;1;2M`, `\x1b[=5h`, `\x1b[>4;2m` (private forms) | reaches `JSON.parse`, frame dropped | stripped |

No change for anyone whose backend emits only the already-covered sequences.

## Non goal

- No change to how frames are buffered, split across chunk boundaries, or logged.
- No handling of escape sequences that are neither CSI nor OSC.
- No change to stderr routing or to the debug frame log.

## Screenshot

Not applicable. The change is inside the stdout byte stream feeding the ACP parser; nothing about it renders. The behavior evidence is the table above, each row covered by a test.

## Risk

**Low**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | `AcpProcessManager.test.ts` covers the truecolor and private CSI forms, the OSC case, split chunks, escape-free frames, and the missing-trailing-newline flush |
| A defect would fail CI or be obvious on first use | ✅ | Those tests run in CI, and a mis-stripped frame breaks agent connect on first use |
| A revert fully restores prior state, including persisted data | ✅ | Pure in-memory stream transform, nothing persisted |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | Only which escape bytes are removed before parsing |
| No public API, plugin API, message, or on-disk contract changes | ✅ | `sanitizeAcpStdout` and `start()` keep their signatures |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | The reader loop and buffering are untouched |
| No hot-path behavior lacks deterministic coverage | ✅ | Every branch of the widened character class is asserted |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Confined to Agent Mode connect, visible immediately |

Review: read `TERMINAL_ESCAPE_SEQUENCE` in `src/agentMode/acp/AcpProcessManager.ts` and confirm the widened class only ever applies after a literal `\x1b[`, so JSON payload bytes cannot match. Then run Verification step 1.

## Verification

1. Run `npm test -- src/agentMode/acp`. The `sanitizeAcpStdout()` truecolor case fails against the previous character class and passes with this one.
2. Start Agent Mode with an opencode or codex backend and confirm the session still connects and streams a reply as before.

## Note

The `it(...)` descriptions and the matcher comment now carry the full issue URL, which AGENTS.md requires for behavioral branches added for a specific report.

### DIFF
--- a/src/agentMode/acp/AcpProcessManager.test.ts
+++ b/src/agentMode/acp/AcpProcessManager.test.ts
@@ -1,4 +1,15 @@
-import { sanitizeAcpStdout } from "@/agentMode/acp/AcpProcessManager";
+import { requireNodeModule } from "@/utils/desktopRuntime";
+import { AcpProcessManager, sanitizeAcpStdout } from "@/agentMode/acp/AcpProcessManager";
+
+jest.mock("@/logger", () => ({
+  logInfo: jest.fn(),
+  logWarn: jest.fn(),
+  logError: jest.fn(),
+}));
+
+jest.mock("@/utils/desktopRuntime", () => ({
+  requireNodeModule: jest.fn(),
+}));
 
 /** The exact prefix reported in the opencode hang (two OSC title writes). */
 const OSC_TITLES = "\x1b]0;opencode: ready\x07\x1b]0;second-brain: ready\x07";
@@ -28,8 +39,33 @@ async function readLines(stream: ReadableStream<Uint8Array>): Promise<string[]> 
 }
 
 describe("AcpProcessManager", () => {
+  describe("AcpProcessManager", () => {
+    describe("start()", () => {
+      it("returns a stdout stream with the child's terminal escape sequences already stripped (https://github.com/logancyang/obsidian-copilot/issues/2876)", async () => {
+        const identity = <T>(value: T): T => value;
+        const child = {
+          stdin: {},
+          // `toWeb` is the identity below, so the manager receives this as-is.
+          stdout: streamOf([`${OSC_TITLES}${ENVELOPE}\n`]),
+          stderr: { setEncoding: jest.fn(), on: jest.fn() },
+          on: jest.fn(),
+        };
+        (requireNodeModule as jest.Mock).mockImplementation((id: string) =>
+          id === "child_process"
+            ? { spawn: () => child }
+            : { Readable: { toWeb: identity }, Writable: { toWeb: identity } }
+        );
+        const manager = new AcpProcessManager({ command: "/bin/agent", args: ["acp"], env: {} });
+
+        const { stdout } = manager.start();
+
+        expect(await readLines(stdout)).toEqual([ENVELOPE]);
+      });
+    });
+  });
+
   describe("sanitizeAcpStdout()", () => {
-    it("makes a frame prefixed with OSC title sequences parse as JSON", async () => {
+    it("makes a frame prefixed with OSC title sequences parse as JSON (https://github.com/logancyang/obsidian-copilot/issues/2876)", async () => {
       const lines = await readLines(sanitizeAcpStdout(streamOf([`${OSC_TITLES}${ENVELOPE}\n`])));
 
       expect(lines).toHaveLength(1);
@@ -40,7 +76,16 @@ describe("AcpProcessManager", () => {
       });
     });
 
-    it("makes a frame parse as JSON when a chunk boundary splits an escape sequence", async () => {
+    it("makes a frame prefixed with CSI sequences that use the full parameter-byte range parse as JSON (https://github.com/logancyang/obsidian-copilot/issues/2876)", async () => {
+      // Truecolor uses `:` separators; `<`, `=` and `>` appear in private forms.
+      const prefix = "\x1b[38:2:255:0:0m\x1b[<0;1;2M\x1b[=5h\x1b[>4;2m";
+
+      const lines = await readLines(sanitizeAcpStdout(streamOf([`${prefix}${ENVELOPE}\n`])));
+
+      expect(lines).toEqual([ENVELOPE]);
+    });
+
+    it("makes a frame parse as JSON when a chunk boundary splits an escape sequence (https://github.com/logancyang/obsidian-copilot/issues/2876)", async () => {
       const payload = `\x1b[32m${OSC_TITLES}${ENVELOPE}\n`;
       // Cuts land inside the CSI colour code and inside the first OSC title.
       const chunks = [payload.slice(0, 3), payload.slice(3, 12), payload.slice(12)];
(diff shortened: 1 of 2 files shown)


zeroliu:
@codex review

PR #2981, opened 2026-08-26
Guide users through multi-question agent prompts

zeroliu:
Fixes Brevilabs/obsidian-copilot-private#117

## Why

Multi-question agent prompts show a disabled **Submit** button from the first question onward. After answering the visible question, users must infer that another tab needs attention, so the card can appear stuck even though the flow still works.

## What

The tabs remain available for review and direct navigation, while the footer now communicates the action for the visible step.

| State | Primary action | Availability |
| --- | --- | --- |
| Before the final question | **Next** | Enabled when the visible question is answered |
| Final question with an earlier answer missing | **Submit** | Disabled |
| Final question with every answer complete | **Submit** | Enabled |

Cmd/Ctrl+Enter from a custom response follows the same visible Next or Submit action.

## Non goal

- Changing question types, answer validation, or the response payload.
- Redesigning the single-question card flow.
- Changing cancellation behavior.

## Screenshot

| Before | After |
| --- | --- |
| Disabled **Submit** on the first question | Disabled **Next** on the first question |
| Before: the first question shows a disabled Submit button | After: the first question shows a disabled Next button |

## Risk

**Low**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Component tests cover Next gating, final Submit gating, complete answer submission, tab skipping, and the keyboard shortcut |
| A defect would fail CI or be obvious on first use | ✅ | The focused component suite exercises every changed branch |
| A revert fully restores prior state, including persisted data | ✅ | The card keeps only transient React state and writes no persisted data |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | Existing option and custom-text answer handling is unchanged |
| No public API, plugin API, message, or on-disk contract changes | ✅ | The request and resolved-answer contracts are unchanged |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Navigation remains synchronous component-local state |
| No hot-path behavior lacks deterministic coverage | ✅ | Every new primary-action outcome is covered in the component suite |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Any navigation error appears immediately in the inline question card |

Review: run Verification steps 2-4 once and confirm the footer matches the visible question state.

## Verification

1. Run `npm run gallery:vault`, then open **Agent Mode / Ask User Question Card / MultipleQuestions** in the component gallery at 400 px.
2. Confirm **Next** is disabled on the first question, select an answer, then confirm **Next** enables and advances to the second tab without submitting.
3. Select the final tab while the middle question is unanswered, answer the final question, and confirm **Submit** remains disabled.
4. Answer the skipped middle question, return to the final tab, and confirm **Submit** is enabled.

### DIFF
--- a/docs/agent-mode-and-tools.md
+++ b/docs/agent-mode-and-tools.md
@@ -80,7 +80,7 @@ opencode supports **Default** and **Auto**. Claude supports **Default**, **Plan*
 
 When an action needs approval, Agent Chat displays a **Permission required** card with the proposed change or tool input. Choose one of the temporary or persistent allow or deny options offered by that agent. Stopping the turn cancels unanswered requests.
 
-When an agent asks a set of questions, answer every question tab before **Submit** becomes available; **Cancel** declines the entire request.
+When an agent asks a set of questions, answer the current tab and select **Next**. On the final tab, **Submit** becomes available after every question has an answer. You can use the tabs to review or skip ahead; **Cancel** declines the entire request.
 
 Your vault or project is the agent's working directory, not a security sandbox. Auto or bypass permissions can reach other files and services available to the agent or your account. Use **Default** for unfamiliar work and review persistent permissions carefully.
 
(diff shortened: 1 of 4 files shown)


zeroliu:
@codex review

PR #2994, opened 2026-08-27
[Relevant Notes on Miyo] - Step 6: Sync Index scope changes to Miyo

zeroliu:
Fixes Brevilabs/obsidian-copilot-private#310

## Why

Index scope changes currently affect Copilot's result filtering only. Miyo keeps the folder and extension filters captured when the vault was registered, so its index—and anything using that registered folder through Miyo—can drift from the settings the user just saved in Copilot.

A safe update cannot replace the registration wholesale: Miyo owns additional folder settings, including filters Copilot cannot represent and remote read/write choices. Rapid settings changes also must arrive in order, and a queued update must not be sent to a different Miyo endpoint or folder after the user changes configuration.

## What

When an existing or previously recorded Miyo registration is known, changing Copilot's inclusion or exclusion setting patches that folder in place.

| Copilot pattern or state | Miyo update |
| --- | --- |
| Folder pattern | Projected to the registered folder filter |
| Extension pattern | Projected to the registered folder filter |
| Tag, individual-note, or property pattern | Remains Copilot-side; not sent to Miyo |
| Direct Miyo-only filter | Preserved |
| Direct Copilot include whitelist | Takes precedence over broader Miyo include filters Copilot previously projected |
| Unregistered folder | No registration is created implicitly |
| Unreachable Miyo or rejected patch | Copilot keeps the local setting and shows **Saved in Copilot, but couldn't update Miyo. Open Miyo and retry.** |
| Rapid consecutive edits | Patches run in order through the existing Miyo mutation queue |
| Endpoint, device, folder, or plugin lifecycle changes while queued | The stale patch is abandoned instead of targeting the new owner |

A successful update records which endpoint, device, and folder received the projection so later changes reconcile only state Copilot owns. Documentation now distinguishes Copilot-only patterns from folder and extension filters synchronized to Miyo.

## Known review blockers

This PR is not merge-ready yet. The current implementation still needs to address:

- queued edits reusing a receipt captured before earlier queued work completes;
- an overlapping direct Miyo exclusion being claimed and later removed as Copilot-owned;
- endpoint or folder changes allowing registration evidence from the previous owner to drive a patch to the next owner;
- remote mutation starting before settings persistence succeeds, plus no reconciliation on plugin startup;
- failure copy and actions that do not match the documented contract.

The bundle-size check is also over budget on this layer.

## Non goal

- Synchronizing tags, individual notes, or property expressions that Miyo cannot represent.
- Re-registering an unregistered vault automatically.
- Replacing folder filters, Relay permissions, or write permissions configured directly in Miyo.
- Rolling back the local Copilot setting when Miyo is offline.
- Changing how Miyo indexes or searches content.

## Screenshot

No screenshot was captured because the primary change is background synchronization. Exercising the success path would mutate the shared Miyo folder registration, and forcing the failure notice would require deliberately redirecting the shared test vault to an unavailable endpoint. The exact visible failure evidence is the notice: **Saved in Copilot, but couldn't update Miyo. Open Miyo and retry.** The condition/result table in **What** captures the full before/after contract.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Client and reconciliation tests cover PATCH payloads, ownership, ordering, lifecycle changes, preservation, and failures |
| A defect would fail CI or be obvious on first use | ✅ | Every projected and preserved field is asserted by deterministic tests that run in CI |
| A revert fully restores prior state, including persisted data | ❌ | Successful patches change Miyo's registered folder, and reverting Copilot cannot reconstruct the prior server filters |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | The PATCH reuses the existing authenticated Miyo client and never changes remote read/write permissions |
| No public API, plugin API, message, or on-disk contract changes | ❌ | Copilot starts using Miyo's folder PATCH contract and persists an internal ownership receipt in settings |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | Settings subscriptions enqueue serialized network mutations and reject stale lifecycle ownership |
| No hot-path behavior lacks deterministic coverage | ✅ | Rapid edits, endpoint changes, unmounts, unregistered folders, unreachable servers, and patch failures are covered |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | A failed synchronization reports one immediate Notice from Miyo settings |

Review: inspect the projection, ownership, and queue logic in `syncMiyoIndexScope()` and its helpers in `src/miyo/miyoResync.ts`, the PATCH contract in `MiyoClient.updateFolder()` in `src/miyo/MiyoClient.ts`, subscription wiring in `CopilotPlugin.onload()` in `src/main.ts`, and the persisted receipt in `CopilotSettings` in `src/settings/model.ts`; then run Verification steps 2–6, including rapid edits, endpoint changes, direct Miyo filters, and a forced failure.

## Verification

1. Load the branch in an isolated Obsidian test vault registered with a disposable local Miyo folder, and note its current folder filters and remote read/write choices.
2. Add folder and extension inclusions/exclusions in Copilot. Confirm Miyo updates the existing registration without creating a second folder and that the receipt is stored only after success.
3. Add a tag, individual-note, and property pattern. Confirm Copilot saves them but the Miyo folder filter is unchanged.
4. Add Miyo-only filters and remote permission choices directly in Miyo, then change Copilot scope again. Confirm those Miyo-owned values survive; add a direct Copilot include whitelist and confirm it governs Copilot's projected include set.
5. Save several scope edits quickly, then repeat while changing the Miyo endpoint or disconnecting the plugin. Confirm ordered edits reach only the original owner and stale queued work is abandoned.
6. Stop Miyo or force the PATCH to fail. Confirm the Copilot setting remains saved, the failure notice appears, and reconnecting allows a later edit to synchronize normally.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -193,6 +193,12 @@ Miyo supplies Copilot's current semantic search and selected chat-history search
 
 If Miyo is reachable but the vault is not registered, the connection dialog offers **Register & connect** for a local Miyo. With a remote server or mobile setup, select **Open Miyo**, add the vault there, then **Retry**. Register only folders you intend Miyo and any enabled Relay clients to access.
 
+### Index scope
+
+Use **Exclusions** to skip folders, tags, individual notes, or file extensions. Use **Inclusions** to search only selected folders, tags, or notes. Exclusions take precedence, and Copilot's own working folder always stays excluded.
+
+Copilot applies every pattern to its own search results and Relevant Notes. Folder and file-extension changes also update the registered Miyo folder automatically, even when this settings tab is closed. Copilot preserves filters configured directly in Miyo and does not widen a stricter Miyo inclusion whitelist. Miyo's folder API cannot represent tags or individual notes, so those patterns remain Copilot-side filters. If Miyo is unavailable when a representable pattern changes, Copilot keeps the setting and shows that Miyo was not updated.
+
 ### Powered by Miyo
 
 | Control                          | Default                                    | Dependency and effect                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
--- a/docs/vault-search-and-indexing.md
+++ b/docs/vault-search-and-indexing.md
@@ -39,6 +39,8 @@ Use **Search scope** in the Miyo settings tab:
 
 The scope is a retrieval preference, not a security boundary. Keep Miyo's registered folders intentional, especially when you use an unrestricted scope or connect to a shared remote server.
 
+Under **Index scope**, Copilot's inclusions and exclusions control which notes its searches and Relevant Notes can return. Folder and file-extension changes also update this vault's registered Miyo folder automatically. Filters configured directly in Miyo remain in place, and Copilot does not widen a stricter Miyo inclusion whitelist. Tags and individual-note patterns still filter Copilot results, but Miyo continues indexing them because its folder API cannot represent those pattern types. If Miyo cannot be reached during an update, Copilot keeps the local setting and tells you that Miyo was not updated.
+
 ### Relevant Notes
 
 Miyo is the source of Relevant Notes results and similarity percentages; Copilot's legacy local embedding index no longer scores this pane. When Miyo is available, results can include direct links and backlinks that pass Copilot's search scope.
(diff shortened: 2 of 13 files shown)


zeroliu:
Closing per the inclusion/exclusion design in `designdocs/COPILOT_ACCESS_SCOPE_AND_MIYO_INDEX_STATUS.md`: Copilot no longer edits or syncs index scope for Miyo. Miyo owns its folder filters, and a separate epic (Brevilabs/obsidian-copilot-private#358) brings inclusion and exclusion back as a Copilot access scope after the Orama retirement epic completes.

PR #3006, opened 2026-08-28
[Live Plus catalog] - Step 4: Keep unavailable defaults explicit

zeroliu:
Fixes Brevilabs/obsidian-copilot-private#319

## Why

After OpenCode starts without a saved Plus default, its backend can report a usable native model. Treating that fallback as the user's selection would silently send prompts through the wrong model and make the saved default disappear from the picker.

## What

| Saved default state | Selected row | Prompt behavior |
| --- | --- | --- |
| Catalog or backend registration is still pending | Saved Plus model stays selected as **Loading…** | Prompts wait; the backend fallback is not accepted |
| Catalog failed or no longer contains the model | Saved Plus model stays selected as **Unavailable** | Prompts remain blocked until another model is selected |
| User selects a ready alternative | The alternative becomes selected | Queued prompts may run through the explicit selection |

Opening and closing the model picker does not commit the first enabled fallback, and both model-RPC and config-option model changes clear the pending default.

## Non goal

This does not retry the catalog within the same plugin lifecycle, automatically replace an unavailable default, or change non-Plus defaults.

## Screenshot

The exact states are deterministic in the adjacent component-gallery stories, but the Radix popover closes when the Obsidian CLI triggers a screenshot, so a captured image would omit the changed row. Reviewable state contract:

| Story | Saved row | Ready alternative |
| --- | --- | --- |
| `SavedModelLoading` | **Copilot Plus Flash · Loading…** | **GPT-5** selectable |
| `SavedModelUnavailable` | **Copilot Plus Flash · Unavailable** | **GPT-5** selectable |

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Session, picker, prompt-gate, config-option, trigger, and story-state tests cover the behavior |
| A defect would fail CI or be obvious on first use | ✅ | Tests cover fallback prevention and the selected-row state is immediately visible |
| A revert fully restores prior state, including persisted data | ✅ | This layer does not migrate or delete persisted data |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | It consumes the existing catalog verdict without changing credentials |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Pending selection is internal Agent-session state only |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | Agent sessions now defer a saved model across startup and explicit replacement |
| No hot-path behavior lacks deterministic coverage | ✅ | Fresh, adopted, loading, unavailable, config-option, and prompt paths are covered |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Any visual regression appears in the model picker before a prompt can run |

Review: inspect initial selection handling and `sendPrompt()` in `AgentSession` in `src/agentMode/session/AgentSession.ts`, the `deferModelSelection` policy in `src/agentMode/index.ts`, `buildPickerEntries()` in `src/agentMode/ui/agentModelPickerHelpers.ts`, and `ModelEffortPicker` in `src/components/ui/ModelEffortPicker.tsx`; then run Verification steps 1-4, including both loading and failed-catalog variants.

## Verification

1. Save a Plus model as the OpenCode default, throttle the models endpoint, reload Copilot, and open Agent Chat.
2. Confirm the saved row remains selected as **Loading…**, another ready model is selectable, and opening then closing the picker does not change the selection.
3. Attempt to send before settlement, then select a ready alternative. Confirm the prompt waits before the explicit selection and runs afterward.
4. Repeat with the endpoint blocked. Confirm the saved row changes to **Unavailable**, the native fallback is never used, and selecting an alternative recovers the session.


logancyang:
P1 verified in the live OpenCode runtime: a saved Plus model with an unsupported saved effort is sent with an `/<effort>` suffix, rejected, and the new chat silently falls back to Big Pickle. The merged fix in #3098 is correct and covers this exact case by applying the bare model ID and dropping the unsupported effort. This stack predates that merge, so please rebase it onto current `master` before further testing and review.

PR #3011, opened 2026-08-29
[Live Plus catalog] - Step 1: Load one catalog per plugin lifecycle

zeroliu:
Relates to Brevilabs/obsidian-copilot-private#319

## Why

Copilot Plus model availability was split between a hardcoded client list, persisted configured-model rows, and ad hoc models-endpoint reads. A stale row could therefore survive after the service removed a model, while separate consumers could disagree about the current lineup.

## What

Each plugin lifecycle now owns one models-endpoint request and one observable catalog snapshot. A successful response reconciles the Plus provider and its persisted model metadata; malformed, empty, timed-out, or failed responses publish an error snapshot without erasing the last saved rows. Sign-in and sign-out writes remain serialized, and a plugin reload is the retry boundary.

## Non goal

This does not switch Chat, Settings, or Agent Mode pickers to the live snapshot, and it does not gate OpenCode startup. Those behaviors belong to later PRs in this stack.

## Screenshot

Not applicable. This PR establishes the catalog lifecycle and persistence boundary without changing rendered UI.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Catalog, reconciliation, timeout, malformed-response, and rapid license-change tests cover the new behavior |
| A defect would fail CI or be obvious on first use | ✅ | The deterministic lifecycle tests run in CI |
| A revert fully restores prior state, including persisted data | ❌ | A successful authoritative reconciliation can remove saved rows for models the server removed |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | License validation and credential destinations are unchanged; the catalog request is public |
| No public API, plugin API, message, or on-disk contract changes | ❌ | Persisted Plus model metadata now carries server-provided reasoning efforts and limits |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | Plugin startup now owns a serialized one-shot catalog lifecycle |
| No hot-path behavior lacks deterministic coverage | ✅ | Success, timeout, malformed data, failure, and overlapping settings changes are covered |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | No behavior in this layer depends only on human verification |

Review: inspect `CopilotPlugin.onload()` in `src/main.ts`, `createCopilotPlusSyncQueue()` and `readLiveModels()` in `src/modelManagement/setup/copilotPlusSync.ts`, and Plus reconciliation in `CopilotPlusSetupApi` in `src/modelManagement/setup/CopilotPlusSetupApi.ts`; then run Verification steps 1-3, including the failed-response variant.

## Verification

1. Sign in to Plus, reload Copilot, and confirm the public models endpoint is requested once.
2. Change Plus settings and visit model-related surfaces without reloading. Confirm no second catalog request starts during the same plugin lifecycle.
3. Repeat with the endpoint blocked or returning malformed data. Confirm saved Plus rows are not erased and reloading the plugin starts a new request.

### DIFF
--- a/src/LLMProviders/brevilabsClient.ts
+++ b/src/LLMProviders/brevilabsClient.ts
@@ -193,8 +193,11 @@ export interface UsageResponse {
 export interface BrevilabsModelEntry {
   id?: string;
   label?: string;
+  description?: string;
   /** Input context window as a display string: `1M`, `256K`. */
   context_length?: string;
+  supports_images?: boolean;
+  supports_reasoning?: boolean;
   /**
    * Thinking-effort levels this model distinguishes, ascending. Empty means the model
    * honors none of them. Absent from services older than the field, which is why the
@@ -485,11 +488,11 @@ export class BrevilabsClient {
   }
 
   /**
-   * The models host's public catalog. Unauthenticated, and the only place the context
-   * window of a Copilot Plus model is published, so the usage meter can size its ring.
+   * The models host's public catalog. Unauthenticated and authoritative for the
+   * current Plus lineup; it also supplies usage-meter capability metadata.
    *
-   * Returns null rather than throwing, for the same reason as {@link getUsage}: this
-   * feeds a gauge, and a gauge that cannot be drawn is not an error worth raising.
+   * Returns null rather than throwing so callers can represent network failure
+   * without turning it into plugin-startup failure.
    */
   async getModels(): Promise<BrevilabsModelsResponse | null> {
     try {
--- a/src/main.test.ts
+++ b/src/main.test.ts
@@ -58,6 +58,11 @@ import { resetMiyoMutations } from "@/miyo/miyoResync";
 import { flushPersistence } from "@/services/settingsPersistence";
 import { isDesktopRuntime } from "@/utils/desktopRuntime";
 import { disposeNotificationSound } from "@/utils/notificationSound";
+import { KeychainService } from "@/services/keychainService";
+import { BrevilabsClient } from "@/LLMProviders/brevilabsClient";
+import * as modelManagement from "@/modelManagement";
+import * as settingsMigrations from "@/settings/migrations";
+import * as settingsModel from "@/settings/model";
 
 /**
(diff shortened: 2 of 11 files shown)


logancyang:
The backend catalog contract is in Brevilabs/brevilabs-models#94. Please use this minimal design for the client adoption:

- Fetch `/v1/models` once when the plugin starts. Treat IDs as opaque strings: each separately selectable version has its own ID, so multiple versions of one family can coexist. Do not parse model names or add model-specific branches.
- For an existing ID, refresh its display name and capabilities (including `supports_tools`, image support, context, and reasoning efforts), preserving its saved identity, selection, and enabled/disabled choices.
- Add a newly discovered ID using `default_enabled`, for both new and existing users. Apply this only when adding the ID; subsequent refreshes must preserve the user's toggle.
- On a successful catalog fetch, remove absent IDs from the selectable lineup. If the selected ID disappears, select Copilot Plus Flash.
- If fetching fails or the response is malformed, retain the last successful catalog and saved settings. The next plugin restart retries.
- Chat, Settings, and Agent Mode/OpenCode must consume the same reconciled catalog.

For this rollout the published IDs are `minimax-m3` (MiniMax M3) and `glm-5.3-flash` (GLM-5.3-Flash), both with `default_enabled: true`. There is no `replaces` or `enable_for_existing` field and no `minimax` or `glm-flash` family alias. A saved M2.7 selection falls back to Copilot Plus Flash after the successful refresh; M3 and GLM-5.3-Flash appear enabled as new IDs.

The backend temporarily accepts the old M2.7 request IDs for older plugin versions, but excludes them from discovery. That compatibility path is separate from selectable versions.

Acceptance: change only the backend catalog, restart the updated plugin, and see the new models with the correct names and defaults in every picker and OpenCode. Also verify two versioned IDs in the same family coexist, user-disabled models stay disabled across restarts, removed selections fall back, and failed fetches preserve the cached lineup. No further plugin release should be necessary for ordinary model additions or retirements.

PR #3084, opened 2026-09-02
[Relevant Notes on Miyo] - Step 2: Preserve hidden property filters

zeroliu:
Relates to Brevilabs/obsidian-copilot-private#280

## Why

`PatternListEditor` displays folders, tags, notes, and extensions, but the same stored value can also contain property filters created by another surface. Editing any visible badge currently rebuilds the value from only the four displayed categories and silently deletes those hidden property filters.

The component has no active settings caller on `master`, but the later Index-scope change reuses it. Its data-integrity rule should land and be reviewed before that feature makes the latent deletion reachable.

## What

| Edit | Before | After |
| --- | --- | --- |
| Remove a visible pattern | Hidden property filters are discarded | Hidden property filters round-trip unchanged |
| Add a custom pattern | Hidden property filters are discarded | Hidden property filters round-trip unchanged |

The editor continues to display and edit the same four visible pattern categories.

## Non goal

- Exposing `PatternListEditor` on a settings page.
- Displaying, adding, or removing property filters in this editor.
- Changing Relevant Notes, Miyo, local Orama indexing, or filter matching semantics.

## Screenshot

Not applicable. The visible editor is unchanged; this preserves a hidden part of its serialized value.

## Risk

**Low**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Tests cover both editor rebuild paths that could discard hidden filters |
| A defect would fail CI or be obvious on first use | ✅ | Each output value is decoded and compared after the edit |
| A revert fully restores prior state, including persisted data | ✅ | This component returns values to callers but does not persist or migrate data itself |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | Existing parsed filter values are only carried through unchanged |
| No public API, plugin API, message, or on-disk contract changes | ✅ | The internal component keeps the existing serialized format |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Both changed paths are synchronous editor callbacks |
| No hot-path behavior lacks deterministic coverage | ✅ | Removal and custom-add behavior are covered directly |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Any regression appears in the editor's next emitted value |

Review: inspect the value reconstruction in `PatternListEditor()` in `src/settings/v2/components/PatternListEditor.tsx`, then run Verification.

## Verification

1. Run `npm test -- --runInBand src/settings/v2/components/PatternListEditor.test.tsx`.
2. Confirm removing a visible tag leaves the hidden property filter in the emitted value.
3. Confirm adding a custom pattern leaves both the existing visible tag and hidden property filter in the emitted value.

### DIFF
--- a/src/settings/v2/components/PatternListEditor.test.tsx
+++ b/src/settings/v2/components/PatternListEditor.test.tsx
@@ -1,0 +1,63 @@
+/* eslint-disable @eslint-react/hooks-extra/no-unnecessary-use-prefix -- Mock exports must preserve production hook names. */
+import { PatternListEditor } from "@/settings/v2/components/PatternListEditor";
+import { createPatternSettingsValue, getDecodedPatterns } from "@/search/searchUtils";
+import { fireEvent, render, screen } from "@testing-library/react";
+import React from "react";
+
+jest.mock("@/context", () => ({ useApp: () => ({}) }));
+
+jest.mock("@/components/modals/CustomPatternInputModal", () => ({
+  CustomPatternInputModal: jest
+    .fn()
+    .mockImplementation((_app: unknown, onSubmit: (value: string) => void) => ({
+      open: () => onSubmit("Archive"),
+    })),
+}));
+
+beforeAll(() => {
+  (window as unknown as { activeDocument: Document }).activeDocument = window.document;
+  if (!("PointerEvent" in window)) {
+    (window as unknown as { PointerEvent: typeof MouseEvent }).PointerEvent = MouseEvent;
+  }
+  Element.prototype.hasPointerCapture = () => false;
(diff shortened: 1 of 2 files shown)


zeroliu:
Closing per the inclusion/exclusion design in `designdocs/COPILOT_ACCESS_SCOPE_AND_MIYO_INDEX_STATUS.md`: the only planned caller of this editor (Step 5) is closed, and the future Copilot access scope editor (Brevilabs/obsidian-copilot-private#358) will display property patterns rather than hide them, which supersedes this fix.

PR #3105, opened 2026-09-04
Add llmman as a Self Host BYOK provider template

ericcurtin:
Adds [llmman](https://github.com/llmmanorg/llmman), a local model runner that serves the Ollama and OpenAI APIs on port 17434.

- `src/modelManagement/catalog/builtinDefinitions.ts`: one new `LOCAL_PROVIDER_DEFINITIONS` entry (`id: "llmman"`, `providerType: "openai-compatible"`, `defaultBaseUrl: "http://localhost:17434/v1"`, `requiresApiKey: false`), the same shape as the Ollama and LM Studio templates, so it appears in the **Self Host** group and gets chat, embeddings, and `/v1/models` discovery with no further code.
- Appended after LM Studio; the existing list is not alphabetical and I did not reorder it.
- Legacy `ChatModelProviders` / `byokMigration` maps untouched: llmman never existed in pre-BYOK settings, so there is nothing to migrate.
- Docs: listed llmman next to Ollama and LM Studio in `docs/llm-providers.md`, `docs/settings.md`, and `docs/copilot-plus-and-self-host.md` (two tables re-aligned by Prettier).

**Testing:** `npx tsc --noEmit`, `npm run format:check`, and `npx jest src/modelManagement/catalog src/modelManagement/ui/dialogs/AddProviderDialog.test.tsx src/modelManagement/ui/dialogs/ConfigureProviderDialog.test.tsx` all pass; not run against a live llmman server or inside Obsidian.

> AI-assisted, reviewed before submitting.

### DIFF
--- a/docs/copilot-plus-and-self-host.md
+++ b/docs/copilot-plus-and-self-host.md
@@ -73,14 +73,14 @@ After the included Plus period ends, the lifetime Self-Host and Miyo benefits re
 
 The label in the model picker tells you which service handles the conversation:
 
-| Model source                | How access works                                               | Who processes the prompt                                        |
-| --------------------------- | -------------------------------------------------------------- | --------------------------------------------------------------- |
-| **Copilot-hosted**          | Lite, Plus, or included Supporter access                       | Brevilabs and the enterprise model provider used for that model |
-| **Cloud BYOK**              | Your API key under BYOK                                        | The provider and endpoint you configured                        |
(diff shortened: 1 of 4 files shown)


zeroliu:
Thanks for the contribution! llmman can already be configured through Add a custom provider with http://localhost:17434/v1 as the base URL and no API key.

We’re keeping the built-in provider list small, so I’m going to close this PR. A docs example showing how to connect llmman through the custom provider flow would be welcome. Thanks again for taking the time to put this together!

Issue #2886, opened 2026-08-18
Chat log fail to renew


fesolla:
- [ ] Disable all other plugins besides Copilot **(required)**
- [ ] Log file generated via "Copilot: Create Log File" command or Settings -> Advanced -> Create Log File **(required)**
- [ ] Screenshot of note + Copilot chat pane + dev console added **(optional)**

Copilot version: v4.0.1
Model used: Github Copilot gpt-4.1

**Describe how to reproduce**
Sending a message during the log file is opened causes the chat to fail to renew.
The new message would be lost and the whole conversation cannot be renew anymore

https://www.youtube.com/watch?v=MOeUEDm0vdQ&feature=youtu.be

As shown in the video, the "testing" and "testing2" were correctly saved in the log file, and can be called back after switching to other chats. However, after I sent the "testing3" message with the log file opened, the log file will not be renewed and the message disappeared after switching to another chat. The whole conversation cannot be renewed anymore after that.

Before the v4 update, as I recall, the log file would automatically renew when a new message is sent.

zeroliu:
Next step: reproduce the save failure while the conversation Markdown file is open, identify whether the editor/file watcher races the autosave write, and add a regression test proving the newest message survives view switches and reopen.

PR #3158, opened 2026-09-09
Keep reopened Quick Chat conversations in their original notes

zeroliu:
Relates to #2886

## Why

In Quick Chat, reopen a saved conversation after its note receives a generated title, then send another message with the source note open. The original note stops updating and the new turn disappears when reopening that history entry. Reloading drops milliseconds from the conversation identity, so the next save creates a different note.

## What

> **Before:** Reopen the renamed chat, send `testing3`, and switch away and back: the original chat ends at `testing2`.
>
> **After:** The same sequence updates the original note and retains `testing3` and its response.

Existing saved chats retain their original identity. Chats without a valid saved identity keep their existing timestamp fallback.

## Non goal

Recovering turns already split across duplicate notes or changing how Markdown editors save.

## Screenshot

No visual change. The observable result is the original note's saved content and the messages restored from history.

## Risk

**High**, because selecting the destination for a conversation save affects vault files.

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Save/load continuation and invalid-identity tests cover the change |
| A defect would fail CI or be obvious on first use | ✅ | The regression asserts that continuation updates the original note |
| A revert fully restores prior state, including persisted data | ❌ | Reverting does not undo conversation writes made while installed |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | Loading reads the saved YAML identity, with malformed-input fallback |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Existing epoch field and Markdown format remain unchanged |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Only synchronous history parsing changes |
| No hot-path behavior lacks deterministic coverage | ✅ | Numeric, quoted, absent, invalid, and missing-display cases are covered |
| No new dependency | ✅ | Uses Obsidian's existing YAML parser |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Quick Chat history continuation is visible after switching chats |

Review: inspect `parseChatContent` in `src/core/ChatPersistenceManager.ts` and the `loadChat()` tests in `src/core/ChatPersistenceManager.test.ts`. Run both verification cases below. A rollback leaves saved notes readable but restores the old continuation bug.

## Verification

1. Enable Quick Chat autosave. Send `testing` and `testing2`, wait for responses and a generated note title, then start another conversation. Reopen the first chat from Chat History, open its source note, and send `testing3`. Confirm that the original note updates, no duplicate appears, and all three turns survive switching away and back.
2. Repeat with an existing saved chat and the source note closed. Also load a copied legacy chat with no epoch or malformed YAML; its messages should remain readable through the timestamp fallback.

### DIFF
--- a/docs/chat-interface.md
+++ b/docs/chat-interface.md
@@ -55,7 +55,7 @@ After a response, the token counter in the top bar shows the context used for th
 
 Use **New Chat** in the top bar or run **New Copilot Quick Chat**. Starting over clears the current messages and resets context to the active note when automatic context is enabled.
 
-Use **Chat History** to search saved conversations, reopen one, rename it, open its Markdown source, or delete it.
+Use **Chat History** to search saved conversations, reopen one, rename it, open its Markdown source, or delete it. Continuing a reopened conversation updates its existing note, including after the note has been renamed or while its source is open.
 
 **Autosave Chat as Markdown** is enabled by default. Copilot saves after each user message and response under `<Copilot folder>/copilot-conversations/`. If autosave is off, use **Save Chat as Note** in the top bar. Change autosave and **Conversation Filename Template** under **Settings → Copilot → Basic → Saving conversations**.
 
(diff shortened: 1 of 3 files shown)


zeroliu:
@codex review

PR #3167, opened 2026-09-09
Keep agent running state correct after Stop

zeroliu:
Relates to Brevilabs/obsidian-copilot-private#365

## Why

Pressing Stop with a queued follow-up can send that follow-up, then hide its running timer and Stop button while the agent is still working.

## What

Stop discards queued follow-ups. A new message sent after Stop keeps its running indicator until its own turn ends.

| Action | Result |
| --- | --- |
| Stop with a queued follow-up | Cancel the turn and discard the queue |
| Send again after Stop | Show the timer and Stop button throughout the new turn |
| Let a turn finish normally | Send queued follow-ups automatically |

## Non goal

Changing normal queue flushing or deriving transcript running state from session status.

## Screenshot

Real Obsidian runs with Claude, with only the cancellation acknowledgement delayed by 800 ms to expose the race.

Before: the queued message is sent, but its assistant row and composer appear idle while the session is running.

Before Stop race

After: Stop discards the queued message and the cancelled turn becomes idle.

After Stop clears queue

## Risk

**High**, because this changes cancellation ordering on the chat send path.

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Real draft-hook tests cover queued Stop, delayed cancellation, and cancellation failure |
| A defect would fail CI or be obvious on first use | ✅ | Regression tests assert sends and loading state |
| A revert fully restores prior state, including persisted data | ✅ | No persisted format or migration changes |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | Stop changes handling of queued messages |
| No public API, plugin API, message, or on-disk contract changes | ✅ | All existing contracts remain intact |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | Cancellation no longer performs delayed loading cleanup |
| No hot-path behavior lacks deterministic coverage | ✅ | Tests include normal queue flush and composer unmount cleanup |
| No new dependency | ✅ | Dependencies are unchanged |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Timer and Stop button are visible in Agent Chat |

Review: inspect `handleStopGenerating` and `runSend` in `src/agentMode/ui/AgentChatInput.tsx`, then follow steps 2–4 below.

## Verification

1. Open Agent Chat with an installed agent and request a long response.
2. While it runs, send a follow-up, then press Stop. The queue should disappear without adding that follow-up to the transcript.
3. Send a new message immediately after Stop. Its timer and Stop button should remain visible until it finishes, including when cancellation acknowledgement is delayed.
4. Queue a follow-up and let the first turn finish normally. The follow-up should send automatically and show its running state. Run the focused `AgentChatInput.test.tsx` suite to exercise cancellation rejection and delayed acknowledgement deterministically.

### DIFF
--- a/docs/agent-mode-and-tools.md
+++ b/docs/agent-mode-and-tools.md
@@ -119,6 +119,7 @@ Agent Chat keeps each conversation separate:
 
 - Select **+** for another session. Each tab keeps its own history, draft, attachments, and queued follow-ups.
 - Select **New Chat** to reset the current tab.
+- Select **Stop** to cancel the current turn and discard its queued follow-ups.
 - Use **Recent Chats** from the Agent Chat home screen, or **Chat History** inside a conversation, to resume saved work. The **Recent Chats** list can be scrolled or searched.
 - Add the active note, selected text, other notes, folders, a Copilot Web Viewer tab, or supported images. You can also mention a note with `[[Note title]]`.
 - Hover the context ring beside the send controls to see how much of the model's context window is in use. The ring stays empty until the agent reports usage, and a stopped response keeps the last reported reading. If the connected account reports usage limits, the same panel shows the available limit and reset time.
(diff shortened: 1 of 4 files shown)


zeroliu:
@codex review

zeroliu:
@codex review

PR #3238, opened 2026-09-14
Pack the issue report in memory and guide the upload in two pages

Emt-lin:
Relates to Brevilabs/obsidian-copilot-private#202

## Why

Reporting a bug ends with the user assembling the evidence by hand. **Report an issue** writes `report.md`, `screenshot.png`, and the logs into a folder, opens that folder in the file manager, opens a GitHub issue page, and says "Drag the saved files into the issue to attach them, then submit." The user has to recognise which files matter and drag them in one at a time.

Three concrete problems behind that:

- **A source that could not be read is dropped silently.** The report arrives missing the log the maintainer needed, with nothing on the issue saying so — the user believes they sent it.
- **The attached chat log stops one request short of the failure being reported.** It is exported without draining the prompt recorder, so the request the user is complaining about is the one request missing.
- **The bundle lands in a fixed path under the shared temp dir.** On a machine where that dir is shared between accounts, another account can create that path first, and a symlink planted inside it is followed by the write that lands there. Reproduced locally: a victim file outside the temp dir was overwritten and chmod-ed.

#2755 replaced the hand-off with a guided upload but grew around an editable staging folder, a Rebuild step and a three-page dialog. The maintainer asked for the same outcome without that machinery: remove the staging-folder model, pack once in memory, write only the zip, collapse the flow to describe/review, use one retry behaviour, and open GitHub directly after success. This PR is that rebuild.

## What

> **Before:** the user clicks **Report an issue**, describes the problem, and is handed a folder plus a browser tab. They drag `report.md`, `screenshot.png`, and two logs into the issue one at a time, with no way to tell that the activity log was empty and never written.
>
> **After:** one dialog with two pages. Page one takes the description and the sources to include. Page two lists every source with what actually happened to it and the one `copilot-report-….zip` that was packed in memory. **Upload & open issue** is a separate click; the issue opens with the report ID already in the body and only **Submit** left to press, and the dialog closes.

Behavior by outcome:

| Condition | Before | After |
| --- | --- | --- |
| A selected log is missing, empty, or unreadable | dropped silently | listed with a cross and the reason, before anything is sent |
| A log is larger than the room left in the 24 MiB report | dropped silently | the whole log is redacted first, then its newest entries are kept; the row and the file's opening banner say so. A log over 64 MiB is left out and says why |
| A credential sits where the log is cut | n/a | redaction runs over the whole log before it is cut, so a `password=` on one line and its value on the next cannot be split apart |
| Upload fails, for any reason | n/a | the same three actions every time: **Retry upload** re-sends the identical bytes under the same idempotency key, so it is stored at most once; **Show zip**; **Open issue anyway** |
| The dialog is closed while uploading | n/a | the upload finishes in the background; a notice shows the report ID and a link to the issue; the zip stays on disk |
| The browser cannot be opened after a successful upload | n/a | the dialog still closes; the notice carries the ID and the issue link |

The uploaded bundle is private. The issue carries only an opaque report ID, with no download surface behind it. Nothing leaves the machine until **Upload & open issue** is pressed, and the only file written to disk is the zip, in a private directory created for that report.

## Non goal

Carried forward from the issue:

- A pre-report diagnostics skill that triages user-fixable problems first — Brevilabs/obsidian-copilot-private#203.
- Upload progress or a cancel button: Obsidian's `requestUrl` supports neither, so the uploader promises neither.

Added during implementation:

- Redacting a home-directory username that contains a space — Brevilabs/obsidian-copilot-private#302. The rules predate this branch; both obvious fixes trade the leak against swallowing prose that follows a bare path, so it needs its own design.
- Editing the packed files before upload, and a Rebuild step. Untick a source and prepare again instead.
- A completion page or a copy button for the report ID; the ID lives in the issue body and the notice.
- Deleting the local zip after a successful upload.
- Per-chat separators and an identifier list inside the exported log, asked for by a user on #2755 — a separate issue.
- Restoring the **Create Log File** settings row. The guided flow attaches the same chat log, and the command palette still has **Copilot: Create log file**.

## Screenshot

Before: on `master`, **Report an Issue** writes the files into a folder, opens that folder and a browser tab, and asks the user to drag the files into the issue. There is no dialog to capture; the after state below is the whole new surface.

**Settings → Advanced → Debugging & support** (after):

Debugging & support section: Report an issue button, Debug Mode switch, and the Agent Mode activity log switch with Open and Clear

Page one, describe (after):

Report an issue dialog: description field, four attachment checkboxes, the Before you upload consent callout, Cancel and Prepare report

Page two, review (after):

Report contents: one row per packed source with its size, the zip name and size, and Cancel, Show zip, Upload & open issue

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | `issueReport.test.ts` covers every source outcome, the budget order, redaction before the cut on four adversarial fixtures over both the in-memory and on-disk paths, the 64 MiB ceiling and the 64 KiB floor; `ReportIssueModal.test.tsx` covers prepare, upload, discard and close-mid-upload; `ReportIssueFlow.test.tsx` covers both pages and the failure actions |
| A defect would fail CI or be obvious on first use | ❌ | The adapter's tests drive a fake `request`, so a drift in the live endpoint's contract surfaces only against the real service, as a failed upload after the user has already packed a bundle |
| A revert fully restores prior state, including persisted data | ✅ | No settings migration and no vault write. The one persisted artifact is an install-id UUID in `localStorage`, which a revert leaves behind unread and inert |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | `redactLog.ts` decides what is stripped from every text attachment and from the issue title and body; `reportInstallId.ts` mints and persists a new identifier; the flow uploads user-selected private content |
| No public API, plugin API, message, or on-disk contract changes | ❌ | Adds the client side of the report upload contract the backend already ships; the zip layout is the folder's (`report.md` at the root, screenshot, logs) |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | An upload may outlive the dialog; results that land after unmount are dropped and reported by notice, covered by tests |
| No hot-path behavior lacks deterministic coverage | ✅ | Every branch of the uploader — refused, timed out, thrown transport, malformed 2xx — and of the assembler is covered; the only untestable piece is the real browser open, which has a notice fallback |
| No new dependency | ❌ | Adds `fflate` (MIT, no transitive dependencies) for `zipSync` |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Confined to the report flow; a wrong per-source outcome is visible on the review page before anything is sent |

Review: read `logEntry` in `src/utils/issueReport.ts` line by line — redaction over the whole log before any cut, the ceiling, the floor — then `replaceEachRun` in `src/utils/redactLog.ts`. Then `createReportUploader` in `src/utils/reportUpload.brevilabs.ts` for response validation and the fixed error copy. Then `prepare`, `upload` and `onUploaded` in `src/agentMode/ui/ReportIssueModal.tsx` for what touches disk and what runs after the dialog is gone. Run Verification steps 2–7, including the failure and close-mid-upload variants.

## Verification

1. Install this branch, open **Settings → Copilot → Advanced → Debugging & support**, and enable **Debug Mode**. Open an Agent chat and send a message so there is activity to attach.
2. Press **Report an issue**. Confirm the consent copy names redaction, screenshots, and the 60-day retention, and that the screenshot option is offered because an agent pane is open. Type a description containing `password=hunter2` and press **Prepare report**.
3. On the review page, confirm the dialog hid itself for the capture — the screenshot should show the chat pane, not the dialog. Confirm every ticked source has a row with a size or a reason, and that the zip is named `copilot-report-….zip`.
4. Press **Show zip**. The folder holds exactly one file. Open the zip: `report.md` sits at the root and reads `password=<redacted>`; a log that was too large opens with the "earlier entries omitted" banner naming its original size.
5. Press **Upload & open issue**. The browser opens a prefilled issue carrying the report ID, the dialog closes, and a notice confirms. Nothing on the issue links to the bundle. Do not submit.
6. **Failure path:** disconnect the network, prepare again, upload. The page shows the error and offers **Retry upload**, **Show zip** and **Open issue anyway**. Reconnect and press **Retry upload**: success as in step 5.
7. **Closing mid-upload:** prepare again, press upload and close the dialog immediately. A notice with the report ID and the issue link appears when the upload lands; the zip is still on disk.
8. **Mobile:** on a phone or under `app.emulateMobile(true)`, confirm **Report an issue** declines with a notice rather than throwing while loading the settings tab.

## Note

Supersedes #2755. Against the shape asked for there: no staging folder, no Rebuild, no completion page, no retryable/final classification; the review page is read-only and the only disk write is the zip.

Size is above the 2,500–3,000 lines suggested: production code is about 2,250 lines, the rest is tests and docs. Log tailing is the one block that could still go (roughly 600 lines with tests); without it a log larger than the room left is dropped whole instead of keeping its newest entries.

The 16 MiB ReDoS guard in `redactLog.test.ts` occasionally hits V8's regexp stack limit under a full parallel run and passes alone; it predates this change and is worth a follow-up that bounds the address-run quantifier.


Emt-lin:
@codex review

PR #3261, opened 2026-09-16
Fix large-log redaction and test every PR on Windows

logancyang:
Fixes Brevilabs/obsidian-copilot-private#479

## Why

A diagnostic log with a very long email candidate or Basic credential can overflow the regex stack and be omitted from an issue report. This surfaced on Windows only after the 4.0.9 release PR merged; the release workaround skipped its stress test.

## What

Redact complete email runs and Basic credentials without matching their entire payload through a repeated regex. Restore the 16 MiB regression test on every platform while preserving punctuation, padding, and the existing redaction policy.

Run the full native Windows suite on every PR, including drafts. Releases reuse the same checks against their exact merge commit and still wait for both the publishing-wrapper and remaining-suite jobs.

## Non goal

Change report uploads, Windows permissions, or other credential/path matchers. Republish 4.0.9 or change branch protection.

## Screenshot

Not applicable: this changes log sanitization and CI, with no visual changes.

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Restored 16 MiB credential and new whole-address, padding, delimiter, and repeated-header cases cover the scanners |
| A defect would fail CI or be obvious on first use | ✅ | Windows runs the restored cases normally and with regex optimization disabled |
| A revert fully restores prior state, including persisted data | ✅ | No settings or stored-data migration |
| No auth, permissions, secrets, or input-handling surface changes | ❌ | Changes secret scrubbing before diagnostic logs are attached |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Redaction markers, public signatures, and report formats stay unchanged |
| No core-path concurrency, async-lifecycle, or state-machine changes | ✅ | Sanitization remains synchronous |
| No hot-path behavior lacks deterministic coverage | ✅ | Existing redaction tests and long-input regressions exercise the changed scans |
| No new dependency | ✅ | Uses the existing runtime and Actions dependencies |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Sanitization and workflow contracts are automatically checked |

Review: inspect `redactAddressRuns`, `redactBasicCredentials`, and `RULES` in `src/utils/redactLog.ts`; verify the `windows-tests` dependency in `.github/workflows/release.yml` and both jobs in `.github/workflows/windows-tests.yml`. Follow Verification 1–3.

## Verification

1. Run `npm test -- --runInBand src/utils/redactLog.test.ts src/agentMode/session/debugSink.test.ts`. All tests should pass, including the previously skipped stress case, complete-address removal, empty credentials, padding, and repeated headers.
2. Run `node --no-regexp-optimization node_modules/jest/bin/jest.js --runInBand --runTestsByPath src/utils/redactLog.test.ts`. The same redaction cases should pass without stack overflow.
3. Inspect this PR's Windows checks: both jobs should pass, and the retained artifact should include the tested commit, full-suite results, and the unoptimized sanitization results. Release callers must pass their exact merge SHA into this same workflow.

## Note

Other pre-existing regex rules can still overflow on pathological multi-megabyte provider keys, bearer tokens, field values, usernames, or header whitespace. This fixes the confirmed address/Basic payload failures; it does not claim arbitrary-input hardening for every rule. Report assembly continues to omit a log on redaction failure rather than attaching unredacted text.


logancyang:
Review (Andrej): pushed 34579f9a, which drops `--runInBand` from the full Windows suite step. That step ran 516 suites serially in 322s, against 111s for the same suites on the Linux job with default workers. The PowerShell wrapper test that originally needed isolation already runs in `publish-windows`. The `--no-regexp-optimization` step stays serial. If a worker runs out of memory on the 16 MiB redaction cases, the fallback is `--maxWorkers=2`, not a return to serial.

PR #3262, opened 2026-09-16
Keep outdated agents selectable and guide users through upgrades

zeroliu:
Relates to Brevilabs/obsidian-copilot-private#480

## Why

Bumping the managed download alone leaves existing users with inconsistent behavior. OpenCode still reports its old installation as ready. Codex reports it as outdated, but its saved models disappear or become unselectable. Users need to reach the agent and its upgrade instructions without running an unsupported binary.

## What

Managed downloads and minimum support now use the same release version: OpenCode **1.18.31**, Codex ACP **1.12.0**. The minimum applies to both managed and custom installations; newer supported versions remain usable.

| State | Result |
| --- | --- |
| Installed version is too old | Saved models remain selectable; chat cannot execute |
| Outdated model selected | Chat shows the installed and required versions, with Configure |
| Configure opened from chat or Agents settings | Choose a managed upgrade or supply your own binary |
| Configure dismissed | Return to the original chat and unsent draft |
| Upgrade succeeds | Start one chat on the chosen model; never send automatically |
| Startup or selected model fails | Keep the source chat and selection; allow retry or another model |

## Non goal

- Polling upstream releases or upgrading automatically.
- Changing Claude's version requirement.
- Fixing OpenCode's UNC/network-share ancestor-walk crash.

## Screenshot

Outdated OpenCode selected from a working chat, with the saved model and return action retained:

OpenCode upgrade recovery with Configure, selected model, and Back to chat

The same flow for Codex:

Codex upgrade recovery with Configure, selected model, and Back to chat

## Risk

**High**

| --- | :---: | --- |
| No behavior change, or a cosmetic/copy/docs/config change visible where it renders, or deterministic tests cover the changed behavior | ✅ | Regression tests cover version floors, selection recovery, rejected models, and install ordering |
| A defect would fail CI or be obvious on first use | ✅ | Real-factory integration tests assert a single chat after an upgrade |
| A revert fully restores prior state, including persisted data | ✅ | No settings migration; existing paths and models remain readable |
| No auth, permissions, secrets, or input-handling surface changes | ✅ | Existing authentication and upgrade ownership remain unchanged |
| No public API, plugin API, message, or on-disk contract changes | ✅ | Settings and session formats are unchanged |
| No core-path concurrency, async-lifecycle, or state-machine changes | ❌ | Recovery coordinates installation, preload, and session startup |
| No hot-path behavior lacks deterministic coverage | ✅ | Startup, resume, prompt blocking, cancellation, and queued installs have regression coverage |
| Human-only behavior stays in one feature area and surfaces quickly | ✅ | Remaining visual and live-runtime checks are confined to agent configuration and chat |

Review: inspect `onInstallStateChanged`, `resumeRecoverySelection`, and `createSession` in `src/agentMode/session/AgentSessionManager.ts`, plus the installation subscription in `src/agentMode/index.ts`. Check `versionInstallState` in `src/agentMode/backends/shared/versionInstallState.ts` and run the failure/cancellation variants below.

## Verification

1. Use an old managed OpenCode or Codex installation with saved models. Open Agent Chat and select one. Confirm the version warning appears, the model stays selected, and no message can be sent.
2. Open Configure from chat and from **Settings → Copilot → Basic → Agents**. Confirm both lead to the same dialog with managed/custom choices. Repeat with an old custom binary and confirm its path is retained.
3. Select an outdated agent from a fresh working chat with an unsent draft. Dismiss Configure, then choose **Back to chat**. Confirm the original session and draft are unchanged.
4. Repeat the selection and upgrade in Configure. Confirm exactly one new chat opens on the chosen model, the draft survives, and nothing sends automatically. Send a short no-tool prompt and confirm the agent responds.
5. While recovery is pending, cancel or choose another model. Confirm completion cannot override the newer choice. Simulate a startup failure or select a model the upgraded agent no longer offers; confirm the original chat remains available and Retry or another model can recover.


zeroliu:
Superseded by a seven-PR stack that delivers the same change in reviewable steps:

1. #3281 Refuse to run an unsupported install
2. #3282 Serialize install work in one queue
3. #3283 Show the model picker on the status pane
4. #3284 Move agent upgrades into Configure
5. #3285 Keep outdated agents selectable
6. #3286 Bump managed Codex and opencode pins
7. #3287 Apply one support floor to all installs

The stack tip carries this branch's full content. The pin bump moved ahead of the support floor so recovery is in place before existing managed Codex installs start reporting as outdated.

